PR #11, opened 2020-09-25
[Long term work] Add a transactions pool

tomaka:
Adds a transactions pool to the code.


tomaka:
This PR is dead, but I'm keeping it open in order to not lose the code changes.

PR #45, opened 2020-09-30
Restore the sled database, this time in a more mature way

tomaka:
The Substrate code is extremely complicated, and I can't even figure out what exactly we store in the database.

Rather than copying what Substrate does, I went for my own approach:

- Store the header and body of all blocks of the finalized chain, and of all blocks that we discover that are descendant of the latest finalized block.
- When a block gets finalized, remove from the database all block that aren't an ancestor or descendant of the latest finalized block.
- Store the full key-value storage for the latest finalized block. For all non-finalized blocks, store the diff of their storage compared to their parent.

As a consequence, we cannot find the state of the storage of arbitrary already-finalized block without restarting blocks verification from scratch. This is a bit annoying, but we basically never need this information, and I think it's ok for the JSON-RPC server to deny storage requests concerning blocks that are too old.

Note that we could theoretically also discard the headers and bodies of the finalized chain, but we want to be good citizens and serve them to nodes that request them when syncing.

While this is opinionated, substrate-lite is designed in such a way that only the highest-level code (`full_node.rs`) needs to know about all these constraints. In other words, replacing this code with something else doesn't require changing any code outside of `full_node.rs`.


tomaka:
Progress report: only thing left to do is properly implement `set_finalized` (which is also the hardest).

It seems that my sync speed has considerably dropped, although it's hard to know how much considering that sync stops at block 512 (`set_finalized` causes a panic).
If it's indeed the case, and if the drop is too harsh, I'll restore the in-memory cache, or keep the keys in memory and not the values. It seems to me that the actual storage state is pretty small all things considered (~3MiB), and keeping it in memory is completely realistic.

PR #52, opened 2020-10-05
[Long term work] Replace the Substrate networking with new code

tomaka:
Replaces the networking code, copy-pasted from Substrate, with locally-written code.

Progress:
- [x] Noise.
- [x] Yamux.
- [x] Request-responses.
- [ ] Notification protocols.
- [ ] Ping protocol.
- [ ] Discovery.
- [x] Timeouts.
- [x] Listening for connections.
- [x] Connecting automatically to nodes, and not just manually create a `TcpStream` and read/write it.
- [x] Dispatch connections between various background tasks.


tomaka:
While this code is clearly not finished, I'd like to merge this soon, as its state is already way better (albeit more fragile) than the current networking.

As such I'm going to switch to fixing the browser node.

PR #127, opened 2020-10-30
Use the correct protocol ID for block requests

tomaka:
This should fix Kusama and Westend. I haven't tested it because the chain specs fail to parse at the moment.


tomaka:
Confirming that Westend now syncs. Kusama fails to verify block for an unknown reason, but that is out of scope from this PR.

PR #142, opened 2020-11-01
Implement block building

tomaka:
Implements the runtime-related part of the process of building a block. In other words, generating the header and body. The consensus part isn't in this PR.

The test is unfortunately failing, and I'm not 100% sure that I'm actually doing this correctly.


tomaka:
So what I'm doing wrong is that the output of `BlockBuilder_inherent_extrinsics` must actually be passed back using `BlockBuilder_apply_extrinsic`.

PR #163, opened 2020-11-07
Setup logging in the full node using tracing

### LINKED ISSUE #93: Setup logging
While logging is forbidden in the library code (i.e. anything in `src`), we can and should set it up in the actual binaries.

For the full node, I'd go with `slog`, as the full node has several tasks running in parallel, and `slog` will make it possible to filter logs by tasks. Plus it's got a variety of backends, which are very useful.

For the browser node, everything is single threaded, and good ol' `log` with the `console_log` crate should be enough.

tomaka:
Close #93


tomaka:
What confuses me and annoys me is that the informant is now extremely more slow. I don't know if `tracing` is somehow holding a lock on stdout or something, but the fact that the behaviour of the code changes is a bit frustrating and makes me want to ditch `tracing`.

PR #168, opened 2020-11-08
Fix memory leak with wasmtime

tomaka:
Reading https://docs.rs/wasmtime/0.21.0/wasmtime/struct.Store.html:

> A Store is not intended to be a long-lived object in a program. No form of GC is implemented at this time so once an instance is created within a Store it will not be deallocated until all references to the Store have gone away (this includes all references to items in the store). This makes Store unsuitable for creating an unbounded number of instances in it because Store will never release this memory. It's instead recommended to have a long-lived Engine and instead create a Store for a more scoped portion of your application.

### DIFF
--- a/src/executor/vm/jit.rs
+++ b/src/executor/vm/jit.rs
@@ -31,8 +31,12 @@ pub struct JitPrototype {
         ToCoroutine,
     >,
 
-    /// Reference to the memory imported by the module, if any.
-    imported_memory: Option<wasmtime::Memory>,
+    /// Reference to the memory used by the module, if any.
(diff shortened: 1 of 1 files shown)


tomaka:
Memory usage annoyingly still grows even after this PR (e.g. around 1.8 GiB after 380k blocks), but way slower.

tomaka:
Err, that's bad:

https://github.com/paritytech/substrate-lite/blob/f369808e882f0854abbf8013860363933f23032f/src/executor/vm/jit.rs#L141

PR #172, opened 2020-11-10
Add a function to convert a light sync state into a chain information.

expenses:


src/chain_spec.rs
```diff
@@ -39,23 +39,90 @@ use alloc::string::String;
...
+                    _ => None
+                }
+            })
+            .collect();
+
+        epochs.sort_unstable_by_key(|(&block_num, _)| block_num);
+
+        // Get the latest two epochs.
+        let prev_epoch = &epochs[epochs.len() - 2].1;
+        let current_epoch = &epochs[epochs.len() - 1].1;
```
expenses, line 84: I still don't get how these epochs work, honestly. I thought that at least one epoch would start in the future relative to the finalized header, but in [polkadot-2251297.json](https://github.com/paritytech/substrate-lite/blob/main/bin/polkadot-2251297.json) the finalized header is for block `2251297`, `current_epoch` is for `2248920` and `prev_epoch` is for `2246525`.

tomaka: It might be that `babe_epoch_changes` contains the block number that contains the information about the epoch, and not the block number where the epoch starts.

tomaka: In other words, `prev_epoch` (value `2246525`) contains the information about the epoch that starts at block `2248920`, and `current_epoch` (value `2248920`) starts around block `226000`.

At least that's my guess.

tomaka: Since you can't know in advance at which block number an epoch is going to start, this is indeed the only thing that makes sense.

src/chain_spec.rs
```diff
@@ -39,23 +39,102 @@ use alloc::string::String;
 mod light_sync_state;
 mod structs;
 
+use crate::chain::chain_information::{
+    BabeEpochInformation, ChainInformation, ChainInformationConsensus,
+};
+
```
tomaka, line 45: Please move this next to the other `use` above.

src/chain_spec.rs
```diff
@@ -39,23 +39,102 @@ use alloc::string::String;
...
+                    .grandpa_authority_set
+                    .current_authorities
+                    .iter()
+                    .map(|authority| crate::header::GrandpaAuthority {
+                        public_key: authority.public_key,
+                        weight: authority.weight,
+                    })
+                    .collect()
+            },
+            grandpa_finalized_scheduled_change: None,
```
tomaka, line 112:
```suggestion
            grandpa_finalized_scheduled_change: None,  // TODO: unimplemented
```

src/chain_spec.rs
```diff
@@ -39,23 +39,102 @@ use alloc::string::String;
...
+        epochs.sort_unstable_by_key(|(&block_num, _)| block_num);
+
+        // Get the latest two epochs.
+        let prev_epoch = &epochs[epochs.len() - 2].1;
+        let current_epoch = &epochs[epochs.len() - 1].1;
+
+        ChainInformation {
+            finalized_block_header: self.inner.finalized_block_header.clone(),
+            consensus: ChainInformationConsensus::Babe {
+                slots_per_epoch: std::num::NonZeroU64::new(prev_epoch.duration).unwrap(),
```
tomaka, line 93:
```suggestion
                slots_per_epoch: NonZeroU64::new(prev_epoch.duration).unwrap(),
```

`use core::num::NonZeroU64` above. We don't use `std`.

src/chain_spec.rs
```diff
@@ -34,28 +34,107 @@
...
+            .filter_map(|((_, block_num), epoch)| match epoch {
+                light_sync_state::PersistedEpoch::Regular(epoch) => Some((block_num, epoch)),
+                _ => None,
+            })
+            .collect();
+
+        epochs.sort_unstable_by_key(|(&block_num, _)| block_num);
+
+        // Get the latest two epochs.
+        let prev_epoch = &epochs[epochs.len() - 2].1;
```
tomaka, line 87:
```suggestion
        let current_epoch = &epochs[epochs.len() - 2].1;
```

src/chain_spec.rs
```diff
@@ -34,28 +34,107 @@
...
+        let current_epoch = &epochs[epochs.len() - 1].1;
```
tomaka, line 88:
```suggestion
        let next_epoch = &epochs[epochs.len() - 1].1;
```

src/chain_spec.rs
```diff
@@ -34,28 +34,107 @@
...
+
+impl LightSyncState {
+    pub fn as_chain_information(&self) -> ChainInformation {
+        // Create a sorted list of all regular epochs that haven't been pruned from the sync state.
+        let mut epochs: Vec<_> = self
+            .inner
+            .babe_epoch_changes
+            .epochs
+            .iter()
+            .filter_map(|((_, block_num), epoch)| match epoch {
```
tomaka, line 78:
```suggestion
            .filter(((_, block_num), _)| block_num <= self.inner.finalized_block_header.number)
            .filter_map(|((_, block_num), epoch)| match epoch {
```

tomaka (changes requested):

tomaka (approved):

PR #187, opened 2020-11-13
Add a no_std network service

tomaka:
Rather than duplicating code between the full node and wasm node, let's create a `network::Network` struct that acts as a state machine for the entire network, but in a `no_std` fashion.


tomaka:
Code is extremely dirty, but isn't much dirtier than right now.
Going to merge this PR and iterate on it.

Issue #195, opened 2020-11-15
Integrate Jaeger


tomaka:
One ambition of substrate-lite would be to connect to a [Jaeger server](https://www.jaegertracing.io/) and send telemetry information to it.

What we could send is:

- For each block:
  - [ ] A span indicating when the block has been authored, with a span per transaction applied and a span per storage access.
  - [ ] A span indicating when the block announces have been sent out.
  - [ ] A span indicating the incoming block requests, subdivided into sub-spans about the various stages of the block request.
  - [ ] For each node, a span indicating the block verification process, with a span per storage access.
  - [ ] For each node, when it gets written to the database.

- For each network connection:
  - [ ] A span when incoming data is processed or outgoing data is sent out.

- For each transaction: (I'm not familiar enough with transactions)
  - [ ] For each node, a span when its announces are sent out.
  - [ ] For each node, a span when its announce is received.
  - [ ] For each node, a span when it gets verified.
  - [ ] For each node, a span when it gets added to the transaction pool.
  - [ ] For each node, a span when it gets added to a block being authored.
  - [ ] For each node, a span when it gets removed from the transaction pool.

Now in order to have nice visualizations, the span IDs and trace IDs sent to Jaeger should be the same between the multiple nodes. Of course we don't want them to share IDs over the network like you would normally do in a traditional Jaeger setup.

However we can generate deterministic IDs based on for example the block hash for everything block-related and the PeerIds for everything connection-related.

melekes:
There's also [opentelemetry](https://opentelemetry.io/docs/instrumentation/rust/) which supports different "backends" (including Jaeger https://crates.io/crates/opentelemetry-jaeger)

tomaka:
During my initial research, I've tried opentelemetry and decided to not use it because it was clearly overengineered and didn't map 1:1 with Jaeger.

The protocol behind Jaeger is nothing more than a few UDP packets, and we clearly don't need a complicated framework to send a few UDP packets.

Issue #194, opened 2020-11-15
Make it produce blocks


tomaka:
Here's a small roadmap before a node can produce blocks:

- [x] Change a bit `sync::full_optimistic` to allowing access to the various caches regarding the best block (virtual machine, storage, merkle root calculation).
- [x] Add the code in `author` that claims a slot and seals the block (https://github.com/paritytech/substrate-lite/pull/247)
- [x] Add a transaction pool (#11)
- [x] Add a keystore (#154)
- [x] Finish the networking to allow to gossiping out the newly-created block.
- [x] Finish the networking to allow to answering incoming block requests about the newly-created block.
- [x] Change `sync_service.rs` to pause the sync'ing during a claimable slot and instead produce a new block.
- [ ] In the future, support offchain workers (#73). Because of the imonline mechanism, we'd get slashed without offchain workers.

melekes:
Can we close this?

tomaka:
Yeah, block production is still quite buggy but this is all covered by TODOs in the code and it doesn't make sense to leave an issue open without anything actionable in it.

PR #231, opened 2020-11-21
Add a jaeger client

tomaka:
cc #195 

Work in progress.
This PR is at the moment rather useless, because we don't want to report traces during major syncing (there isn't much point considering that nodes that are major syncing can be considered as paused), and major syncing is the only thing that is implemented at the moment.

As such, waiting for #194 to finish this PR.


tomaka:
This PR is dead, but I'm keeping it open in order to not lose the code changes.

PR #297, opened 2020-12-13
Move out dialing from libp2p::next_event()

tomaka:
Small re-architecturing.

Rather than returning a `Event::StartConnect` when a new TCP connection needs to be established, return `StartConnect` structs from a separate function.
This will allow in the future to give a more precise control to the user over the behaviour of dialing. The user will for example be able to limit the number of simultaneous pending connections (by not calling the function when over the limit), or add some rate limit to opening new connections. It also makes it easier to handle timeouts, as the moment the user calls the function is when the dialing starts.


tomaka:
This change actually causes the JS test to behave as expected and start connecting.
I'll adjust the test to quit after receiving the system name RPC response.

PR #304, opened 2020-12-15
Don't report notifications on non-open substreams

tomaka:


rakanalh (approved):
LGTM and does seem to fix my issue where the node panics receiving a best block announcement while the peer which sent it is not registered via `PeerConnected`.

PR #329, opened 2021-01-05
Implement syncing from all forks

### LINKED ISSUE #271: Implement head-of-chain syncing
[This todo](https://github.com/paritytech/substrate-lite/blob/c2d0ce021c7506293a3ca7d3de7a8d81ae19f524/src%2Fchain%2Fsync.rs#L51) must be tackled.

This should be done by duplicating `optimistic.rs` and modifying its logic in order to only download and sync individual blocks, similar to what `sync.rs` is doing in Substrate (`sc_network::protocol::sync`).

This is quite complicated, and I'm personally not familiar with all the details of how it works in Substrate nor of the challenges encountered.

One requirement is that all forks must be downloaded in order for finality to work. For example, if the best block of the local node is 5, and a node announces a block 4 that isn't the same as the block 4 known locally, this newly-announced block should be downloaded as well, because it's possible for validators to end up voting for finalizing this newly-announced block.

tomaka:
Fix #271 
This PR is a work-in-progress.

Adds the code making it possible to sync once at the head of the chain.
While the state machine that this PR adds is completely isolated, it is not really possible to actually integrate it at the moment. What's missing is a layer that uses the optimistic syncing and jumps to the "all-forks syncing" once near the head of the chain.

Additionally, the code in this PR doesn't handle finality. The light client should either listen for GrandPa messages, or periodically try to download justifications, but neither of these mechanisms is in scope of this code.


tomaka:
This works well enough now to be merge-able.

Before merging:

- [ ] Investigate why when we emit a request for 128 blocks we only get one back.
- [ ] Move the integration in a separate branch and restore `bin/wasm-node` as it is in master.

PR #336, opened 2021-01-07
Load finalized block entry on NewFinalized

rakanalh:
Similar to what we do on `NewBest` event, in `NewFinalized` we load the finalized hash to prevent LRU from evicting this newly-inserted unused value.


rakanalh:
For the record, this doesn't consistently work. I can see cases where the recently finalized block is returned and others where null is returned. 

I will close this PR and re-open if i have a better solution.

PR #361, opened 2021-01-15
Add Grandpa gossiping stub

tomaka:
After this PR, we now also open a Grandpa notifications substream.
Will be necessary in order to listen to Grandpa messages.

### DIFF
--- a/src/network/service.rs
+++ b/src/network/service.rs
@@ -157,6 +157,15 @@ where
                     in_slots: chain.in_slots,
                     out_slots: chain.out_slots,
                 }))
+                .chain(iter::once(libp2p::OverlayNetwork {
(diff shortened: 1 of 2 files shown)


src/network/service.rs
```diff
@@ -372,7 +381,7 @@ where
                     .await;
 
                 let peer_id = self.libp2p.connection_peer_id(*id).await;
-                let chain_index = overlay_network_index / 2;
+                let chain_index = overlay_network_index / 3;
```
rakanalh, line 384: Nitpick: isn't it better to have this as a constant defined in this file and used in all of those places?

tomaka: Indeed, I'll do this on top of this.

Issue #288, opened 2020-12-08
`network/libp2p.rs` is full of race conditions


tomaka:
The entire file should be reviewed, possibly re-written.

Deadlocks and race conditions frequently happen. There is also a futures-cancellation problem with `next_event()` that warrants a refactor.

tomaka:
I believe that all the bugs in this file can be fixed. In other words, the API itself isn't racy, just the implementation.

However, the current code design makes it super hard to review whether there's any bug. Given the complexity, any change in the future is very likely to introduce new bugs.
Therefore I'll go with refactoring the code design in order to simplify the logic.

tomaka:
Considered done in https://github.com/paritytech/smoldot/pull/985

PR #379, opened 2021-01-21
Considerably clean up src/libp2p.rs

tomaka:
cc #288 

I've clarified how the implementation is supposed to work, and made the appropriate changes.
It's still not fully finished, but it's considerable progress.


tomaka:
I'm merging quickly because I don't think anyone's going to review this.

PR #416, opened 2021-01-28
Add a babe_current_epoch function for calling `BabeApi_current_epoch`

expenses:


src/verify.rs
```diff
@@ -70,5 +70,6 @@ mod execute_block;
 
 pub mod aura;
 pub mod babe;
+pub mod babe_current_epoch;
```
tomaka, line 73: This module should belong to `src/chain/chain_information`.
Could you please rename the existing `chain_information::babe` module to `babe_config` (and same from `aura` to `aura_config` for consistency), and put it there instead?

tomaka:
:+1:  Looks mostly good!

I think `Babe_next_epoch` should be in the same module, with an extra field in `Config` to choose whether you want the current or next epoch. Feel free to do this now or in another PR.

src/verify/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+use crate::{
+    chain::chain_information::BabeEpochInformation,
+    executor::{host, read_only_runtime_host},
+    header,
+};
+
+use alloc::vec::Vec;
+use parity_scale_codec::{Decode, Encode};
+
+pub struct Config {
```
tomaka, line 27:
```suggestion
/// Configuration for [`babe_current_epoch`].
pub struct Config {
```

src/verify/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Runtime used to get the current babe epoch. Must be built using the Wasm code found at the
+    /// `:code` key of the parent block storage.
+    pub parent_runtime: host::HostVmPrototype,
+}
+
+#[derive(Debug, Clone, derive_more::Display)]
+pub enum Error {
```
tomaka, line 34:
```suggestion
/// Problem encountered during a call to [`babe_current_epoch`].
pub enum Error {
```

src/verify/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Error while starting the Wasm virtual machine.
+    #[display(fmt = "{}", _0)]
+    WasmStart(host::StartErr),
+    /// Error while running the Wasm virtual machine.
+    #[display(fmt = "{}", _0)]
+    WasmVm(read_only_runtime_host::Error),
+    /// Error while decoding the babe epoch.
```
tomaka, line 41:
```suggestion
    /// Error while decoding the Babe epoch.
```

src/verify/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+        parameter: std::iter::empty::<&[u8]>(),
+    });
+
+    match vm {
+        Ok(vm) => Verify::from_inner(vm),
+        Err(err) => Verify::Finished(Err(Error::WasmStart(err))),
+    }
+}
+
+/// Current state.
```
tomaka, line 61:
```suggestion
/// Current state of the operation.
```

src/verify/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+#[must_use]
+pub enum Verify {
```
tomaka, line 63:
```suggestion
pub enum Query {
```

It's not a verification. We're not verifying anything.

src/verify/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+    /// Fetching the babe epoch is over.
```
tomaka, line 64:
```suggestion
    /// Fetching the Babe epoch is over.
```

tomaka (changes requested):

src/chain/chain_information.rs
```diff
@@ -243,12 +246,12 @@ pub enum ChainInformationConsensus {
         /// Babe epoch information about the epoch right after the one the finalized block belongs
         /// to.
         ///
-        /// If [`ChainInformationConsensus::Babe::finalized_block_epoch_information`] is `Some`,
+        /// If [`ChainInformationConsensus::babe_config::finalized_block_epoch_information`] is `Some`,
```
tomaka, line 249:
```suggestion
        /// If [`ChainInformationConsensus::Babe::finalized_block_epoch_information`] is `Some`,
```

That's actually not supposed to pass CI!

src/chain/chain_information/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[display(fmt = "{}", _0)]
+    DecodeFailed(parity_scale_codec::Error),
+}
+
+/// Fetches the current Babe epoch using `BabeApi_current_epoch`.
+pub fn babe_current_epoch(config: Config) -> Query {
+    let vm = read_only_runtime_host::run(read_only_runtime_host::Config {
+        virtual_machine: config.parent_runtime,
+        function_to_call: "BabeApi_current_epoch",
+        // `BabeApi_current_epoch`doesn't take any parameters.
```
tomaka, line 53:
```suggestion
        // `BabeApi_current_epoch` doesn't take any parameters.
```

src/chain/chain_information/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+        parameter: std::iter::empty::<&[u8]>(),
```
tomaka, line 54:
```suggestion
        parameter: core::iter::empty::<&[u8]>(),
```

src/chain/chain_information/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+        Ok(vm) => Query::from_inner(vm),
+        Err(err) => Query::Finished(Err(Error::WasmStart(err))),
+    }
+}
+
+/// Current state of the operation.
+#[must_use]
+pub enum Query {
+    /// Fetching the Babe epoch is over.
+    Finished(Result<BabeEpochInformation, Error>),
```
tomaka, line 67: You also need to return the `HostVmPrototype`, as these are super expensive to create.

```suggestion
    Finished { result: Result<BabeEpochInformation, Error>, runtime: host::HostVmPrototype },
```

src/chain/chain_information/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+};
+
+use alloc::vec::Vec;
+use parity_scale_codec::{Decode, Encode};
+
+/// Configuration for [`babe_current_epoch`].
+pub struct Config {
+    /// Runtime used to get the current babe epoch. Must be built using the Wasm code found at the
+    /// `:code` key of the parent block storage.
+    pub parent_runtime: host::HostVmPrototype,
```
tomaka, line 31:
```suggestion
    pub runtime: host::HostVmPrototype,
```

src/chain/chain_information/babe_current_epoch.rs
```diff
@@ -0,0 +1,168 @@
...
+    header,
+};
+
+use alloc::vec::Vec;
+use parity_scale_codec::{Decode, Encode};
+
+/// Configuration for [`babe_current_epoch`].
+pub struct Config {
+    /// Runtime used to get the current babe epoch. Must be built using the Wasm code found at the
+    /// `:code` key of the parent block storage.
```
tomaka, line 30:
```suggestion
    /// `:code` key of the block storage.
```

tomaka (changes requested):
Another pass

src/chain/chain_information/babe_current_epoch.rs
expenses: We should change the return type of `Finished` in `read_only_runtime_host` then too.

expenses: At the moment we don't have access to the runtime in the case of `read_only_runtime_host::RuntimeHostVm::Finished(Err(err)) ` or `read_only_runtime_host::run` returning `Err` (which is impossible at present).

tomaka: Yes. It's really important to keep this object alive.

expenses: We can change `RuntimeHostVm::Finished` to:
```rust
Finished(Result<Success, (Error, host::HostVmPrototype)>)
```

by adding an `into_prototype()` function to `host::HostVm` variants, including forbidden ones and calling that.

Getting `read_only_runtime_host::run` to return the prototype on `Err` is harder though, as it would require returning a `HostVmPrototype` in `HostVmPrototype::run_vectored` and that requires going into the individual VM implementations and getting them to return `Self` on error as well.

tomaka (approved):
I applied the two little fixes that you missed

tomaka:
Feel free to put the `automerge` label to merge

PR #425, opened 2021-01-29
Don't panic if runtime version is asked when we have no peers

tomaka:
Right now, querying the runtime version does a storage query on the network. If the storage query fails, we simply panic.
The correct way to do is to maintain a version of the runtime of the best block.
Since it's not trivial to do that, let's add another quick hack and fall back to returning the genesis runtime specs.

Should not be merged yet, because querying the genesis runtime specs fails because of some format mismatch. I'm opening another PR for that.


tomaka:
Opened https://github.com/paritytech/smoldot/pull/426 instead

PR #449, opened 2021-02-02
Babe information fetching is wrong format

tomaka:
cc @expenses 

This PR is more of a "bug report".
The `DecodableBabeEpochInformation` struct has the wrong format.


tomaka:
Since the configuration is missing from the returned value, I think you also have to call `BabeApi_configuration`, already implemented in the `chain_information::babe_config` module.
At the moment the Babe configuration never changes in practice after a chain is live, but in theory it can. Opening an issue to fix that later is fine by me.

src/chain/chain_information/babe_fetch_epoch.rs
```diff
@@ -106,8 +106,8 @@ impl Query {
                                 })
                                 .collect(),
                             randomness: epoch.randomness,
-                            c: epoch.c,
-                            allowed_slots: epoch.allowed_slots,
+                            c:  todo!(),//epoch.c,
+                            allowed_slots: todo!(),// epoch.allowed_slots,
```
expenses, line 110: Where do I get this information from then?

tomaka: My comment above: call `BabeApi_configuration` (see `chain_information::babe_config` module)

PR #468, opened 2021-02-05
Issue multiple warp sync requests in grandpa warp sync if needed

expenses:
... And other warp sync changes


tomaka:
Instead of a hardcoded limit of 100, is it not possible to continue issuing requests until it comes back empty?

expenses:
> Instead of a hardcoded limit of 100, is it not possible to continue issuing requests until it comes back empty?

I'm not sure, it might be that the minimum number of requests returned is 1. I would much rather send the fragments along with a boolean that indicates whether more requests need to be made or not though.

tomaka:
The boolean would indeed remove a round trip time.

> I'm not sure, it might be that the minimum number of requests returned is 1.

Note that we must be resilient to that kind of thing if the Substrate node misbehaves (intentionally or not). Smoldot shouldn't be trapped in an infinite loop, whatever what is received is.

expenses:
>  I'm not sure, it might be that the minimum number of requests returned is 1.

I just confirmed this by setting `let final_set_of_fragments = response_fragments.len() == 1;`.

> Note that we must be resilient to that kind of thing if the Substrate node misbehaves (intentionally or not). Smoldot shouldn't be trapped in an infinite loop, whatever what is received is.

I think that having the check above would eliminate most problems.

tomaka:
Is it not possible for the target Substrate node to randomly generate one Grandpa warp sync fragment (with a different random hash every time) and block smoldot forever?

expenses:
> Is it not possible for the target Substrate node to randomly generate one Grandpa warp sync fragment (with a different random hash every time) and block smoldot forever?

I don't think verification would succeed in this case. And as `final_set_of_fragments` would be true, we would only attempt to validate once. What _might_ block smoldot is if a node sent the same (valid) fragment 100 times. I'm not sure what happens with verification then.

tomaka:
Right, sorry. I somehow had in my mind that we were downloading everything first, and then only start the verification.

src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -155,6 +157,11 @@ impl<TSrc> StorageGet<TSrc> {
         &self.state.warp_sync_source
     }
 
+    /// Returns the header that we're warp syncing up to.
+    pub fn warp_sync_header(&self) -> Header {
+        self.state.header.clone()
+    }
```
tomaka, line 163: Let's please not just copy things all the time.

```suggestion
    pub fn warp_sync_header(&self) -> HeaderRef {
        (&self.state.header).into()
    }
```

src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -194,6 +201,11 @@ impl<TSrc> NextKey<TSrc> {
...
```
tomaka, line 207: Same

src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -340,6 +402,11 @@ pub struct VirtualMachineParamsGet<TSrc> {
 }
 
 impl<TSrc> VirtualMachineParamsGet<TSrc> {
+    /// Returns the header that we're warp syncing up to.
+    pub fn warp_sync_header(&self) -> Header {
+        self.state.header.clone()
+    }
```
tomaka, line 408: Ditto

src/finality/grandpa/warp_sync.rs
```diff
@@ -32,10 +32,10 @@ pub struct Verifier {
 
 impl Verifier {
     pub fn new(
-        start_chain_information: &ChainInformation,
+        start_chain_information_finality: &ChainInformationFinality,
```
tomaka, line 35:
```suggestion
        start_chain_information_finality: ChainInformationFinalityRef,
```

tomaka (changes requested):
Good after removing the extra copies

tomaka (approved):

PR #469, opened 2021-02-05
Document the services of the Wasm node

tomaka:


raoulmillais:
Awesome thank you :+1:   I've finally started reading through the implementation of the WASM node in the bin folder.  Any extra documentation is really helpful

PR #472, opened 2021-02-06
Add AllSync syncing

tomaka:
Combines all syncing strategies (optimistic, grandpa warp sync, all forks) into one, and switches the `sync_service` to using it.

Work in progress at the moment.


tomaka:
I'm going to focus on optimistic and all-heads syncing at the moment, as GrandPa warp sync needs some fixing on the Substrate side.

tomaka:
This is now fully working in the normal case (no malicious nodes).
Since it makes things strictly better than they are now, I'm going to merge this.

I went into "technical debt mode" however to finish this (i.e. write code as quickly as possible, not documenting) which I'll work on "reimbursing" next.

src/chain/sync/optimistic.rs
```diff
@@ -910,6 +949,9 @@ impl<TRq, TSrc, TBl> ProcessOne<TRq, TSrc, TBl> {
                         debug_assert!(chain.is_empty());
                         shared.inner.best_to_finalized_storage_diff.clear();
 
+                        shared.inner.finalized_chain_information.chain_information =
+                            chain.as_chain_information().into();
```
tomaka, line 953: Note: this is a bugfix. The variable was never actually updated.

PR #479, opened 2021-02-08
Fix `babe_fetch_epoch`

expenses:
Closes #449. Fixes the structure of `DecodableBabeEpochInformation` and passes in a chain spec to `GrandpaWarpSync` in order to get the babe genesis configuration.

### DIFF
--- a/src/chain/chain_information/babe_fetch_epoch.rs
+++ b/src/chain/chain_information/babe_fetch_epoch.rs
@@ -16,11 +16,7 @@
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use crate::{
-    chain::chain_information::{
-        babe_config::{BabeGenesisConfiguration, FromVmPrototypeError},
-        BabeEpochInformation,
-    },
-    chain_spec::ChainSpec,
+    chain::chain_information::BabeEpochInformation,
     executor::{host, read_only_runtime_host},
     header,
 };
@@ -43,12 +39,12 @@ pub struct Config {
     pub runtime: host::HostVmPrototype,
     /// The Babe epoch to fetch.
     pub epoch_to_fetch: BabeEpochToFetch,
-    /// The chain spec of the chain.
-    pub chain_spec: ChainSpec,
+    /// The Babe genesis config.
+    pub babe_genesis_config: header::BabeNextConfig,
 }
 
 /// Problem encountered during a call to [`babe_fetch_epoch`].
-#[derive(Debug, derive_more::Display)]
+#[derive(Debug, Clone, derive_more::Display)]
 pub enum Error {
     /// Error while starting the Wasm virtual machine.
     #[display(fmt = "{}", _0)]
@@ -59,7 +55,6 @@ pub enum Error {
     /// Error while decoding the babe epoch.
     #[display(fmt = "{}", _0)]
     DecodeFailed(parity_scale_codec::Error),
-    BabeGenesisConfigurationFromVm(FromVmPrototypeError),
 }
 
 /// Fetches a Babe epoch using `BabeApi_current_epoch` or `BabeApi_next_epoch`.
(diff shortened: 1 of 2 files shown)


src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -52,12 +56,17 @@ pub struct Config {
     pub start_chain_information: ChainInformation,
     /// The initial capacity of the list of sources.
     pub sources_capacity: usize,
+    /// The chain spec of the chain.
+    pub chain_spec: ChainSpec,
```
tomaka, line 60: You can know the Babe configuration from the `ChainInformation`, no need to additionally pass the `ChainSpec`.

src/chain/chain_information/babe_fetch_epoch.rs
```diff
@@ -39,6 +39,8 @@ pub struct Config {
     pub runtime: host::HostVmPrototype,
     /// The Babe epoch to fetch.
     pub epoch_to_fetch: BabeEpochToFetch,
+    /// The Babe genesis config.
+    pub babe_genesis_config: header::BabeNextConfig,
```
tomaka, line 43: The objective of the `babe_fetch_epoch` module is to do a runtime call.
You're passing a `BabeNextConfig` here just so that you can later return back what it contains.

Please define a new `EpochInfo` struct in this module and return it, instead of doing that.

```suggestion
```

tomaka (changes requested):
It's very important to actually document the fact that we use the same configuration as the genesis because doing so is a total hack. I really want to avoid ambiguities in which someone looks at this code and thinks that this is how you are supposed to do

src/chain/chain_information/babe_fetch_epoch.rs
```diff
@@ -72,27 +70,50 @@ pub fn babe_fetch_epoch(config: Config) -> Query {
...
         Err(err) => Query::Finished(Err(Error::WasmStart(err))),
     }
 }
 
+/// Partial information about a Babe epoch.
+pub struct PartialBabeEpochInformation {
+    epoch_index: u64,
+    start_slot_number: Option<u64>,
+    authorities: Vec<header::BabeAuthority>,
+    randomness: [u8; 32],
```
tomaka, line 83:
```suggestion
    pub epoch_index: u64,
    pub start_slot_number: Option<u64>,
    pub authorities: Vec<header::BabeAuthority>,
    pub randomness: [u8; 32],
```

src/chain/chain_information/babe_fetch_epoch.rs
```diff
@@ -72,27 +70,50 @@ pub fn babe_fetch_epoch(config: Config) -> Query {
...
+        BabeEpochInformation {
+            epoch_index: self.epoch_index,
+            start_slot_number: self.start_slot_number,
+            authorities: self.authorities,
+            randomness: self.randomness,
+            c,
+            allowed_slots,
+        }
+    }
+}
```
tomaka, line 102: Please do this in `grandpa_warp_sync.rs`.
This module shouldn't touch `BabeEpochInformation` at all. It's objective is purely and solely to do a runtime call.

tomaka (changes requested):

src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -84,25 +87,46 @@ pub enum GrandpaWarpSync<TSrc> {
...
+                let (slots_per_epoch, babe_config_c, babe_config_allowed_slots) =
+                    match state.start_chain_information.consensus {
+                        ChainInformationConsensus::Babe {
+                            slots_per_epoch,
+                            finalized_next_epoch_transition,
+                            ..
+                        } => (
+                            slots_per_epoch,
+                            finalized_next_epoch_transition.c,
+                            finalized_next_epoch_transition.allowed_slots,
```
tomaka, line 104:
```suggestion
                            // TODO: /!\ /!\ shouldn't take the same configuration as the genesis; this is a hack while waiting for https://github.com/paritytech/substrate/issues/8060
                            finalized_next_epoch_transition.c,
                            finalized_next_epoch_transition.allowed_slots,
```

tomaka: Please document or something

tomaka: Basically, someone else than you and me should be able to read the code and figure out how it works

tomaka (approved):
Ideally would have better documentation

PR #483, opened 2021-02-08
Add a grandpa warp sync service to the wasm node

expenses:
We can finally tie everything together now! 🚀


tomaka:
Well, we're not supposed to have a separate GrandPa warp sync service. It should be handled by the sync service. :grimacing:

expenses:
> Well, we're not supposed to have a separate GrandPa warp sync service. It should be handled by the sync service.

How would you like me to integrate it with that?

tomaka:
It's basically #472

expenses:
Ah, okay. I'll close this for now.

tomaka:
I'd appreciate the network function alone though.

PR #495, opened 2021-02-10
Add typescript typings for javascript package

raoulmillais:
This typings file means people using the smoldot client from typescript get the type information without having to write their own definitions.  The trade-off is that any future changes to the JavaScript API should now be accompanied with a corresponding update to the typings file.

I have tested by npm linking my local clone of smoldot to my clone of substrate-connect and removing the typings I created there.  It compiled fine and passed all the tests.

### DIFF
--- a/bin/wasm-node/javascript/index.d.ts
+++ b/bin/wasm-node/javascript/index.d.ts
@@ -1,7 +1,20 @@
-// Copyright 2018-2021 @paritytech/substrate-light-ui authors & contributors
-// This software may be modified and distributed under the terms
-// of the Apache-2.0 license. See the LICENSE file for details.
+// Smoldot
+// Copyright (C) 2019-2021  Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
 
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
 declare class SmoldotError extends Error {
   constructor(message: string) {}
 }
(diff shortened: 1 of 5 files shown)


bin/wasm-node/javascript/index.d.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018-2021 @paritytech/substrate-light-ui authors & contributors
```
raoulmillais, line 1: oops i'll fix this.

raoulmillais:
Maybe worth mentioning that we will have at least 3 packages depending on smoldot that use typescript.  If the typescript file does not live upstream (here) then we have to duplicate it across all our packages (or create a meta-package with just the definitions), which seems undesirable.

bin/wasm-node/javascript/index.d.ts
```diff
@@ -0,0 +1,43 @@
...
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+//
```
tomaka, line 17:
```suggestion

```

tomaka:
Do you know if there's a way to check through CI if the typing file matches the actual file?

raoulmillais:
> Do you know if there's a way to check through CI if the typing file matches the actual file?

I think they do this using dtslint in the official typings repository: https://github.com/DefinitelyTyped/DefinitelyTyped#my-package-teststs but I haven't investigated it properly

raoulmillais:
> > Do you know if there's a way to check through CI if the typing file matches the actual file?
> 
> I think they do this using dtslint in the official typings repository: https://github.com/DefinitelyTyped/DefinitelyTyped#my-package-teststs but I haven't investigated it properly

Yeah here are the docs: https://github.com/microsoft/dtslint  looks like you write a test file that exercises the API and add some comments about what typings to expect.  Do we want the CI to actually call `smoldot.start` though?

tomaka:
If you could just change [the test command](https://github.com/paritytech/smoldot/blob/4ccbe4a37b4accce9f890aef693623d7b6f19da1/bin/wasm-node/javascript/package.json#L22) to do `"node prepare.js && dtslint types && node test.js"` (and add `dtslint` as a dev-dependency).

raoulmillais:
Slightly different from what you said @tomaka but it achieves the goal of running a lint of the typings file

tomaka (approved):
That's fine by me provided you make CI pass

raoulmillais:
> That's fine by me provided you make CI pass

This weird we don't use `tsd` directly `dtslint` does - adding it makes CI happy

PR #521, opened 2021-02-16
Allow exporting events to an SQLite database

tomaka:
Adds a new feature to the full node.
A new `--major_sync_export_events_sqlite` flag can now indicate the path to an SQLite database where the events that happen on the chain will be stored while the node is syncing.

The schema of this SQLite database is the following:

```sql
CREATE TABLE IF NOT EXISTS metadata(
    runtime_version INTEGER NOT NULL PRIMARY KEY,
    metadata BLOB NOT NULL
);

CREATE TABLE IF NOT EXISTS events(
    block_height INTEGER NOT NULL PRIMARY KEY,
    runtime_version INTEGER NOT NULL,
    events_storage BLOB NOT NULL,
    FOREIGN KEY(runtime_version) REFERENCES metadata(runtime_version)
);
```

The writing is done in parallel of the syncing, meaning that no block can be skipped.

Before this feature can be really usable, I'd like to do another change and allow not storing old blocks in the database. Currently the main blocker for the syncing speed is the disk I/O (at least on my local machine), and it would be a good idea in my opinion to improve it by not storing everything.

Additionally,  this PR adds an unfinished `smoldot-events-export-parser` JavaScript crate that can analyse this SQLite database. The code is very much WIP, as it's unclear exactly which API to expose. The code is working well, just not finished.


tomaka:
Closing in favour of https://github.com/tomaka/polkadot-events

Issue #531, opened 2021-02-17
Metadata_metadata might actually access the storage


tomaka:
Right now [smoldot forbids any external function when querying the metadata](https://github.com/paritytech/smoldot/blob/227f767a795b14fe124fb49c82b0861146c46dce/src/metadata/query.rs#L85). It turns out that Rococo's runtime accesses the storage.

tomaka:
I asked Basti, and it's apparently expected for Rococo because of a hack they use to bypass some Rust limitation.
The metadata still is never supposed to change.

In an ideal world, we would track the storage values of the keys requested by the metadata function, and refresh the metadata when one of these values change. However the JSON-RPC layer doesn't support pro-actively pushing metadata changes.

We should either precisely document this shortcoming, or actually track these storage keys and print a warning if their value changes. The latter seems very difficult, so I'm fine with the former.

PR #532, opened 2021-02-17
Add Rococo and Tick chain specs, plus various debugging improvements

tomaka:
Erin has spawned a Rococo and a Tick connect node.


src/libp2p/connection/multistream_select.rs
```diff
@@ -168,9 +168,9 @@ where
         // legitimately requests a protocol that the listener doesn't support. In order to prevent
         // confusion, a minimum length is applied to the protocol name length. Any protocol name
         // smaller than this will never trigger a protocol error, even if it isn't supported.
-        const MIN_PROTO_LEN_NO_ERR: usize = 32;
+        const MIN_PROTO_LEN_NO_ERR: usize = 48;
         let max_frame_len = cmp::max(
-            cmp::min(max_proto_name_len, MIN_PROTO_LEN_NO_ERR),
+            cmp::max(max_proto_name_len, MIN_PROTO_LEN_NO_ERR),
```
tomaka, line 173: This is a bugfix.

tomaka:
Rococo doesn't work because of #531 
Tick doesn't work because of some protocol id issue which I've yet to investigate.

tomaka:
Note keeping: Tick doesn't work because we try to open a GrandPa substream and there's no GrandPa running on Tick.

PR #544, opened 2021-02-21
Change the source system of GP warp sync

tomaka:
There was some misunderstanding in my review of the Grandpa warp sync PR.

The Grandpa warp sync state machine should contain a list of sources. The `TSrc` parameter, however, shouldn't be used as an identifier for a source, but rather as some "extra data" that the user can associate to each source in the list. Instead, an identifier for a source is now a `SourceId` attributed by the GP warp sync state machine itself.
The list of sources is stored in a `Slab` instead of a `Vec`, so that indices don't change even if you remove a source.

While the current code works, this change would be appreciated in order to integrate it with the "all syncing" system.
The "all syncing" system is going to use the `TSrc` object to store for example the best block of the source, which is an information necessary after the end of the Grandpa warp sync verification.

### DIFF
--- a/src/chain/sync/grandpa_warp_sync.rs
+++ b/src/chain/sync/grandpa_warp_sync.rs
@@ -362,7 +362,6 @@ impl<TSrc> WarpSyncRequest<TSrc> {
             let removed = self.sources.remove(to_remove.0).user_data;
 
             let next_state = if let Some(next_id) = next_id {
-                self.sources[next_id.0].already_tried = true;
                 GrandpaWarpSync::WarpSyncRequest(Self {
                     source_id: next_id,
                     sources: self.sources,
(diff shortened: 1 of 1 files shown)


src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -340,44 +351,42 @@ impl<TSrc: PartialEq> WarpSyncRequest<TSrc> {
...
+                .iter()
+                .find(|(_, s)| !s.already_tried)
+                .map(|(id, _)| SourceId(id));
 
-            if next_index == self.sources.len() {
-                GrandpaWarpSync::WaitingForSources(WaitingForSources {
+            let removed = self.sources.remove(to_remove.0).user_data;
+
+            let next_state = if let Some(next_id) = next_id {
+                self.sources[next_id.0].already_tried = true;
```
expenses, line 365: Could we set `already_tried` to `true` in `handle_response` instead?

expenses (approved):
Looks good to me apart from my 1 comment

PR #545, opened 2021-02-21
Allow raw TCP connections from NodeJS

### LINKED ISSUE #219: Consider supporting plain TCP connections when using NodeJS
While browsers only allow WebSockets, NodeJS on the other hand could connect to full nodes that aren't listening for WebSocket connections.

tomaka:
Fix #219 

Work in progress. I'm having issues on the JS side.


tomaka:
Successfully tested on both NodeJS and the browser.

Issue #221, opened 2020-11-19
Make it sync parachains


tomaka:
- [x] Make it sync Rococo.
- [ ] Make it sync Tick (or Trick or Track)

tomaka:
Overview of how parachains work:

- On the parachain, collator nodes produce blocks and send them to the relay chain validator nodes, where they are verified. At the moment there is no restriction on who can actually produce blocks on the parachain. In other words it is possible for hundreds of nodes to produce blocks and create hundreds of forks.
- The relay chain then stores, in its storage, the best block (according to the relay chain validators) of the parachain. The `persisted_validation_data` runtime entry point can be called on the relay chain in order to obtain the parachain heads. See [this](https://github.com/paritytech/polkadot/blob/ae69d7a0dfe7c34b600df23fbfe3f5e0104d84ba/roadmap/implementers-guide/src/runtime-api/persisted-validation-data.md).
- The current best block of the parachain is equal to the head stored in the current best block of the relay chain.
- The latest finalized block of the parachain is equal to the head stored in the latest finalized block of the relay chain.

In terms of changes to make in substrate-lite:

- [ ] Finish the networking to make it possible to connect to multiple chains at once.
- [ ] (for light clients) Implement the `RemoteCallRequest` light client request, in order to be able to call `persisted_validation_data` on the relay chain without having to store the relay chain storage.
- [ ] Add a "FreeForAll" consensus algorithm where the block headers check always passes (and thus all blocks are accepted assuming they execute correctly). We might however still have to check the inherents, something that is still unclear to me at the moment, cc #227.
- [ ] Add a "RelayChain"/"Manual"/"External" finality algorithm.
- [ ] Add a new parachains syncing system in `sync` that emits requests for both the relay chain and the parachain, and that is responsible for calling `persisted_validation_data`. How to organize the code in the `sync` module is however completely unclear at the moment, as the code is presently very messy.

PR #549, opened 2021-02-22
Draft for how to sync parachains

tomaka:
cc #221

Work in progress.

The problem I'm facing at the moment is that we need to be able to call runtime functions on the relay chain, just like the JSON-RPC service, but of course we can't do it at the same time as the JSON-RPC service.
Needs a bit of brainstorming.


tomaka:
I've extracted over time all the components of this PR into individual pull requests.
This PR will ultimately be closed, sooner or later.

PR #555, opened 2021-02-23
Integrate GrandPa warp sync in the all-syncing

tomaka:
This is some work towards integrating GrandPa warp sync with the "all syncing system".

I've added some methods the GrandPa warp sync state machine.

After this PR, we now try to do one request towards each node, then freeze forever because nobody is able to serve us.
(on that topic: I don't actually know whether we should add some sort of timeout, and switch to optimistic syncing afterwards, or if just freezing forever is fine because we assume that GrandPa warp sync is widely available)

I'll wait for GrandPa warp sync to be available to finish the integration.

### DIFF
--- a/bin/wasm-node/rust/src/sync_service.rs
+++ b/bin/wasm-node/rust/src/sync_service.rs
@@ -198,6 +198,8 @@ async fn start_sync(
 
         // List of block requests currently in progress.
         let mut pending_block_requests = stream::FuturesUnordered::new();
+        // List of grandpa warp sync requests currently in progress.
+        let mut pending_grandpa_requests = stream::FuturesUnordered::new();
 
         // TODO: remove; should store the aborthandle in the TRq user data instead
         let mut pending_requests = HashMap::new();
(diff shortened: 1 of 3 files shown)


bin/wasm-node/rust/src/sync_service.rs
```diff
@@ -515,6 +530,27 @@ async fn start_sync(
...
+                    // `result` is an error if the block request got cancelled by the sync state
+                    // machine.
+                    if let Ok(result) = result {
+                        // Inject the result of the request into the sync state machine.
+                        let new_sync = sync_idle.grandpa_warp_sync_response(
+                            request_id,
+                            result.ok(),
+                        );
+
+                        sync = new_sync;
```
expenses, line 547: I think we want to handle `Err(..)` warp sync responses somewhere, either here or in `sync/all`. Do you have a preference on where?

expenses (approved):
I'm not super familiar with all syncing code, but the grandpa stuff looks good to me.

bin/wasm-node/rust/src/sync_service.rs
tomaka: Errors should directly be handled in the GrandPa warp sync state machine. If the verification fails, the GrandPa warp sync code should switch back to `WaitingForSources` or `WarpSyncRequest` to try do another request.

expenses: I mean we already do that with the `Option` that we pass in, but I'm talking more about disconnecting or giving a bad reputation to nodes that give out bad responses. That seems out of scope for grandpa warp sync.

tomaka: > I mean we already do that with the Option that we pass in,

The `Option` covers the situation where someone sends us random garbage data. But if someone sends us for example an incomplete list of signatures, we will jump to the `Verifier` state and fail. In that situation, we should jump back to `WaitingForSources` or `WarpSyncRequest`, and not just return an error and kill the entire GrandPa warp sync state machine.

> but I'm talking more about disconnecting or giving a bad reputation to nodes that give out bad responses. That seems out of scope for grandpa warp sync.

I don't think we need to disconnect or maintain reputations, but for what it's worth the `grandpa_warp_sync_response` function will eventually return an enum indicating what happened after we have received the response (and which will notably contain a follow-up request if relevant).

expenses: > the `grandpa_warp_sync_response` function will eventually return an enum indicating what happened after we have received the response (and which will notably contain a follow-up request if relevant).

Ah, that's a good idea!

PR #560, opened 2021-02-24
Apply changes to warp sync with andre's branch

expenses:


src/network/protocol/grandpa_warp_sync.rs
```diff
@@ -49,7 +49,6 @@ pub fn decode_grandpa_warp_sync_response(
                                 ))
                             })
                     },
-                    crate::util::nom_scale_compact_usize,
```
expenses, line 52: Do we want to request that this value is kept in the response?

tomaka: I said in DM that think it made more sense to keep it. However, after thinking twice about it, I think that this length prefix indeed doesn't really make sense. After all, we can't do much with the response is we can't know the format of a justification.
Plus, this protocol is pretty tied to the Grandpa algorithm. If we implemented another finality algorithm in Substrate, it would likely have to use a different protocol. Therefore we don't need to take into account the fact that justifications might in fact not be Grandpa justifications.

src/chain/sync/grandpa_warp_sync.rs
```diff
@@ -428,13 +428,26 @@ impl<TSrc> WarpSyncRequest<TSrc> {
...
+                            header,
+                            chain_information_finality,
+                            start_chain_information: self.state.start_chain_information,
+                            warp_sync_source: self.sources.remove(self.source_id.0).user_data,
+                        },
+                    })
+                }
+                // Count a response of 0 fragments as a failed response if we haven't previously
+                // received any fragments.
+                None => response = None,
```
tomaka, line 449: What if the `chain_information` you pass at initialization is already the latest finalized block? This will be considered a failure.

tomaka (approved):

PR #562, opened 2021-02-24
Transaction service notifications

rakanalh:
This is a rough implementation of the remaining notifications in the transaction service:

1. Best/Finalized blocks are downloaded.
2. InBlock is sent as soon as a pending transaction is found in a downloaded block.
3. Retracted is sent once a transaction is found in a block where it was previously found in another.
4. Finalized is sent as soon as a finalized block contains a pending transaction. This transaction is then removed from `pending_transactions`.

Moving forward: 
1. Each transaction will know the height of the finalized block at which it was received. Once passing a specific threshold (`current_finalized - transaction.finalized` mod X == 0`) a `Broadcast` notification will be sent out again.
2. A maximum number of broadcasts will be made which will then trigger `Dropped`
3. Each transaction will know the block at which it was found... if no finalization for the block is received, `FinalizedTimeout` will be sent out.


src/header.rs
```diff
@@ -1218,3 +1218,8 @@ fn decode_item_from_parts<'a>(
         _ => unreachable!(),
     })
 }
+
+/// Decodes a SCALE-encoded transaction bytes.
+pub fn transaction_decode(mut bytes: &[u8]) -> Result<Vec<u8>, parity_scale_codec::Error> {
```
rakanalh, line 1223: parity_scale_codec is not part of the `wasm-node` deps and i didn't want it to be. Therefore, added this fn temporarily here... a suggestion of a more suitable for this to live is appreciated.

PR #567, opened 2021-02-25
Finish integration of GrandPa warp sync into all syncing

tomaka:
Couldn't test if it works because at the moment, on Westend, the GrandPa warp sync proof jumps to a block that is already pruned.
I'm opening this PR now, but I obviously want to test it before merging.

The code in `all.rs` is kind of ugly, but at least it's mostly there now.

Note: to test this, you need to tweak the Westend specs to point to the westend-connect nodes.


tomaka:
Successfully worked once. Now getting panics in slab.

However, before merging, we should first add some sort of fallback to optimistic syncing for when the warp syncing doesn't work.

tomaka:
I've restored the `if true ||` that force-disables GP warp sync. This is now ready for merging/reviewing.

tomaka:
Getting panics [here](https://github.com/paritytech/smoldot/blob/6ee1f18a96f3abc553560d43d0988982449fcc21/src/chain/sync/all_forks/pending_blocks.rs#L191) right after the transition to all-forks syncing.
Still haven't figured out what happens.

expenses:
So I don't have the background knowledge to properly review a reasonable chunk of this PR. I can't see any problems with the grandpa stuff though.

expenses (approved):

expenses:
Let's just merge this.

PR #570, opened 2021-02-26
Replace ed25519-dalek with ed25519-zebra

### LINKED ISSUE #53: Don't use batched ed25519 signatures verification
I didn't realize that they were broken.

tomaka:
cc https://github.com/paritytech/substrate/issues/8055

fix #53


tomaka:
As far as I understand, updating smoldot alone could cause consensus issues, as a validator could generate blocks that are valid according to smoldot but not to Substrate.

I think it's fine as long as smoldot isn't used for producing blocks and I'll just open a tracking issue.

PR #573, opened 2021-02-26
Move fields from GrandpaWarpSync into InProgressGrandpaWarpSync

expenses:
As discussed in DMs, this would make implementing the 'all' syncing behaviour easier. I'm not sure if we want to do this incrementally or go all-in with the idea that `Finished` should always return `Success` and we just try a different peer when something goes wrong.


expenses:
I haven't touched `all.rs` yet

tomaka:
I'd rather combine this with #555 somehow, to avoid the merge conflicts.

expenses:
> I'd rather combine this with #555 somehow, to avoid the merge conflicts.

you mean https://github.com/paritytech/smoldot/pull/567 ? I can try merge it with that.

tomaka:
#567 is only waiting for a review from you to be merged :grimacing:

expenses:
Is this going to merge without a review? I'd rather do #567 first.

tomaka (approved):

tomaka:
Next stop could be to add some utility methods on the `InProgress`, like `source_user_data_mut`, and simplify `all.rs` to use them.

PR #575, opened 2021-02-26
Add `source_user_data_mut` and `as_chain_information` to `InProgressGrandpaWarpSync`

expenses:
- Should I remove these functions from the respective variants? So that there's only the `source_user_data_mut` function implemented on `InProgressGrandpaWarpSync` and not `StorageGet` etc..
- Any other functions that I should add?


tomaka:
> Should I remove these functions from the respective variants? So that there's only the source_user_data_mut function implemented on InProgressGrandpaWarpSync and not StorageGet etc..

I think that it in theory each variant should have its own version of the functions. But since they bloat the module quite a bit, I'm more in favour of removing them. They're not exactly difficult to put back if necessary.

tomaka (approved):

tomaka (approved):

PR #576, opened 2021-02-26
Allow querying all keys with a prefix through storage queries

### LINKED ISSUE #523: Properly implement state_getKeysPaged
This function is implemented in a dummy way, and the list of validators visible on PolkadotJS is most likely wrong.

tomaka:
Fix #523 

I've implemented the verification part of the code, but remains the do the actual querying.


tomaka:
I'm utterly confused by the CI failure. The failing line isn't in the code anymore. The thing compiles locally.

PR #582, opened 2021-02-28
Change the way the optimistic syncing handles the runtime

tomaka:
Before: the optimistic sync queries for `:code` and `:heappages` from the API user, then internally compiles the runtime. In case of a runtime upgrade, we demand these values again from the user at the next block.

After: the optimistic sync must be passed the finalized block runtime at initialization. In case of a runtime upgrade, the new runtime overrides the old one.

This is possible thanks to https://github.com/paritytech/smoldot/pull/347


tomaka:
Opening as draft because it doesn't work. State mismatch at Polkadot block 29232 (block 29231 contains a runtime upgrade).

PR #591, opened 2021-03-01
Add a runtime_service that fetches the best block runtime

tomaka:
Before this PR, the JSON-RPC service spawns a background task that downloads from the network the runtime at a regular interval when the best block changes.
It then uses this runtime to feed the metadata to PolkadotJS, to handle runtime spec changes subscrptions, and to perform calls such as getting a nonce.

In order to make parachains work (cc #221) we need to be able to access the runtime of the relay chain.
This PR extracts this background task (that downloads the runtime) from the JSON-RPC service, and puts it in a new "runtime service".
In addition to cleaning up the design, the intention is that in the future the parachain syncing code will use the runtime service to query the parachain heads.

The PR is finished, but I'm opening as a draft because it's untested. I would like to test whether subscribing to runtime upgrades for example still works.


tomaka:
Can't actually test whether runtime change subscriptions work since all blocks that contain an older runtime are pruned.

Issue #301, opened 2020-12-14
Backport `Make it possible to calculate the storage root as often as you want`


tomaka:
https://github.com/paritytech/substrate/pull/7714

tomaka:
After looking at the code, I don't see a reason why this wouldn't work. Closing as done.

PR #608, opened 2021-03-03
Allow getting storage root in read-only runtime call

tomaka:
cc #221 #301

Allows obtaining the storage trie root from within the runtime during a read-only call.
Adds a `RuntimeHostVm::StorageRoot` variant that requires passing the state trie root in order to resume execution.

I've had to turn `GrandpaWarpSync::from_babe_fetch_epoch_query` into a loop in order to allow resuming execution instantly.

### DIFF
--- a/bin/wasm-node/rust/src/runtime_service.rs
+++ b/bin/wasm-node/rust/src/runtime_service.rs
@@ -310,6 +310,9 @@ impl RuntimeService {
                     executor::read_only_runtime_host::RuntimeHostVm::NextKey(_) => {
                         todo!() // TODO:
                     }
+                    executor::read_only_runtime_host::RuntimeHostVm::StorageRoot(storage_root) => {
+                        runtime_call = storage_root.resume(&runtime_block_state_root);
+                    }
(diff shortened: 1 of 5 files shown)


expenses:
Looks good code-wise. I don't get why this storage root needed variant is needed though. Does it save on a storage query call?

expenses (approved):

tomaka:
> I don't get why this storage root needed variant is needed though. Does it save on a storage query call?

We need to *somehow* provide the information when it's requested by the runtime.
There are only two options here: add a variant to the enum (what I did), or require the user to always pass the hash in the `Config`.

For the runtime call in the `metadata::query` module, it clearly doesn't make sense for the runtime to access the storage root, so if this new variant is generated I return `Error::HostFunctionNotAllowed`.
If I had gone for the solution of adding a value to `Config`, I would have had to propagate this to the `Config` in the API of `metadata::query`, which wouldn't have made sense.

expenses:
@tomaka thanks for clarifying!

PR #627, opened 2021-03-05
Implement reverting storage transactions

### LINKED ISSUE #626: Panic a bit above Polkadot block 2609152
We need to implement rolling back storage transactions in order to get past this: https://github.com/paritytech/smoldot/blob/21c26378d17dd5ac05b9d718f60218ab6b21945b/src/executor/runtime_host.rs#L676

tomaka:
Fix #626

Currently under testing.


tomaka:
Passed block 263000 and syncing well.

PR #636, opened 2021-03-08
Prototype for using the browser's Wasm VM

tomaka:
This is a work-in-progress for adding a third "Wasm provider" for executing the runtime, next to wasmi and wasmtime: the browser's virtual machine. Sorry for that pile of mess, I'm mostly using draft PRs to keep track of ongoing work.

Doing so has two main advantages:

- Removing `wasmi` from the code shrinks the generated wasm-node Wasm from ~5MiB to ~3MiB.
- It's more than likely much faster to use the browser's VM than wasmi (if anything, for the fact that wasmi itself is executed by the browser).

This change is implemented by adding an `extern {}` block containing function definitions that the environment (i.e. `index.js`) must implement. These functions include creating a new Wasm VM, running it, and so on. In order to avoid poisoning the required functions, this feature is enabled only when `--features javascript-wasm-vm` is passed.

The main difficulty in this change is that JS doesn't allow asynchronous functions as callback. Basically, it is mandatory for host functions to return instantaneously. There is however one trick that can be used: [`Atomics.wait`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics/wait). It is the one and only JS function that browsers accept and that can block the current "thread". [This repo](https://github.com/hyperdivision/async-wasm) shows that is should indeed work.

The idea is: we spawn a worker thread, start executing the Wasm synchronously, and went it calls a host function, we use `Atomics.wait` to wait until the call to the host function is handled by the non-worker thread.

However, it gets trickier. Smoldot, similarly, assumes that [executing Wasm code is synchronous](https://github.com/paritytech/smoldot/blob/c7e1729d9197a78202b74e1541a2e66fbdc82839/src/executor/vm.rs#L251) (which it normally is). The `VirtualMachine::run` function (which I just linked) is expected to run a little bit of Wasm then report about it has done.

Because of this, smoldot itself must use the same trick: run in a worker and use `Atomics.wait` when `VirtualMachine::run`.

Before continuing this PR, I'm going to try investigate what the trade-offs are when it comes to running smoldot in a worker itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,8 +59,12 @@ jobs:
     - uses: Swatinem/rust-cache@v1
     - run: cargo check --package smoldot --locked --no-default-features
     - run: cargo check --package smoldot --locked --no-default-features --features database-sled
+    - run: cargo check --package smoldot --locked --no-default-features --features javascript-wasm-vm
     - run: cargo check --package smoldot --locked --no-default-features --features std
+    - run: cargo check --package smoldot --locked --no-default-features --features database-sled --features javascript-wasm-vm
     - run: cargo check --package smoldot --locked --no-default-features --features database-sled --features std
+    - run: cargo check --package smoldot --locked --no-default-features --features javascript-wasm-vm --features std
+    - run: cargo check --package smoldot --locked --no-default-features --features database-sled --features std --features javascript-wasm-vm
 
   check-rustdoc-links:
     name: Check rustdoc intra-doc links
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ database-sled = [
     "sled",
     "std"   # A database stored on the filesystem can't reasonably work without a filesystem.
 ]
+javascript-wasm-vm = []
 std = [
     "async-std",
     "futures/thread-pool",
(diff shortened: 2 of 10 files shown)


tomaka:
Note that another possibility for implementing this could be to share some memory between smoldot and its child VMs. In other words, creating a child VM would create a `SharedArrayBuffer` and somehow inject it into smoldot.
This isn't possible yet either, as you can't inject memories at runtime yet nor access secondary memories anyway.

tomaka:
Blocked by a bug in NodeJS. Spawning a worker within a worker seems to be racy. Sometimes it works sometimes not.

tomaka:
Going to change the way it works: instead of having the smoldot Wasm VM spawn workers directly (which doesn't work on NodeJS), it's going to send back a message to its parent to ask to spawn the worker.
The smoldot Wasm VM and its worker then communicate with a `SharedArrayBuffer`.

tomaka:
Nearly done.

tomaka:
Alright, it's working. Needs some clean up before merging.

tomaka:
Unfortunately, browsers disable `SharedArrayBuffer` for security reasons unless you use some CORS headers.
I've restored `wasmi` and we're using `wasmi` as a fallback if that is the case.

tomaka:
Seems to sometimes deadlock.

tomaka:
While the deadlocks happen quite rarely, another problem is that it's actually slower than using `wasmi` due to all the necessary copying around. I clearly don't think it's worth doing this change before some Wasm features land, such as the possibility to add memory at runtime (which would allow direct access to the child VM's memory).

I'm going to leave this PR open for reference, but stop working on it.

tomaka:
Closing after https://github.com/paritytech/smoldot/issues/2488

PR #639, opened 2021-03-09
Spawn smoldot in a worker

tomaka:
Preliminary work for #636 

Spawns smoldot itself within a worker.

Because of the different incompatible flavours of JavaScript, this is waaaaaay more difficult than it really should be.

For the easy part, two main changes:

- Renamed `tcp-*` to `compat-*` to include a workers-related compatibility shim.
- Extracted some code from `index.js` to `worker.js`.

Current blocker is that we do `new Worker('./index.js')`, and this obviously can't work well with today's JavaScript.
Parcel happily accepts this, while Webpack requires passing `new URL(..., meta.import.url)`. Of course Parcel doesn't accept `meta.import.url` because that would be too easy.
We basically have to pick a bundler to use in smoldot, and bundle the worker and index together. However I can't figure how to make this work.


tomaka:
> Of course Parcel doesn't accept meta.import.url because that would be too easy.

I changed my mind after discovering that Parcel actually [should support this](https://github.com/parcel-bundler/parcel/pull/5846).

PR #646, opened 2021-03-11
Babe fetch epoch returns a complete epoch

expenses:
This is to add the changes for https://github.com/paritytech/substrate/pull/8072. I've tested it on a development substrate node and it works there, but on flaming fir I kept getting `Err(BabeFetchEpoch(WasmStart(VirtualMachine(FunctionNotFound), HostVmPrototype)))`, despite the chain being fully synced. I'm unsure whether that was caused by this change or not.


tomaka:
Oh I didn't realize that you changed the output of the runtime function.

src/sync/grandpa_warp_sync.rs
```diff
@@ -139,27 +138,15 @@ impl<TSrc> GrandpaWarpSync<TSrc> {
                             _ => unreachable!(),
                         };
 
+                    log::info!("Got next epoch: {:?}", next_epoch);
+
```
tomaka, line 142:
```suggestion
```

expenses: Oops!

tomaka:
Just so you're aware: this will break GrandPa warp sync until we do a runtime upgrade?

tomaka (approved):

expenses:
I'm aware, I suggest we only merge after a runtime upgrade.

expenses:
Polkadot has been updated so I think we can merge this right?

expenses:
Ah, `sample_decode` in `babe_fetch_epoch failed. We should either remove this test or update the data.

tomaka:
Only Westend has been updated. Polkadot and Kusama haven't.

Fortunately, since we (erroneously?) use `decode` instead of `decode_all` to decode the runtime output, the code in the main branch is still working with Westend, so there is no urgency to merge this.

expenses:
Ah, okay.

tomaka:
This is now ok to merge.

tomaka:
So the story goes like this:

- A PR was made to Westend to change the type of slot from plain to VRF, but this unfortunately only changed the config that would be used to initially populate the genesis block, so it had no effect.
- When we implemented the `Babe_epochConfig` runtime API, the VRF enum was used (since that's what was written in the code), while in reality Westend continued using plain slots.

tomaka:
André changed the config with a sudo call in block 5,567,609

PR #651, opened 2021-03-16
Fix polkadot initialisation performance issues

### LINKED ISSUE #647: Initializing Polkadot takes a long time in the Wasm node
It takes maybe around 20 to 30 seconds on my machine between `Loaded chain specs for Polkadot` and `Initialization complete`.

expenses:
Fixes #647

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2193,7 +2193,6 @@ dependencies = [
  "futures",
  "hashbrown",
  "hex",
- "itertools 0.10.0",
  "libsecp256k1",
  "merlin",
  "multihash",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,6 @@ fnv = { version = "1.0.7", default-features = false }
 futures = "0.3.13"   # TODO: no-std-ize
 hashbrown = { version = "0.9.1", default-features = false, features = ["serde"] }   # TODO: remove serde feature
 hex = { version = "0.4.3", default-features = false }
-itertools = { version = "0.10.0", default-features = false }
 libsecp256k1 = { version = "0.3.5", default-features = false }
 # Reminder: `log` is forbidden
 merlin = { version = "3.0", default-features = false }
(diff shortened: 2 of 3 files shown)


tomaka:
Thanks.

I think it would make more sense to put a `BTreeMap` directly in the chain specs modules, and add some public `genesis_storage_value` method instead of building a `HashMap` just for building the trie root then throwing it away.

expenses:
> I think it would make more sense to put a `BTreeMap` directly in the chain specs modules, and add some public `genesis_storage_value` method instead of building a `HashMap` just for building the trie root then throwing it away.

Any reason why you'd prefer a `BTreeMap` over a `HashMap`?

tomaka:
We're using `BTreeMap` across the code base for anything storage-related, because it keeps keys in order.
The runtime sometimes require you to access all the keys that start with a certain prefix, or access the key that immediately follows a provided key, and `BTreeMap` lets you do that in `O(log(n))`.

While in the chain specs we don't need these operations, I think it's not a bad idea to stay consistent.

src/chain_spec.rs
```diff
@@ -123,6 +124,7 @@ impl LightSyncState {
 #[derive(Clone)]
 pub struct ChainSpec {
     client_spec: structs::ClientSpec,
+    genesis_storage: BTreeMap<Vec<u8>, Vec<u8>>,
```
expenses, line 127: I don't like that we're not just borrowing the keys and values here, but I couldn't think of a better way of doing things that didn't make the struct self-referential.

tomaka: Can't you just modify `structs::ClientSpec`?

expenses: Not easily, as it's a serde struct

tomaka: What's the problem? Are the serde traits not implemented on `BTreeMap`?

expenses: Well the client spec is deserialized from the json right? So to create a BTreeMap inside that, you'd have to implement a custom deserialization or something.

tomaka: What I'm suggesting is to change [these two lines](https://github.com/paritytech/smoldot/blob/e980805d425f444e38972e6dd00d2028dd1b4750/src/chain_spec/structs.rs#L87-L88) to be a `BTreeMap` instead of `HashMap`

tomaka (approved):

Issue #629, opened 2021-03-05
Optimizable cache invalidation


tomaka:
This piece of code is optimizable: https://github.com/paritytech/smoldot/blob/8be5f6b22de4fb2878e0e895ca7988002d8af6b3/src/trie/calculate_root.rs#L169-L175

If a value is already `None`, we can skip the rest.

PR #652, opened 2021-03-16
Optimise calculate_root cache invalidation

expenses:
Fixes https://github.com/paritytech/smoldot/issues/629

### DIFF
--- a/src/trie/calculate_root.rs
+++ b/src/trie/calculate_root.rs
@@ -169,9 +169,15 @@ impl CalculationCache {
         // We invalidate the Merkle value of `node_to_invalidate` and all its ancestors.
         node_to_invalidate.user_data().merkle_value = None;
         let mut parent = node_to_invalidate.into_parent();
-        while let Some(mut p) = parent.take() {
-            p.user_data().merkle_value = None;
-            parent = p.into_parent();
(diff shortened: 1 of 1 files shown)


src/trie/calculate_root.rs
```diff
@@ -169,9 +169,15 @@ impl CalculationCache {
         // We invalidate the Merkle value of `node_to_invalidate` and all its ancestors.
         node_to_invalidate.user_data().merkle_value = None;
         let mut parent = node_to_invalidate.into_parent();
-        while let Some(mut p) = parent.take() {
-            p.user_data().merkle_value = None;
-            parent = p.into_parent();
+        while let Some(mut node) = parent.take() {
+            // If the node has already had it's Merkle value invalidated, then
```
tomaka, line 173:
```suggestion
            // If the node has already had its Merkle value invalidated, then
```

tomaka:
Change looks good.
I'm maybe a bit paranoid, but I think we should try syncing the first ~100k blocks of Polkadot to make sure that this doesn't break anything.

expenses:
I tested this `cargo run --release` (which syncs polkadot) on this branch and it syncs the first 100k blocks just fine.

tomaka:
Thanks

PR #655, opened 2021-03-23
Add a terminate() function to the Wasm client

tomaka:
Call this in order to destroy the smoldot client.


raoulmillais:
This needs to have an accompanying change to the typescript type definitions.   I'll DM you a patch

wirednkod (approved):

PR #657, opened 2021-03-23
`babe_fetch_epoch` returns a prototype in the `Finished(Err(_))` variant as well.

expenses:
This is a prerequisite for https://github.com/paritytech/smoldot/issues/653.

PR #659, opened 2021-03-23
Change grandpa warp sync to not return a `Finished(Err(_))` variant.

### LINKED ISSUE #653: Grandpa warp sync state machine shouldn't produce an error
The grandpa warp sync state machine is supposed to run continuously until it manages to sync.
It doesn't make sense for it to switch to the `Finished(Err(_))` state.

Errors should be reported by returning an `Error` *in addition* to the state machine, not as a replacement to the state machine.

expenses:
Closes #653 and is a follow-up to #657.


src/sync/all.rs
```diff
@@ -771,9 +771,15 @@ impl<TRq, TSrc, TBl> Idle<TRq, TSrc, TBl> {
                 let heap_pages = response.next().unwrap();
                 assert!(response.next().is_none());
 
+                let (grandpa_warp_sync, error) = sync.set_virtual_machine_params(code, heap_pages, ExecHint::Oneshot);
+
+                if let Some(error) = error {
+                    // TODO: error handling
+                }
+
                 // TODO: we use `Oneshot` because the VM is thrown away afterwards; ideally it wouldn't be be thrown away
```
tomaka, line 780:
```suggestion
```

tomaka:
I'll review after it's up to date, the diff is a bit hard to read.

src/sync/all.rs
```diff
@@ -771,11 +771,14 @@ impl<TRq, TSrc, TBl> Idle<TRq, TSrc, TBl> {
                 let heap_pages = response.next().unwrap();
                 assert!(response.next().is_none());
 
-                // TODO: we use `Oneshot` because the VM is thrown away afterwards; ideally it wouldn't be be thrown away
```
tomaka, line 774: Please restore this comment. We still throw away the VM.

src/sync/grandpa_warp_sync.rs
```diff
@@ -190,32 +194,56 @@ impl<TSrc> GrandpaWarpSync<TSrc> {
...
+                        babe_fetch_epoch::babe_fetch_epoch(babe_fetch_epoch::Config {
+                            runtime: virtual_machine,
+                            epoch_to_fetch: if fetched_current_epoch.is_some() {
+                                babe_fetch_epoch::BabeEpochToFetch::NextEpoch
+                            } else {
+                                babe_fetch_epoch::BabeEpochToFetch::CurrentEpoch
+                            },
+                        });
+
+                    return return Self::from_babe_fetch_epoch_query(
```
tomaka, line 217: It's not mentioned in the guidelines but I'd strongly prefer to not use any sort of recursion.
Everything is already wrapped in a `loop`, so you should be able to update the local variables and continue looping.

tomaka: Also, it doesn't make sense to handle errors here this way.
All the parameters to the Babe epoch fetch are verified. The outcome is deterministic. If we try again, it's going to error again and again and again.

What we should do instead is start a warp sync request again (starting from the current state, not from where we were at the beginning), in the hopes that peers will give us another block.

For example, if we warp sync to block 1 million, and the runtime at block 1 million triggers a Babe fetch error, then we should try again making a request (starting from block 1 million) in the hopes that another peer will make us jump to block 2 million, where the runtime *is* compatible.

tomaka (changes requested):

src/sync/grandpa_warp_sync.rs
```diff
@@ -656,16 +681,33 @@ impl<TSrc> VirtualMachineParamsGet<TSrc> {
...
+        GrandpaWarpSync<TSrc>,
+        Option<(Error, Option<HostVmPrototype>)>,
+    ) {
         let code = match code {
             Some(code) => code,
-            None => return GrandpaWarpSync::Finished(Err(Error::MissingCode)),
+            None => {
+                return (
+                    GrandpaWarpSync::InProgress(
+                        InProgressGrandpaWarpSync::VirtualMachineParamsGet(self),
```
tomaka, line 693: You have to switch back to `WarpSyncRequest` or `WaitingForSources` here as well.
The parameters passed to `set_virtual_machine_params` are fully verified before the function is called. There's no way the user can call the function again with different values.

src/sync/grandpa_warp_sync.rs
```diff
@@ -656,16 +681,33 @@ impl<TSrc> VirtualMachineParamsGet<TSrc> {
...
         };
 
         let heap_pages =
             match executor::storage_heap_pages_to_value(heap_pages.as_ref().map(|p| p.as_ref())) {
                 Ok(hp) => hp,
-                Err(err) => return GrandpaWarpSync::Finished(Err(Error::InvalidHeapPages(err))),
+                Err(err) => {
+                    return (
+                        GrandpaWarpSync::InProgress(
+                            InProgressGrandpaWarpSync::VirtualMachineParamsGet(self),
```
tomaka, line 706: Same remark

src/sync/grandpa_warp_sync.rs
```diff
@@ -676,13 +718,23 @@ impl<TSrc> VirtualMachineParamsGet<TSrc> {
...
+                );
+
+                (
+                    grandpa_warp_sync,
+                    error.map(|(error, runtime)| (error, Some(runtime))),
                 )
             }
-            Err(error) => GrandpaWarpSync::Finished(Err(Error::NewRuntime(error))),
+            Err(error) => (
+                GrandpaWarpSync::InProgress(InProgressGrandpaWarpSync::VirtualMachineParamsGet(
```
tomaka, line 733: Ditto

src/sync/grandpa_warp_sync.rs
```diff
@@ -338,7 +360,7 @@ impl<TSrc> StorageGet<TSrc> {
     pub fn inject_value(
         self,
         value: Option<impl Iterator<Item = impl AsRef<[u8]>>>,
-    ) -> GrandpaWarpSync<TSrc> {
+    ) -> (GrandpaWarpSync<TSrc>, Option<(Error, HostVmPrototype)>) {
```
tomaka, line 363:
```suggestion
    ) -> (GrandpaWarpSync<TSrc>, Option<Error>) {
```

You don't need to pass back the virtual machine, since the error is that the runtime in the virtual machine has a problem.

src/sync/grandpa_warp_sync.rs
```diff
@@ -389,7 +411,10 @@ impl<TSrc> NextKey<TSrc> {
     ///
     /// Panics if the key passed as parameter isn't strictly superior to the requested key.
     ///
-    pub fn inject_key(self, key: Option<impl AsRef<[u8]>>) -> GrandpaWarpSync<TSrc> {
+    pub fn inject_key(
+        self,
+        key: Option<impl AsRef<[u8]>>,
+    ) -> (GrandpaWarpSync<TSrc>, Option<(Error, HostVmPrototype)>) {
```
tomaka, line 417:
```suggestion
    ) -> (GrandpaWarpSync<TSrc>, Option<Error>) {
```

src/sync/grandpa_warp_sync.rs
```diff
@@ -656,16 +681,33 @@ impl<TSrc> VirtualMachineParamsGet<TSrc> {
         code: Option<impl AsRef<[u8]>>,
         heap_pages: Option<impl AsRef<[u8]>>,
         exec_hint: ExecHint,
-    ) -> GrandpaWarpSync<TSrc> {
+    ) -> (
+        GrandpaWarpSync<TSrc>,
+        Option<(Error, Option<HostVmPrototype>)>,
```
tomaka, line 686:
```suggestion
        Option<Error>,
```

tomaka (changes requested):

src/sync/grandpa_warp_sync.rs
expenses: Oh right, so the comment `// TODO: we use 'Oneshot' because the VM is thrown away afterwards; ideally it wouldn't be be thrown away` is referring to what happens to the VM upon success?

tomaka: Yes!

src/sync/grandpa_warp_sync.rs
```diff
@@ -198,24 +201,43 @@ impl<TSrc> GrandpaWarpSync<TSrc> {
...
                     query = storage_root.resume(&state.header.state_root);
                 }
                 (babe_fetch_epoch::Query::NextKey(next_key), fetched_current_epoch) => {
-                    return Self::InProgress(InProgressGrandpaWarpSync::NextKey(NextKey {
-                        inner: next_key,
-                        fetched_current_epoch,
-                        state,
-                    }))
+                    return (
+                        Self::InProgress(InProgressGrandpaWarpSync::NextKey(NextKey {
```
tomaka, line 234: `InProgressGrandpaWarpSync::warp_sync_request_from_next_source`

src/sync/grandpa_warp_sync.rs
```diff
@@ -198,24 +201,43 @@ impl<TSrc> GrandpaWarpSync<TSrc> {
...
+                    )
+                }
                 (babe_fetch_epoch::Query::StorageGet(storage_get), fetched_current_epoch) => {
-                    return Self::InProgress(InProgressGrandpaWarpSync::StorageGet(StorageGet {
-                        inner: storage_get,
-                        fetched_current_epoch,
-                        state,
-                    }))
+                    return (
+                        Self::InProgress(InProgressGrandpaWarpSync::StorageGet(StorageGet {
```
tomaka, line 220: `InProgressGrandpaWarpSync::warp_sync_request_from_next_source`

tomaka (changes requested):

src/sync/grandpa_warp_sync.rs
expenses: I'm confused, in what circumstances would you call that here instead of returning `StorageGet`?

tomaka: Nevermind, I assumed that it was an error because it shows up in the diff

tomaka (approved):

PR #670, opened 2021-03-25
Change relay_chain_spec field to parachain_spec

tomaka:
cc #221 

Before this PR, you're expected to pass a parachain spec (simply called `chain_spec`), relay chain spec, and database.
There's a problem with that: the database shouldn't apply to the parachain but to the relay chain.

This PR renames the `relay_chain_spec` field to `parachain_spec`.
If you passed for example `{ chain_spec: tick, relay_chain_spec: rococo }` you now have to pass `{ chain_spec: rococo, parachain: tick }`.
This makes the database apply to the relay chain, as intended.


tomaka:
The PR is rather short, because the actual glue between the relay chain and parachain isn't implemented yet (https://github.com/paritytech/smoldot/pull/549).

PR #673, opened 2021-03-25
Add a log_callback field in the config

tomaka:
Allows passing a `log_callback` field in the configuration of `start`.

It gets called whenever a log message is emitted by smoldot, such as `[smoldot_js] Initialization complete`.
In this example, the `level` would be 3 (for "info"), the target would be `smoldot_js`, and the message would be `Initialization complete`.


bin/wasm-node/javascript/src/index.js
```diff
@@ -27,6 +27,20 @@ export async function start(config) {
...
+      console.error("[" + target + "]", message);
+    } else if (level == 2) {
+      console.warn("[" + target + "]", message);
+    } else if (level == 3) {
+      console.info("[" + target + "]", message);
+    } else if (level == 4) {
+      console.debug("[" + target + "]", message);
+    } else {
+      console.trace("[" + target + "]", message);
+    }
```
tomaka, line 41: This is the default implementation if no callback is provided, to get an idea of how it's called.

PR #676, opened 2021-03-25
Set 1.51.0 as the minimum Rust version

tomaka:
~~I didn't set `resolver = "2"` just yet, because it causes build errors.~~ EDIT: done


tomaka:
The `rust` image on Docker hub isn't up to date yet. Will try again tomorrow.

tomaka:
Actually, let's merge without the new resolver, and I'll open another one with the new resolver.

PR #682, opened 2021-03-26
Rework lib.rs to integrate parachains

### LINKED ISSUE #221: Make it sync parachains

tomaka:
I reorganized `start_services` in order to make parachains work.

The code has three known flaws:

- We match relay chains and parachains by the chain identifier. This means that for example the `relay_chain` field in the Tick chain spec must contain a string, and that the `id` field in the Rococo spec must contain the exact same string. This isn't the case upstream right now: Rococo's id is `rococo_v1`, while Tick contains just `rococo`.

- The database saving task is spawned for the first relay chain in the list of chains. I'm leaning towards removing database altogether in the future (in favour of GrandPa warp sync), which is why I'm not really dealing with this problem. If we're not removing databases, we instead have to modify the FFI between the JS and Rust.

- The JSON-RPC task is spawned for the first chain in the list. In the future, I think we should add a way for the JS to specify which chains have a JSON-RPC task, and for the JS to indicate which chain each individual request applies to.

Close #549 
Close #221

### DIFF
--- a/bin/wasm-node/rust/src/lib.rs
+++ b/bin/wasm-node/rust/src/lib.rs
@@ -212,7 +212,7 @@ async fn start_services(
     chain_specs: Vec<chain_spec::ChainSpec>,
 ) {
     // The network service is responsible for connecting to the peer-to-peer network
-    // of the chain.
+    // of all chains.
     let (network_service, mut network_event_receivers) =
         network_service::NetworkService::new(network_service::Config {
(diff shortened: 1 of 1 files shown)


tomaka:
I haven't tested the code very thoroughly, but will do after merging.

Issue #678, opened 2021-03-25
Full node initialization very slow with existing database


tomaka:
The full node initialization is currently extremely slow because it copies the entire storage in memory: https://github.com/paritytech/smoldot/blob/5b95d60892d1e2349af4b4e73e7bb7be60d4e199/bin/full-node/src/sync_service.rs#L140-L151

At the block where I am, the storage is only ~83 MiB but loading it takes a minute or two.

expenses:
I wanted to test this out so I can the full node using perf and opening the database took a minute or two, but because of code in sled, not the code you link.
![1616762812](https://user-images.githubusercontent.com/13566135/112634037-647b9480-8e3a-11eb-893d-cd070037ec1a.png)

I then quit the node and ran it again, but this time it loaded in much less time. Does sled do any kind of decompression to disk?

tomaka:
If the node shut down in an unclean way, reopening a database can take a lot of time because it has to go through the journal. This isn't really specific to sled and is a consequence of how databases are designed.

The issue I mention happens for me when I'm around Polkadot block 2 million. If I start syncing from scratch, I don't immediately have the slowness.

tomaka:
Fixed by https://github.com/paritytech/smoldot/pull/689

PR #689, opened 2021-03-27
Replace sled with SQLite

tomaka:
This PR replaces the `sled` database with SQLite.
SQLite has several advantages:

- It is extremely mature. No risk of database corruption, breaking changes, or weird slowness.
- There exists external tools that allow looking inside of an SQLite database. You can realistically stop smoldot and look at the database content manually, whereas with sled you'd have to write code for that.
- Having a strong relational database schema reduces complexity overall.
- SQLite seems to use around 0.8-0.9 kiB per block, compared to around 3.5kiB/block with sled. This is surprising, considering that the volume of data is strictly the same in both cases (there isn't anything stored after this PR that wasn't already stored before) and that sled has compression enabled.
- Initialization is considerably faster with SQLite. https://github.com/paritytech/smoldot/issues/678 is fixed.

One drawback of SQLite is that it requires a C compiler. However, `sled`, when compression is enabled, requires one as well. It is not realistic to run `sled` with compression disabled, as the database just explodes in size.

Another drawback is that the API of the Rust `sqlite` library is a bit tedious to use.

In terms of speed, syncing smoldot is capped by the Wasm verification speed, not by the database writing speed. Both sled and SQLite are the same here.


tomaka:
> Another drawback is that the API of the Rust sqlite library is a bit tedious to use.

To expand: in particular, I wish it was possible to prepare all statements in advance, but because of the lifetimes system I have to recreate them every single time.
Having to bind parameters to statements is also very error-prone.

None of these two problems is critical.

tomaka:
> SQLite seems to use around 0.8-0.9 kiB per block, compared to around 3.5kiB/block with sled. This is surprising, considering that the volume of data is strictly the same in both cases

After verification, we were never removing entries from `non_finalized_changes_keys_tree`, which is probably what causes this.

tomaka:
I guess this would normally need a very careful review, but :shrug:

PR #691, opened 2021-03-29
Return the index of the chain that responds to json rpc calls

### LINKED ISSUE #685: Improve the JSON-RPC system in the wasm-node FFI
The wasm-node's API supports connecting to multiple chains at the same time, but the JSON-RPC task is spawned for the first chain in the list.
This is at the moment the only solution, because JSON-RPC requests and responses don't mention anything about which chain is concerned. As such, we can't spawn multiple JSON-RPC tasks.

We should modify the FFI between the wasm-node Rust and JS to:

- Indicate which chains should have a JSON-RPC service running.
- Indicate, when a request is sent or response sent back, which chain is concerned.

expenses:
I believe this would close #685, but it's possibly completely the wrong way to do things.


tomaka:
The idea is to allow spawning multiple JSON-RPC services, not to report which one we spawn.

expenses:
Right, okay. Well this is a start on that at least.

expenses:
So I guess I misunderstood the issue. I guess we want an ffi import like `pub fn json_rpc_running_services(ptr, len)` which returns a boolean array perhaps? And then `json_rpc_send` takes in a chain id.

tomaka:
> And then json_rpc_send takes in a chain id.

Yes

> we want an ffi import like pub fn json_rpc_running_services(ptr, len) which returns a boolean array perhaps?

No. The user decides which services are running. You can just modify the `ChainConfig` struct to add `json_rpc_running: bool`. You don't have to touch `ffi::init`, as it needs deeper changes.

bin/wasm-node/rust/src/ffi.rs
```diff
@@ -499,40 +499,52 @@ fn init(
...
+    pub(crate) chain_index: usize,
+}
+
 lazy_static::lazy_static! {
-    static ref JSON_RPC_CHANNEL: (mpsc::UnboundedSender<Box<[u8]>>, futures::lock::Mutex<mpsc::UnboundedReceiver<Box<[u8]>>>) = {
+    static ref JSON_RPC_CHANNEL: (mpsc::UnboundedSender<Request>, futures::lock::Mutex<mpsc::UnboundedReceiver<Request>>) = {
         let (tx, rx) = mpsc::unbounded();
         (tx, futures::lock::Mutex::new(rx))
     };
 }
```
expenses, line 527: Okay, so do we want multiple channels (1 per chain / 1 per json rpc service?) or a single channel that passes in a `chain_index`? If so, how should I rewrite `next_json_rpc` so that it waits on a request for a specific chain index and doesn't discard any requests that are for different chain index?

bin/wasm-node/rust/src/ffi.rs
```diff
@@ -499,40 +499,52 @@ fn init(
...
             database_content,
+            json_rpc_running: true,
         })
         .chain(
             parachain_specs
                 .into_iter()
                 .map(|specification| super::ChainConfig {
                     specification,
                     database_content: None,
+                    json_rpc_running: true,
```
tomaka, line 510:
```suggestion
                    json_rpc_running: false,
```

bin/wasm-node/rust/src/lib.rs
```diff
@@ -413,45 +420,52 @@ async fn start_services(
...
-                genesis_block_state_root: first_chain_finalized_header.state_root,
+                sync_service: sync_service.clone(),
             })
-            .boxed(),
-        )
-        .unwrap();
+            .await,
+        );
+
+        if json_rpc_running {
```
tomaka, line 447: The `transactions_service` is more or less part of the JSON-RPC service and doesn't need to be started if `json_rpc_running` is false.
I suggest to just do `if !json_rpc_running { continue; }`

bin/wasm-node/rust/src/ffi.rs
tomaka: I'd suggest to keep one channel, but instead of having the JSON-RPC service poll the channel directly, you spawn an additional new task common to all JSON-RPC services that is dedicated to polling it.
In other words, move [this piece of code](https://github.com/paritytech/smoldot/blob/ee4895be2a58399953aaca085c9a404361ae5a24/bin/wasm-node/rust/src/json_rpc_service.rs#L162-L201) outside of the JSON-RPC service, and make `handle_rpc` public.

tomaka (changes requested):

bin/wasm-node/rust/src/ffi.rs
expenses: Okay, how would this task call `handle_rpc` on the clients? Should it just be passed in a `Vec`/`HashMap` of clients?

tomaka: Yes

bin/wasm-node/rust/src/ffi/bindings.rs
```diff
@@ -243,8 +243,8 @@ pub extern "C" fn init(
 ///
 /// Responses and subscriptions notifications are sent back using [`json_rpc_respond`].
 #[no_mangle]
-pub extern "C" fn json_rpc_send(text_ptr: u32, text_len: u32) {
-    super::json_rpc_send(text_ptr, text_len)
+pub extern "C" fn json_rpc_send(text_ptr: u32, text_len: u32, chain_index: u32) {
+    super::json_rpc_send(text_ptr, text_len, chain_index)
 }
```
tomaka, line 249: You also need to modify the `json_rpc_respond` function to pass back the chain index as well.

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -44,6 +44,65 @@ use std::{
...
+                };
+
+                match clients.get(&chain_index).cloned() {
+                    Some(client) => client.handle_rpc(request_id, call).await,
+                    None => {
+                        log::warn!(
+                            target: "json-rpc",
+                            "Ignoring JSON-RPC call for chain index {}, for which a JSON-RPC service has not been started.",
+                            chain_index
+                        )
```
tomaka, line 98: This should send back a JSON-RPC error response actually. You can find how to do that elsewhere in the file.

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -44,6 +44,65 @@ use std::{
     sync::{atomic, Arc},
 };
 
+/// Spawns a task to handle incoming JSON-RPC requests.
```
tomaka, line 47:
```suggestion
/// Spawns a task to handle incoming JSON-RPC requests.
///
/// The task queries incoming requests and dispatches them to the JSON-RPC
/// services passed as parameter.
```

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -44,6 +44,65 @@ use std::{
...
+pub async fn request_handling_task(
+    mut tasks_executor: Arc<Mutex<Box<dyn FnMut(Pin<Box<dyn Future<Output = ()> + Send>>) + Send>>>,
+    clients: Arc<HashMap<usize, Arc<JsonRpcService>>>,
```
tomaka, line 50:
```suggestion
    json_rpc_services: HashMap<usize, Arc<JsonRpcService>>,
```

Please create the `Arc` inside of the function's body, as it's required because of implementation details.

tomaka (changes requested):

bin/wasm-node/javascript/README.md
```diff
@@ -19,10 +19,10 @@ const chain_spec = Buffer.from(fs.readFileSync('./westend.json')).toString('utf8
...
   .start({
     chain_spec: chain_spec,
-    json_rpc_callback: (resp) => {
+    json_rpc_callback: (resp, chain_index) => {
         // Called whenever the client emits a response to a JSON-RPC request,
         // or a JSON-RPC pub-sub notification.
-        console.log(resp)
+        console.log(resp, chain_index)
     }
   })
```
tomaka, line 27: Well, you need to update the README to pass the `chain_index` when sending the request as well

bin/wasm-node/javascript/README.md
```diff
@@ -19,10 +19,10 @@ const chain_spec = Buffer.from(fs.readFileSync('./westend.json')).toString('utf8
 smoldot
   .start({
     chain_spec: chain_spec,
-    json_rpc_callback: (resp) => {
+    json_rpc_callback: (resp, chain_index) => {
         // Called whenever the client emits a response to a JSON-RPC request,
         // or a JSON-RPC pub-sub notification.
-        console.log(resp)
+        console.log(resp, chain_index)
```
tomaka, line 25:
```suggestion
        console.log(resp)
```

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -88,14 +93,23 @@ pub async fn request_handling_task(
...
+                match json_rpc_services.get(&chain_index).cloned() {
+                    Some(service) => service.handle_rpc(request_id, call).await,
                     None => {
-                        log::warn!(
-                            target: "json-rpc",
-                            "Ignoring JSON-RPC call for chain index {}, for which a JSON-RPC service has not been started.",
-                            chain_index
-                        )
+                        // Just pick any service to send back an error response.
+                        let any_service = json_rpc_services.values().next().unwrap().clone();
```
tomaka, line 100: Well no. You have to send back using the `chain_index` that the user wanted.

tomaka (changes requested):

bin/wasm-node/javascript/README.md
```diff
@@ -15,18 +15,19 @@ import * as smoldot from 'smoldot';
 
 // Load a string chain specifications.
 const chain_spec = Buffer.from(fs.readFileSync('./westend.json')).toString('utf8');
+const chain_index = 0;
```
tomaka, line 18:
```suggestion
```

bin/wasm-node/javascript/README.md
```diff
@@ -15,18 +15,19 @@ import * as smoldot from 'smoldot';
...
-    json_rpc_callback: (resp) => {
+    json_rpc_callback: (resp, chain_index) => {
         // Called whenever the client emits a response to a JSON-RPC request,
         // or a JSON-RPC pub-sub notification.
         console.log(resp)
     }
   })
   .then((client) => {
-    client.send_json_rpc('{"jsonrpc":"2.0","id":1,"method":"system_name","params":[]}');
+    client.send_json_rpc('{"jsonrpc":"2.0","id":1,"method":"system_name","params":[]}', chain_index);
```
tomaka, line 30:
```suggestion
    client.send_json_rpc('{"jsonrpc":"2.0","id":1,"method":"system_name","params":[]}', 0);
```

bin/wasm-node/javascript/src/worker.js
```diff
@@ -32,6 +32,8 @@ import { default as wasm_base64 } from './autogen/wasm.js';
 //
 let state = null;
 
+const chain_index = 0;
+
```
tomaka, line 36:
```suggestion
```

bin/wasm-node/javascript/src/worker.js
```diff
@@ -102,7 +104,7 @@ const startInstance = async (config) => {
     const len = Buffer.byteLength(json_rpc_request, 'utf8');
     const ptr = result.instance.exports.alloc(len);
     Buffer.from(result.instance.exports.memory.buffer).write(json_rpc_request, ptr);
-    result.instance.exports.json_rpc_send(ptr, len);
+    result.instance.exports.json_rpc_send(ptr, len, chain_index);
```
tomaka, line 107: The `chain_index` needs to come from the message, not hardcoded to 0.

bin/wasm-node/javascript/src/worker.js
```diff
@@ -124,6 +126,6 @@ compat.setOnMessage((message) => {
     const len = Buffer.byteLength(message, 'utf8');
     const ptr = state.exports.alloc(len);
     Buffer.from(state.exports.memory.buffer).write(message, ptr);
-    state.exports.json_rpc_send(ptr, len);
+    state.exports.json_rpc_send(ptr, len, chain_index);
```
tomaka, line 129: Same

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -80,10 +153,13 @@ pub struct Config {
     /// >           this value, doing so is quite expensive. We prefer to require this value
     /// >           from the upper layer instead.
     pub genesis_block_state_root: [u8; 32],
+
+    /// The index of the chain that this service is handling requests for.
```
tomaka, line 157:
```suggestion
    /// The index of the chain that this service is handling requests for. Used only for the FFI layer.
```

tomaka (changes requested):

expenses:
I don't know how to fix the javascript errors.

bin/wasm-node/javascript/src/compat-browser.js
```diff
@@ -21,5 +21,5 @@ export const net = null;
 export const Worker = typeof window != 'undefined' ? window.Worker : null;
 export const workerOnMessage = (worker, callback) => { worker.onmessage = (event) => callback(event.data) };
 export const workerOnError = (worker, callback) => { worker.onerror = callback; };  // TODO: unclear if the parameter of the callback is same as with NodeJS
-export const postMessage = (msg) => self.postMessage(msg);
+export const postMessage = (msg, chain_index) => self.postMessage(msg, chain_index);
```
tomaka, line 24:
```suggestion
export const postMessage = (msg) => self.postMessage(msg);
```

This code is here to mimic https://developer.mozilla.org/fr/docs/Web/API/Window/postMessage

bin/wasm-node/javascript/src/index.js
```diff
@@ -114,15 +116,18 @@ export async function start(config) {
   // JSON-RPC request and wait for the response. While this might seem like an unnecessary
   // overhead, it is the most straight-forward solution. Any alternative with a lower overhead
   // would have a higher complexity.
-  worker.postMessage('{"jsonrpc":"2.0","id":1,"method":"system_name","params":[]}')
+  worker.postMessage(
+    '{"jsonrpc":"2.0","id":1,"method":"system_name","params":[]}',
+    initialization_chain_index
+  );
```
tomaka, line 127:
```suggestion
  worker.postMessage({
    request: '{"jsonrpc":"2.0","id":1,"method":"system_name","params":[]}',
    chain_index: initialization_chain_index
  });
```

tomaka (changes requested):

tomaka (approved):

Issue #654, opened 2021-03-23
JSON-RPC requests subscriptions clean up helpers


tomaka:
In the context of substrate-connect, multiple web pages "connect" to the substrate-connect extension and send JSON-RPC queries. Substrate-connect in turns sends these JSON-RPC queries to smoldot.

For regular queries, this is rather easier. However, when it comes to subscriptions, substrate-connect has to know which methods are subscription/un-subscriptions, and keep track of which web page has which active subscriptions.
If a web page is closed, substrate-connect needs to send un-subscription requests to smoldot.

In order to avoid having to maintain a list in substrate-connect of JSON-RPC methods that are subscriptions and un-subscrptions, I think that smoldot should handle this natively.

- Instead of passing only the stringified JSON-RPC request, substrate-connect would additionally pass to smoldot an integer representing the web page (or anything similar) that is performing this request.
- The responses sent by smoldot, would also contain this same integer.
- A new function should be added to ask smoldot to clean all subscriptions and cancel all requests corresponding to a certain integer, which substrate-connect would call when a web page closes.

expenses:
How would this 'integer representing the web page' differ from [`request_id`](https://github.com/paritytech/smoldot/blob/91683a38a21146a1f8dfe51dc0d8a2e84eb8cc7c/bin/wasm-node/rust/src/json_rpc_service.rs#L285)?

tomaka:
The `request_id` is chosen by the webpage itself. If there are multiple webpages connect to a single `smoldot`, we must be sure that they don't conflict.

As far as I know, right now `subtrate-connect` currently re-maps all the `request_id`s for this reason, but we should do this in smoldot instead, as explained in the OP.

PR #692, opened 2021-03-29
Store most JSON RPC subscriptions via a source id and add a `json_rpc_unsubscribe_all` ffi call.

expenses:
This closes https://github.com/paritytech/smoldot/issues/654.


expenses:
If we're going to be having multiple JSON-RPC services, then we probably want to store the chain id for the subscriptions as well, as just using `source_id` would result in clashes. We probably want a multi-layered hashmap where we sort by `source_id` first so that `unsubscribe_all` is efficient.

bin/wasm-node/rust/src/ffi.rs
```diff
@@ -512,29 +512,53 @@ fn init(
...
+    pub(crate) source_id: u32,
+}
+
 lazy_static::lazy_static! {
-    static ref JSON_RPC_CHANNEL: (mpsc::UnboundedSender<Box<[u8]>>, futures::lock::Mutex<mpsc::UnboundedReceiver<Box<[u8]>>>) = {
+    static ref JSON_RPC_CHANNEL: (mpsc::UnboundedSender<Request>, futures::lock::Mutex<mpsc::UnboundedReceiver<Request>>) = {
+        let (tx, rx) = mpsc::unbounded();
+        (tx, futures::lock::Mutex::new(rx))
+    };
+    static ref UNSUBSCRIBE_ALL_CHANNEL: (mpsc::UnboundedSender<u32>, futures::lock::Mutex<mpsc::UnboundedReceiver<u32>>) = {
```
tomaka, line 526: Please use only one channel and turn `Request` into an enum. We don't want to lose the ordering of the things sent by the outside.

bin/wasm-node/rust/src/ffi.rs
```diff
@@ -512,29 +512,53 @@ fn init(
     ));
 }
 
+#[derive(Debug)]
+pub(crate) struct Request {
+    pub(crate) method_name: Box<[u8]>,
```
tomaka, line 517:
```suggestion
    pub(crate) json_rpc_request: Box<[u8]>,
```

It's not just a method name

bin/wasm-node/javascript/src/worker.js
```diff
@@ -98,11 +98,13 @@ const startInstance = async (config) => {
...
   );
 
+  const source_index = 0;
+
   state.forEach((json_rpc_request) => {
     const len = Buffer.byteLength(json_rpc_request, 'utf8');
     const ptr = result.instance.exports.alloc(len);
     Buffer.from(result.instance.exports.memory.buffer).write(json_rpc_request, ptr);
-    result.instance.exports.json_rpc_send(ptr, len);
+    result.instance.exports.json_rpc_send(ptr, len, source_index);
```
tomaka, line 107: There's another call to `json_rpc_send` below

bin/wasm-node/javascript/src/worker.js
```diff
@@ -98,11 +98,13 @@ const startInstance = async (config) => {
     config.maxLogLevel
   );
 
+  const source_index = 0;
```
tomaka, line 101: `0`? This seems unfinished. It needs to be exposed in the API of `index.js`

bin/wasm-node/rust/src/ffi/bindings.rs
```diff
@@ -241,10 +241,19 @@ pub extern "C" fn init(
...
 /// Responses and subscriptions notifications are sent back using [`json_rpc_respond`].
 #[no_mangle]
-pub extern "C" fn json_rpc_send(text_ptr: u32, text_len: u32) {
-    super::json_rpc_send(text_ptr, text_len)
+pub extern "C" fn json_rpc_send(text_ptr: u32, text_len: u32, request_source_id: u32) {
+    super::json_rpc_send(text_ptr, text_len, request_source_id)
+}
+
+/// Unsubscribe all the JSON-RPC subscriptions for a source. Should be called when closing a window
+/// that's connected to smoldot.
```
tomaka, line 253:
```suggestion
/// Unsubscribe all the JSON-RPC subscriptions for a source.
```

The concept of "window" isn't relevant here.

bin/wasm-node/rust/src/ffi/bindings.rs
```diff
@@ -241,10 +241,19 @@ pub extern "C" fn init(
 /// The buffer passed as parameter **must** have been allocated with [`alloc`]. It is freed when
 /// this function is called.
 ///
+/// Additionally, an index that represents the source of the request is also passed as a parameter.
```
tomaka, line 244: Please mention that this `source_id` is completely opaque and can be chosen freely

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -231,22 +264,22 @@ struct JsonRpcService {
     /// For each active finalized blocks subscription (the key), a sender. If the user
     /// unsubscribes, send the unsubscription request ID of the channel in order to close the
     /// subscription.
-    all_heads: Mutex<HashMap<String, oneshot::Sender<String>>>,
+    all_heads: Mutex<HashMap<u32, oneshot::Sender<String>>>,
```
tomaka, line 267:
```suggestion
    all_heads: Mutex<HashMap<SubscriptionId, oneshot::Sender<String>>>,
```

I don't understand why you replaced some `String` with `u32` and some others with `SubscriptionId`. They should all be `SubscriptionId`.

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -231,22 +264,22 @@ struct JsonRpcService {
...
 
     /// Same principle as [`JsonRpcService::all_heads`], but for new heads subscriptions.
-    new_heads: Mutex<HashMap<String, oneshot::Sender<String>>>,
+    new_heads: Mutex<HashMap<u32, oneshot::Sender<String>>>,
```
tomaka, line 270:
```suggestion
    new_heads: Mutex<HashMap<SubscriptionId, oneshot::Sender<String>>>,
```

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -231,22 +264,22 @@ struct JsonRpcService {
...
 
     /// Same principle as [`JsonRpcService::all_heads`], but for finalized heads subscriptions.
-    finalized_heads: Mutex<HashMap<String, oneshot::Sender<String>>>,
+    finalized_heads: Mutex<HashMap<u32, oneshot::Sender<String>>>,
```
tomaka, line 273:
```suggestion
    finalized_heads: Mutex<HashMap<SubscriptionId, oneshot::Sender<String>>>,
```

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -282,7 +315,12 @@ impl JsonRpcService {
     ///
     /// Depending on the request, either calls [`JsonRpcService::send_back`] immediately or
     /// spawns a background task for further processing.
-    async fn handle_rpc(self: Arc<JsonRpcService>, request_id: &str, call: MethodCall) {
+    async fn handle_rpc(
+        self: Arc<JsonRpcService>,
+        request_id: &str,
+        call: MethodCall,
+        source_id: u32,
```
tomaka, line 322:
```suggestion
        source_id: u32,
        request_id: &str,
        call: MethodCall,
```

This is a bit nit-picky, but putting arguments in order of importance isn't a bad idea.

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -791,11 +839,33 @@ impl JsonRpcService {
...
+             _| s_id != source_id,
+        );
+    }
+
     /// Handles a call to [`methods::MethodCall::author_submitAndWatchExtrinsic`].
     async fn submit_and_watch_extrinsic(
         self: Arc<JsonRpcService>,
         request_id: &str,
         transaction: methods::HexString,
+        source_id: u32,
```
tomaka, line 868:
```suggestion
        source_id: u32,
        request_id: &str,
        transaction: methods::HexString,
```

(same below)

tomaka (changes requested):

expenses:
Now based on #691. Let's get that merged first.

bin/wasm-node/rust/src/ffi/bindings.rs
```diff
@@ -226,10 +226,25 @@ pub extern "C" fn init(
...
+/// This source ID is completel opaque and can be chosen freely by the caller.
```
tomaka, line 230:
```suggestion
/// This source ID is completely opaque and can be chosen freely by the caller.
```

tomaka (changes requested):
The callback (`send_back` and `json_rpc_callback` and all) also needs to be passed the `source_id`

tomaka:
There's also some complication: if you just drop all the `cancel_tx` channels, you'll stop the subscription, but there's a possible race condition in which a notification is sent back after the force-unsubscription.

This is why for example `chain_unsubscribeNewHeads` is sent back in the task that handles the subscription, and not directly in the task that handles the unsubscription request.

I haven't looked into how to fix that.

tomaka:
> if you just drop all the cancel_tx channels, you'll stop the subscription, but there's a possible race condition in which a notification is sent back after the force-unsubscription.

I'd be grateful if you find a way to address that.

expenses:
I don't understand the problem. What do you mean by force-unsubscription? Why is sending back a notification a problem?

tomaka:
If the external API calls `json_rpc_unsubscribe_all(id)`, then I would expect that no notification corresponding to `id` are sent back anymore. If that's not the case, then the external API doesn't know how much time to wait before it can re-use an `id` that has been used before.

expenses:
Right, okay. I just don't understand the deadlock then. How can `oneshot::Sender` trigger a notification?

tomaka:
It's not a deadlock but a race condition.

Each subscription spawns its own dedicated task in the background whose job is to send back notifications when needed.

Sending an item on the `Sender` instructs this background task to send an unsubscription confirmation then stop.
If you simply drop the channel then return, the task is still running and might send back notifications.

The most sane way to fix this is that the tasks somehow make sure that their corresponding subscriptions are still alive before sending a notification.

om8n777:
Gg34

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -57,72 +57,84 @@ pub async fn request_handling_task(
...
+                        }
+                    }));
+                }
+                ffi::JsonRpcMessage::UnsubscribeAll { source_id } => {
+                    let json_rpc_services = json_rpc_services.clone();
+                    (tasks_executor.lock().await)(Box::pin(async move {
+                        for service in json_rpc_services.values().cloned() {
+                            service.handle_unsubscribe_all(source_id).await;
+                        }
+                    }));
```
tomaka, line 135:
```suggestion
                        for service in json_rpc_services.values().cloned() {
                            service.handle_unsubscribe_all(source_id).await;
                        }
```

You're trying to solve race conditions. We shouldn't spawn move asynchronous tasks!

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -1004,48 +1143,60 @@ impl JsonRpcService {
...
+                            );
+                        }
+                        future::Either::Right((Ok(unsub_request_id), _)) => {
+                            let response = methods::Response::chain_unsubscribeAllHeads(true)
+                                .to_json_response(&unsub_request_id);
+                            client.send_back(&response, source_id);
+                            break;
+                        }
+                        future::Either::Right((Err(_), _)) => break,
+                    },
```
tomaka, line 1180: While this technically solves the race condition, it only does so in a single-threaded environment and because we don't have use any other `await` after getting notified of a new block. I'd like to veto this solution.

expenses: 😮‍💨 what do you recommend instead?

bin/wasm-node/rust/src/json_rpc_service.rs
```diff
@@ -1232,30 +1387,38 @@ impl JsonRpcService {
...
                 match future::select(next_block, &mut unsubscribe_rx).await {
                     future::Either::Left((changes, _)) => {
-                        client.send_back(&smoldot::json_rpc::parse::build_subscription_event(
-                            "state_storage",
-                            &subscription,
-                            &serde_json::to_string(&changes).unwrap(),
-                        ));
+                        if source_disconnected.load(atomic::Ordering::Relaxed) {
+                            break;
+                        }
```
tomaka, line 1407: What if it's immediately set to `false` afterwards and before you reach the line below?

tomaka:
The solution to me is to put a `Arc<JsonRpcService>` in each subscription, lock the `Mutex` before sending back a notification, check if the subscription is still in the `HashMap`, and unlock only after the notification has been sent back.

bin/wasm-node/rust/src/ffi/bindings.rs
```diff
@@ -73,8 +73,8 @@ extern "C" {
     ///
     /// The response or notification is a UTF-8 string found in the memory of the WebAssembly
     /// virtual machine at offset `ptr` and with length `len`. `chain_index` is the chain
-    /// that the request was made to.
-    pub fn json_rpc_respond(ptr: u32, len: u32, chain_index: u32);
+    /// that the request was made to. `source_id` is the source that made the request.
```
tomaka, line 76:
```suggestion
    /// that the request was made to. `source_id` is the value that was passed to [`json_rpc_send`].
```

tomaka:
It's actually still racy. Someone might unsubscribe_from_all, then immediately after create a new subscription with the same `source_id`, and all the old subscriptions will be kept alive.

I can finish the changes if you want.

expenses:
> I can finish the changes if you want.

Yes please 😅

tomaka (approved):
Well, looks good to me :eyes:

tomaka:
What I ended up doing is: `PerSourcesSubscriptions` now stores `Arc`s.
When subscribing, we clone the `Arc`. When a notification must be sent back, we compare the expected `Arc` with the `Arc` in the hashmap. If they're the same, the notification is sent. If they're different, the task stops.

This means that as soon as "unregister_all" is called, all these `Arc` comparisons will be guaranteed to compare un-equal in the future, even if the same "source_id" is re-allocated immediately after.

tomaka:
@expenses Feel free to review, or if you don't want to review put the `automerge`

PR #697, opened 2021-03-31
Improve typescript typings test coverage

raoulmillais:
- Removes unneeded tsd library - dtslint bundles its assertions
- Add explicit test for return type of the package
- Add explicit test that terminate() returns void
- No need to fully qualify type names

### DIFF
--- a/bin/wasm-node/javascript/src/index.test-d.ts
+++ b/bin/wasm-node/javascript/src/index.test-d.ts
@@ -13,5 +13,7 @@ const sp = smoldot.start({
 
 sp.then(sm => {
   // $ExpectType void
+  sm.send_json_rpc('{"id":8,"jsonrpc":"2.0","method":"system_health","params":[]}');
+  // $ExpectType void
   sm.terminate();
 });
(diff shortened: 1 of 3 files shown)


bin/wasm-node/javascript/src/index.test-d.ts
```diff
@@ -1,15 +1,19 @@
-import { expectType } from 'tsd';
-import smoldot from 'smoldot';
+import smoldot from 'smoldot'; // smoldot is Smoldot
```
tomaka, line 1: What does `smoldot is Smoldot` mean?

tomaka (approved):

bin/wasm-node/javascript/src/index.test-d.ts
raoulmillais: It is a type assertion that means:

 "Is the variable `smoldot` that is returned by the import statement of the type `Smoldot`?"  where the type `Smoldot` is the [type of the default export of the package](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/javascript/src/index.d.ts#L41-L47)

raoulmillais: I.E. when `dtslint` parses `index.test-d.ts` it checks that the AST of the type that gets imported from the javascript file matches the interface defined as the export in the typescript typings file.  I missed an explicit test for that before.  Now we will get errors if we add to the public API but don't update the typing file

tomaka: Oh, so this comment is actually interpreted by `dtslint`?

Issue #366, opened 2021-01-15
Implement listening for GrandPa commit messages


tomaka:
The light client needs to listen to GrandPa commit messages (which are being gossiped around) to determine which blocks get finalized.

tomaka:
Rough code design draft:

A good part of the GrandPa logic should go in the `src/network` module.
The `network/service` code should get some sort of `set_grandpa_info(authorities_set_id: u??, round_number: u??)` method that sends out GrandPa neighbour packets.

The `network` module should also verify the commit messages (or give ways to verify commit messages, if directly verifying isn't possible) and report that.

The `chain/sync/all` module should get a `set_finalized_block` method that updates the locally-known finalized block.
The "outcome" objects reported by `chain/sync/all` should also indicate changes in the authorities set.

PR #702, opened 2021-04-01
Make all_forks ask for justifications when doing ancestry searches

tomaka:
Relates to #366 

In order to make finality with with the `all_forks` strategy, we actually need two mechanisms: GrandPa commit messages and justifications.

Importantly, GrandPa commit messages alone aren't sufficient because nodes don't send them to nodes that aren't in the same authority set id. With only GrandPa commit messages, the finality of a node could get stuck if it loses Internet connectivity at the wrong moment.

This means that even if we implement listening for GrandPa commit messages (#366), we still need to implement justifications.
This is partly what this PR does.

### DIFF
--- a/src/sync/all_forks.rs
+++ b/src/sync/all_forks.rs
@@ -174,10 +174,15 @@ struct Inner<TSrc> {
     /// Also contains a list of ongoing requests. The requests are kept up-to-date with the
     /// information in [`Inner::sources`].
     /// For each request, contains as user data the source that has emitted it.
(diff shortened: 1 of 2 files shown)


src/sync/all_forks.rs
```diff
@@ -452,6 +465,8 @@ impl<TSrc, TBl> AllForksSync<TSrc, TBl> {
             return None;
         }
 
+        // TODO: need to periodically query for justifications of non-finalized blocks that change GrandPa authorities
```
tomaka, line 468: This TODO is rather important, but quite complicated to do.

PR #716, opened 2021-04-06
Fix GRANDPA commit and catchup message

FlorianFranzen:
Use unsigned pre-commits in compact commit message (see [grandpa crate](https://github.com/paritytech/finality-grandpa/blob/master/src/lib.rs#L318)) and add authority set id to CatchUp message (see [substrate](https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/communication/gossip.rs#L398)).

### DIFF
--- a/src/network/protocol/grandpa.rs
+++ b/src/network/protocol/grandpa.rs
@@ -123,6 +123,7 @@ pub struct CatchUpRequest {
 
 #[derive(Debug, Clone)]
 pub struct CatchUpRef<'a> {
+    pub set_id: u64,
     pub round_number: u64,
     pub prevotes: Vec<PrevoteRef<'a>>,
     pub precommits: Vec<PrecommitRef<'a>>,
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks! When experimenting I realized that we failed to decode messages from the network, but I had not managed to figure out why.

FlorianFranzen:
No problem, I was updating the spec and realized some discrepancies. I discovered that you do not currently implement [versioning of the neighbor messages](https://github.com/paritytech/substrate/blob/master/client/finality-grandpa/src/communication/gossip.rs#L372-L375), but I am not sure how you would like to implement that on your side. An enum like substrate would probably be the cleanest.

tomaka:
> I also discovered that you do not currently implement versioning of the neighbor messages,

Well, I'm going under the assumption that we'll stay at version 1:
https://github.com/paritytech/smoldot/blob/c5b4271f8b777d0820b4ff8e8b72b8efc91040c7/src/network/protocol/grandpa.rs#L354

If version 2 ever happens, we can always add an enum later.

FlorianFranzen:
@tomaka Fixed all build issue and rebased on latest `main`.

tomaka:
Thank you

tomaka (approved):

PR #755, opened 2021-04-11
Refactor all_forks to allow body requests

tomaka:
This PR heavily refactors the `all_forks` syncing code.

Instead of having an API that looks like `fn do_something(&mut self) -> WhatHasChanged`, the API now looks more like `fn do_something(&mut self); fn what_has_changed(&self) -> ...`.
This makes the code considerably easier both to read and to write.


tomaka:
To do before merging:

- [ ] Update documentation of `pending_blocks`.
- [ ] Remove uninteresting blocks if the number of disjoint blocks is too high.
- [ ] Fix doc tests.

Issue #769, opened 2021-04-14
`InProgressGrandpaWarpSync` is missing `remove_source`


tomaka:
At the moment, removing a source (i.e. disconnecting from a peer) reaches a `todo!()` when we're doing a GrandPa warp sync.

The GrandPa warp sync state machine should support removing a source at any moment.

PR #770, opened 2021-04-14
Add a `remove_source` function to `InProgressGrandpaWarpSync`

expenses:
Closes https://github.com/paritytech/smoldot/issues/769.

### DIFF
--- a/src/sync/grandpa_warp_sync.rs
+++ b/src/sync/grandpa_warp_sync.rs
@@ -601,7 +601,7 @@ impl<TSrc> PostVerificationState<TSrc> {
                 ),
             )
         } else {
-            (removed, StateRemoveSourceResult::PostVerification(self))
+            (removed, StateRemoveSourceResult::RemovedOther(self))
         }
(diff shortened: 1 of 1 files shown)


tomaka (approved):
Haven't checked the logic super closely, but looks good

(EDIT: except it doesn't compile)

expenses:
How do you recommend I implement `remove_source` in all sync? We want to do something like this:
```rust
(IdleInner::GrandpaWarpSync(sync), SourceMapping::GrandpaWarpSync(src)) => {
    let (removed, grandpa_warp_sync) = sync.remove_source(src);
    *sync = grandpa_warp_sync;
    (Vec::new(), removed.user_data)
}
```

but we can't because that involves moving `sync`. Should I use `std::mem::replace` with some default value of `InProgressGrandpaWarpSync`?

tomaka:
Implementing the thing in `all.rs` is kind of complicated because we have to indicate which ongoing request has been cancelled.

PR #783, opened 2021-04-22
Remove obsolete TODOs regarding RSA

tomaka:
After https://github.com/libp2p/specs/pull/316, support for RSA is no longer mandatory in the libp2p specifications.

### DIFF
--- a/src/libp2p/peer_id.rs
+++ b/src/libp2p/peer_id.rs
@@ -28,14 +28,12 @@ mod keys_proto {
 
 /// Public key of a node's identity.
 ///
-/// > **Note**: This enum isn't fully compatible with the libp2p specifications, as support for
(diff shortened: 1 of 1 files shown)


src/libp2p/peer_id.rs
```diff
@@ -28,14 +28,12 @@ mod keys_proto {
 
 /// Public key of a node's identity.
 ///
-/// > **Note**: This enum isn't fully compatible with the libp2p specifications, as support for
-/// >           RSA is mandatory.
-// TODO: add support for RSA ^
+/// Libp2p specifies multiple different possible algorithms, but  only ed25519 support is
```
tomaka, line 31:
```suggestion
/// Libp2p specifies multiple different possible algorithms, but only ed25519 support is
```

PR #801, opened 2021-05-04
Fix `babe_finalized_epoch` meta key never being inserted

### LINKED ISSUE #766: Occasional error when loading existing database
Sometimes happens:

```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `23552`,
 right: `0`', /home/null/data/stow/code/code/smoldot/src/chain/blocks_tree.rs:125:17
```

tomaka:
Fix #766

As documented, the `babe_finalized_epoch` key in the `meta` table is missing if the chain doesn't use Babe or if the finalized block is 0.

However, we were using `UPDATE` to actually set its value once needed. If the key is missing, it would therefore never be inserted.


tomaka:
Couldn't reproduce #766 anymore after this PR.

PR #805, opened 2021-05-06
Enable discovery on the Wasm light client

tomaka:
Can't merge now as it seems to trigger some kind of infinite loop

### DIFF
--- a/bin/wasm-node/rust/src/network_service.rs
+++ b/bin/wasm-node/rust/src/network_service.rs
@@ -38,7 +38,7 @@
 
 use crate::ffi;
 
-use core::{num::NonZeroUsize, pin::Pin, time::Duration};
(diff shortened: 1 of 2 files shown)


bin/wasm-node/rust/src/network_service.rs
```diff
@@ -360,11 +365,52 @@ impl NetworkService {
...
+                    }
+                }
+            }));
+        }
+
         (network_service.guarded.try_lock().unwrap().tasks_executor)(Box::pin({
             // TODO: keeping a Weak here doesn't really work to shut down tasks
             let network_service = Arc::downgrade(&network_service);
             async move {
                 loop {
```
tomaka, line 410: The infinite loop mentioned in OP was here.

PR #820, opened 2021-05-11
Fix panic when attempting to verify empty proof

tomaka:


tomaka:
Going to merge as insubstantial.

Issue #731, opened 2021-04-07
Properly choose request targets


tomaka:
When we want to know for example the value of a storage key at a certain block, we should only query a node that knows about this specific block.

There are already a lot of TODOs in the code related to this, but an issue gives this problem more visibility.

PR #828, opened 2021-05-12
Work towards properly querying nodes that know about desired block

tomaka:
This PR performs a lot of progress towards https://github.com/paritytech/smoldot/issues/731

- All the methods on `NetworkService` that apply on the network as a whole and query things about a block are now moved to `SyncService`, as the `SyncService` is the piece of code that tracks which peer has which block.

- Adds `AllSync::knows_block` and `AllSync::source_knows_block` methods.

- Tweak the similar methods in `sources.rs` to assume that sources always know about all blocks that are part of the finalized chain if their best block is superior to the desired block.

### DIFF
--- a/bin/wasm-node/rust/src/json_rpc_service.rs
+++ b/bin/wasm-node/rust/src/json_rpc_service.rs
@@ -573,13 +573,17 @@ impl JsonRpcService {
                 let mut lock = self.blocks.lock().await;
 
                 let block_hash = lock.best_block;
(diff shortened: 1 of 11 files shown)


tomaka:
I just realized that this is actually not correct.
`sources.rs` and `all_sync.rs` should not return `true` for all blocks below the finalized block, as nothing actually says that the requested block is actually supposed to be part of the finalized chain.

tomaka:
Fixed.

PR #835, opened 2021-05-13
Revamp the public JavaScript API

tomaka:
cc @raoulmillais @wirednkod

This PR revamps the public JavaScript API:

- All the `snake_case` methods and fields now use `camelCase`, since that's what is idiomatic in JavaScript. `chain_spec` has been renamed to `chainSpecs` (with an `s`) because of the next point.

- Instead of passing one chain spec for the relay chain (and one for the parachain), you can now pass an array of chain specs (of arbitrary length). When you call `sendJsonRpc`, you should pass the index, within that array, of the chain that is concerned. This index is also passed back in return in the JSON-RPC callback.

See the changes to `demo.rs` to know how to propagate these changes to substrate-connect.

I'll publish 0.2.0 afterwards.

### DIFF
--- a/bin/wasm-node/rust/src/ffi.rs
+++ b/bin/wasm-node/rust/src/ffi.rs
@@ -25,6 +25,7 @@ use core::{
     iter, marker,
     ops::{Add, Sub},
     pin::Pin,
+    ptr,
     slice,
     task::{Context, Poll, Waker},
     time::Duration,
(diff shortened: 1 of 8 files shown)


bin/wasm-node/rust/src/ffi.rs
```diff
@@ -427,38 +427,49 @@ fn alloc(len: u32) -> u32 {
...
+                &chain_specs_pointers[(chain_spec_index * 8 + 4)..(chain_spec_index * 8 + 8)],
+            )
+            .unwrap();
+            usize::try_from(u32::from_le_bytes(val)).unwrap()
+        };
+
+        let chain_spec: Box<[u8]> =
+            unsafe { Box::from_raw(slice::from_raw_parts_mut(spec_pointer as *mut u8, spec_len)) };
+
+        let chain_spec = String::from_utf8(Vec::from(chain_spec)).expect("non-utf8 chain specs");
```
tomaka, line 464:
```suggestion
        let chain_spec = String::from_utf8(Vec::from(chain_spec)).expect("non-utf8 chain spec");
```

PR #742, opened 2021-04-08
Backport support for zstd-compressed runtime blobs

tomaka:
Close #725 

Opening as draft because I'm really not satisfied with the idea of removing `no_std` compatibility in smoldot.

PR #836, opened 2021-05-14
Add chain spec for shell

tomaka:
Doesn't work because depends on https://github.com/paritytech/smoldot/pull/742


tomaka:
I'm going to merge anyway. It doesn't hurt to have this chain spec in the repo even it doesn't work.

Issue #480, opened 2021-02-08
Finish the implementation of the transactions service


tomaka:
See the TODOs in `bin/wasm-node/src/transactions_service.rs`

tomaka:
https://github.com/paritytech/smoldot/pull/845 doesn't completely close this issue, as the implementation isn't 100% finished. It is functional, but not robust.

tomaka:
I'm going to close this as done. While there might be some occasional panics remaining, saying that the implementation is unfinished is clearly not true anymore.

PR #845, opened 2021-05-16
Properly implement transactions service

tomaka:
Close https://github.com/paritytech/smoldot/pull/562
cc https://github.com/paritytech/smoldot/issues/480

This finished implementing the transactions service.
The service listens to all blocks and to best and finalized blocks in order to notify of the transactions that are included, retracted, or finalized.

Not completely finishable because we're missing the primitive in `sync_service.rs` to listen to *all* blocks.


tomaka:
It turns out that this PR is more complicated than initially thought, as the naive approach doesn't work.

With the current code it is impossible to know locally if the transaction to send out is valid or not, which has two consequences.

- We would be punished by the network when sending out a bad transaction.
- In the current PR we would re-announce the transaction periodically until it is included in a block. But if the transaction is invalid it will never be included, and we don't know when to stop re-announcing the transaction.

Despite being a light client we will need a full transactions pool.

tomaka:
> Despite being a light client we will need a full transactions pool.

Actually, no.

A full transactions pool also needs to validate every single transaction that comes in every single block, which would result in a lot of storage proof requests.

Instead, a good compromise is to validate each transaction submitted by the user once when it is submitted, then again at the end of its longevity.

PR #846, opened 2021-05-16
Process GrandPa commit messages

tomaka:
Close https://github.com/paritytech/smoldot/issues/366

Implements processing the incoming GrandPa commit messages to update finality.

Doesn't actually work because full nodes don't send commit messages to light clients. In order to test, one must replace `Light` with `Full` in the network service. André is working on changing this in Substrate.
Apart from this, the code actually works well. Most notably, there's no panic in the syncing or anywhere else.

This PR is still work in progress because the commit messages aren't actually verified and just assumed to be correct. The problem I'm facing is that the actual verification code is very close to the code that verifies a justification, but I don't want to copy-paste it and I'm having trouble finding some common abstraction/ground.


tomaka:
Finished the commits verification code.

PR #856, opened 2021-05-24
Propagate change to default heap pages

tomaka:
Update for this sneaky consensus change https://github.com/paritytech/substrate/pull/8892

Let's wait for a few days before merging just in case it gets reverted or something.


tomaka:
Looks like it's here to stay since we've released 0.9.3 with it, and nobody is suggesting a better change at the moment.

Issue #501, opened 2021-02-11
Allow adding and removing chains at runtime in the Wasm node


tomaka:
If the Wasm node is to be used as a browser extension, it's probably a good idea to be able to provide the possibility of connecting and disconnecting from chains after the node has started.

tomaka:
First step should be to modify the networking code to allow adding/removing chains at runtime, as this is the most complicated part of this issue.

tomaka:
> First step should be to modify the networking code to allow adding/removing chains at runtime, as this is the most complicated part of this issue.

And more specifically: adding/removing protocols in the code of the `established` module.
I think that we should instead refactor `established` to not interpret the meaning of substreams, and report low-level events instead.

tomaka:
https://github.com/paritytech/smoldot/pull/994 mostly implemented this, except for the fact that a network service is created for every chain.

PR #874, opened 2021-05-28
Networking refactor

tomaka:
Networking refactoring with the objective of simplifying the code. Will be extremely useful for #501.

After this PR, the `libp2p` module only allows management of connections and substreams.
The `network` module contains all the higher-level logic: request-responses, notifications, ping, etc.

### DIFF
--- a/src/libp2p.rs
+++ b/src/libp2p.rs
@@ -117,7 +117,6 @@ pub mod discovery;
 pub mod peer_id;
 pub mod peerset;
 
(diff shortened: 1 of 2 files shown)


tomaka:
I've changed my mind about this change.
The entire point of designing `libp2p.rs` the way it is today is that no upper layer requires holding a lock to anything, and that the channel of events keeps a reasonable size.

PR #885, opened 2021-06-02
Compare if Wasm node versions match on CI

tomaka:


tomaka:
Can't test this because GitHub actions aren't starting.

PR #889, opened 2021-06-03
Add a transactions pool

tomaka:
Work in progress.

Adds a transactions pool to the code. It is written in a very generic way, and the ordering system (based on tags) that validators use when authoring blocks is intentionally not implemented because it's rather complicated and not useful yet.


tomaka:
The API of `append_block` should be improved before merging.

PR #892, opened 2021-06-04
Make full node use the "all" syncing strategy

tomaka:
The full node would now use the same syncing code as the wasm node, allowing further clean ups, in particular in `optimistic.rs`.


tomaka:
The resulting code is a bit hacky, but I can't do better without refactoring `optimistic.rs`, and refactoring `optimistic.rs` is exactly the reason why I'm opening this PR.

PR #947, opened 2021-06-16
Zero the memory of the Wasm VM

### LINKED ISSUE #946: Zero Wasm VM memory above `__heap_pages`

tomaka:
Fix #946 

Unfortunately the full node syncing speed is now abysmal. Need to figure out what happens.


tomaka:
Apparently Substrate restores what's below `__heap_pages` instead of zero-ing what's above.

src/executor/vm/jit.rs
```diff
@@ -301,6 +311,23 @@ impl JitPrototype {
...
+                    memory.data_size(),
+                    usize::try_from(zero_memory_above).unwrap(),
+                );
+
+                // Soundness: the documentation of wasmtime precisely explains what is safe or not.
+                // Basically, we are safe as long as we are sure that we don't potentially grow the
+                // buffer (which would invalidate the buffer pointer).
+                unsafe {
+                    memory.data_unchecked_mut()[offset..].fill(0);
+                }
```
melekes, line 327: don't we want to set `self.memory_zeroed` to `true` here?

tomaka: `self` is thrown away afterwards so this doesn't matter

melekes (approved):

tomaka:
For context I haven't merged this, because the last time I tried these changes considerably slows down the blocks verification speed of the full node (something like 50 times slower if I remember correctly), which means that I'm probably missing something that Substrate does or doesn't do.

PR #948, opened 2021-06-16
Refactor RuntimeService::recent_best_block_runtime_call

tomaka:
The current API of this function is very dumb and doesn't allow for example combining multiple calls made on the same runtime.


tomaka:
All three calls (metadata, account nonce, and paraheads) tested and working.

PR #1020, opened 2021-07-20
Update to libsecp256k1 v0.6.0

tomaka:
Before this PR, some invalid signatures might have been erroneously accepted. The change to `parse_standard_slice` fixes that.
In theory we should try sync Polkadot from scratch to make sure that everything works, but in practice the chances of that breaking are almost 0.


tomaka:
Added some TODOs regarding the usage of features

cc https://github.com/paritytech/libsecp256k1/issues/77

PR #1032, opened 2021-07-21
Allow asking for chain health

raoulmillais:
Introduces a new `health()` method to the `SmoldotChain` interface that returns back the health of the chain asynchronously.


bin/wasm-node/javascript/src/index.js
```diff
@@ -79,6 +104,34 @@ export async function start(config) {
       // `expected` was pushed by the `addChain` method.
       // Resolve the promise that `addChain` returned to the user.
       expected.resolve({
+        health: () => {
+          // are we already waiting for a health message reply?
+          if (chainHealthRequestCallbacks.chainId) {
+            const reject = chainHealthRequestCallbacks.chainId.reject;
+            delete chainHealthRequestCallbacks.chainId;
+            return reject(new Error('Called health before pending health request replied'));
+          }
```
raoulmillais, line 113: Hmm... When the first request does eventually get replied to, it will just be passed through to the consumer as if it was a regular RPC response

raoulmillais: In SmoldotProvider that will just result in a debug log because it won't know about the ID: https://github.com/paritytech/substrate-connect/blob/master/packages/connect/src/SmoldotProvider/SmoldotProvider.ts#L147

bin/wasm-node/javascript/src/index.js
```diff
@@ -79,11 +106,40 @@ export async function start(config) {
...
+          // are we already waiting for a health message reply?
+          if (chainHealthRequestCallbacks[chainId]) {
+            const reject = chainHealthRequestCallbacks[chainId].reject;
+            delete chainHealthRequestCallbacks[chainId];
+            return reject(new Error('Called health before pending health request replied'));
+          }
+
+          // craft a new system_health messsage
+          const id = uuidv4();
+          const request = `{"id":"${id}","jsonrpc":"2.0","method":"system_health","params":[]}`;
```
tomaka, line 119: I'd prefer not doing weird string constructions.

```suggestion
          const request = JSON.stringify({ id, jsonrpc: "2.0", method: "system_health", params: [] });
```

bin/wasm-node/javascript/src/index.js
```diff
@@ -16,29 +16,33 @@
...
+    console.debug(fmt(target, message));
+  } else {
+    console.trace(fmt(target, message));
+  }
+}
+
 export async function start(config) {
   config = config || {};
 
-  const logCallback = config.logCallback || ((level, target, message) => {
```
tomaka, line 29: I don't think extracting this improves the code readability.

bin/wasm-node/javascript/src/index.js
```diff
@@ -79,11 +106,40 @@ export async function start(config) {
       // `expected` was pushed by the `addChain` method.
       // Resolve the promise that `addChain` returned to the user.
       expected.resolve({
+        health: () => {
+          // are we already waiting for a health message reply?
+          if (chainHealthRequestCallbacks[chainId]) {
+            const reject = chainHealthRequestCallbacks[chainId].reject;
+            delete chainHealthRequestCallbacks[chainId];
+            return reject(new Error('Called health before pending health request replied'));
```
tomaka, line 114: This feels very arbitrary. Why would it be forbidden to call `health()` multiple times in parallel?

bin/wasm-node/javascript/src/index.js
```diff
@@ -79,11 +106,40 @@ export async function start(config) {
...
+          worker.postMessage({ ty: 'request', request, chainId });
+          return hrPromise;
+        },
         sendJsonRpc: (request) => {
           if (workerError)
             throw workerError;
           if (!chainsJsonRpcCallbacks[message.chainId])
             throw new SmoldotError('Chain isn\'t capable of serving JSON-RPC requests');
+
           worker.postMessage({ ty: 'request', request, chainId });
```
tomaka, line 143: I think we should have a completely different namespace for regular JSON-RPC requests vs health requests.

Something like that:

```suggestion
          const parsed = JSON.parse(request);
          const newId = 'user:' + JSON.stringify(parsed.id);
          request.id = newId;
          worker.postMessage({ ty: 'request', request, chainId });
```

bin/wasm-node/javascript/src/index.js
```diff
@@ -79,11 +106,40 @@ export async function start(config) {
...
+        health: () => {
+          // are we already waiting for a health message reply?
+          if (chainHealthRequestCallbacks[chainId]) {
+            const reject = chainHealthRequestCallbacks[chainId].reject;
+            delete chainHealthRequestCallbacks[chainId];
+            return reject(new Error('Called health before pending health request replied'));
+          }
+
+          // craft a new system_health messsage
+          const id = uuidv4();
```
tomaka, line 118: (continuing the serialization)

```suggestion
          const id = 'health:' + nextHealthRequestId;
          nextHealthRequestId += 1;
```

bin/wasm-node/javascript/src/index.js
tomaka: Needs a `try ... catch` around `JSON.parse` which ignores invalid requests.

bin/wasm-node/javascript/src/index.js
```diff
@@ -49,22 +53,45 @@ export async function start(config) {
...
-  let chainsJsonRpcCallbacks = {};
+  const chainsJsonRpcCallbacks = {};
 
   // The worker can send us messages whose type is identified through a `kind` field.
   workerOnMessage(worker, (message) => {
+    console.log(message);
     if (message.kind == 'jsonrpc') {
+
+      // check if this was a response to a health request first
+      const msgdata = JSON.parse(message.data);
```
tomaka, line 79: Something like:

```js
if (msgdata.id.startsWith('health')) {
} else if (msgdata.id.startsWith('user')) {
}
```

tomaka (changes requested):

bin/wasm-node/javascript/src/index.js
raoulmillais: You were creating the callback inside the function which potentially leads to multiple copies of the same function

bin/wasm-node/javascript/src/index.js
raoulmillais: Sure but why would you make multiple health requests to the same chain in parallel? How about replacing the request when subsequent requests are made instead? Tracking multiple concurrent requests adds more complexity

bin/wasm-node/javascript/src/index.js
tomaka: Is there a downside in having multiple copies of the same function?

bin/wasm-node/javascript/src/index.js
tomaka: "why would anyone use this API in a way that is different from the way I use it" is generally what leads to bugs and clusterfucks.

bin/wasm-node/javascript/src/index.js
raoulmillais: its not just the function it's a closure so its references to everything in its outerscope.  I dont feel strongly about it though but in general avoiding creating closures where they're not needed is a considered a good practice in javascript because that's the most common source of memory leaks.

bin/wasm-node/javascript/src/index.js
raoulmillais: yeah you're right :+1: now I think about it not tracking all the requests is also a bad idea because users will end up with pending (never to be resolved) promises

raoulmillais: Done

bin/wasm-node/javascript/src/index.js
```diff
@@ -49,24 +63,55 @@ export async function start(config) {
   const worker = new Worker(new URL('./worker.js', import.meta.url));
   let workerError = null;
 
+  // map of chain ids to health requests - their ids and the their promise resolve and reject funtions
+  // { 1: [{ requestId: <n>, resolve: fn(), reject: fn() }, /* ... */ }]
+  const chainHealthRequestCallbacks =  {};
```
tomaka, line 68: This doesn't need to track the `chainId`. It can just directly track the values generated from `nextHealthRequestId`.

tomaka: You need to track the `chainId` internally, though, and clean up entries when `remove()` is called on a chain.

raoulmillais: Yeah that also makes the code much neater

raoulmillais: Done

bin/wasm-node/javascript/src/index.js
```diff
@@ -79,18 +119,54 @@ export async function start(config) {
...
+          nextHealthRequestId += 1;
+          const request = JSON.stringify({ id, jsonrpc: "2.0", method: "system_health", params: [] });
+          let hrResolve, hrReject;
+          const hrPromise = new Promise((resolve, reject) => {
+            hrResolve = resolve;
+            hrReject = reject;
+          });
+
+          // track the new message
+          healthRequestCallbacks.push({
```
tomaka, line 134: What I was thinking of is:

```suggestion
healthRequestCallbacks[id] = {
    chainId,
    resolve: hrResolve,
    reject: hrReject
});
```

raoulmillais: Changing to this shape would make the cleanup code in `remove()` a lot more complicated.

raoulmillais: How about using a proper [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instead of an object since that's really what we're doing?

tomaka: Yeah that works

PR #1096, opened 2021-08-07
Publish v0.3.7 and as @substrate/smoldot

tomaka:


tomaka:
Also bumping the version at the same time, as it could be confusing it we published a @substrate/smoldot version 0.3.6 that isn't the same as smoldot version 0.3.6.

PR #1098, opened 2021-08-07
Properly implement payment_queryInfo

tomaka:


tomaka:
This is ready for merging, and normally more commits should have been pushed, but GitHub is having a moment.

PR #1111, opened 2021-08-09
Connect to a parachain in the demo

tomaka:
Draft because it triggers an infinite loop somewhere.

### DIFF
--- a/bin/wasm-node/javascript/demo/demo.js
+++ b/bin/wasm-node/javascript/demo/demo.js
@@ -25,6 +25,7 @@ import * as process from 'process';
 import * as fs from 'fs';
 
 const chainSpec = fs.readFileSync('../../westend.json', 'utf8');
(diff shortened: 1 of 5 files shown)


tomaka:
Infinite loop is caused by https://github.com/paritytech/smoldot/blob/a82ab46a6d1d5d764b56bf571d2af626e27c22e3/bin/wasm-node/rust/src/sync_service.rs#L1314
The transactions service detects that the sender has been dropped and resets.

tomaka:
Seems to fail initializing the chains if connecting too quickly. Should be fixed before merging.

tomaka:
> Seems to fail initializing the chains if connecting too quickly. Should be fixed before merging.

Was caused by using `pop()` instead of `shift()` for the pending confirmations. Can't believe I didn't catch that.

PR #1138, opened 2021-08-17
Remove randomness from health checker

tomaka:
Apparently, importing `randombytes` from the health checker causes issues with some JS thing somewhere, so for the sake of making things move we'll remove this import.

Consequently, the API of the health checker needs to change:

- You no longer need to pass the `sendJsonRpc` function to the `start()` method, but instead pass it to a new `setSendJsonRpc()` method. Call `setSendJsonRpc()` right after initializing the health checker.
- You **must** now use `sendJsonRpc` on the health checker, which modifies it and propagates it to the callback that you passed to `setSendJsonRpc`. Do **not** call `sendJsonRpc` directly on the smoldot chain.

The rest stays the same.

**I also would like to remind that the health checker is documented** as comments to the `healthChecker` function.

I modified `demo.js` in the first commit, in order to showcase how to use the API. The second commit reverts these changes.

### DIFF
--- a/bin/wasm-node/javascript/demo/demo.js
+++ b/bin/wasm-node/javascript/demo/demo.js
@@ -69,13 +69,20 @@ wsServer.on('request', function (request) {
     let chain;
     if (request.resource == '/relay') {
         chain = client.then(async client => {
+            const healthChecker = smoldot.healthChecker();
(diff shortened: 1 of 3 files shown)


wirednkod (commented):
lgtm

PR #1166, opened 2021-08-22
Implement compact trie proofs

### LINKED ISSUE #910: Backport compact trie proofs
https://github.com/paritytech/substrate/pull/8574

tomaka:
Fix https://github.com/paritytech/smoldot/issues/910

PR is ready, but two questions remain to be solved before merging:

- In a compact proofs, are the children of a node present in this node but with a length of 0? Or is the "children section" skipped altogether?
- If the answer to the first question is that children are present, is it possible for some children to not have a length of 0? Or would the proof be invalid?


tomaka:
Oops, I must also verify that hashes match, otherwise it's completely unsecure.

tomaka:
We can still find inline nodes and hashes in the proof. Hashes can only be found for nodes that aren't interesting for what we want to prove.

PR #1240, opened 2021-09-06
Properly implement peer slots

tomaka:


tomaka:
Requires https://github.com/paritytech/smoldot/pull/1245 to be merged first, otherwise we reach a `todo!()` when rejecting substreams.

PR #1258, opened 2021-09-10
Fix paths to specs

wirednkod:


tomaka:
This demo is supposed to be run through `npm start`, which sets the path to the `javascript` directory.

PR #1321, opened 2021-09-17
Update Rococo chain spec

tomaka:
Propagates https://github.com/paritytech/polkadot/pull/3871


tomaka:
The new Rococo fails to start because of new host functions, but that will be in a further PR.

Issue #1377, opened 2021-10-04
Runtime version updates doesn't work for parachains because of subscribe_all unimplemented


tomaka:
https://github.com/paritytech/smoldot/blob/871afd887fbbae4e86047c0452b3721af76a2200/bin/wasm-node/rust/src/sync_service.rs#L1474-L1475

tomaka:
- [ ] Add `RuntimeService::subscribe_all`.
- [ ] Add `RuntimeService::block_runtime_call` that accepts a block hash or something similar.
- [ ] Make `fetch_paraheads` produce a stream of notifications, instead of just a single block, and remove the `track_finalized` parameter.

PR #1398, opened 2021-10-07
Revert subscribing to runtime spec changes in health.js

tomaka:
Reverts the change to `health.js` in introduced in https://github.com/paritytech/smoldot/pull/1349

Doing this change was a hack, and it is no longer necessary after https://github.com/paritytech/smoldot/pull/1380. Instead, the best block subscriptions should come from the runtime service.
I'm opening this PR in order to not forget to do this. It is to be merged after the best block subscriptions are updated to come from the runtime service, which itself can only be done after https://github.com/paritytech/smoldot/issues/1377

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,9 +281,9 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "1.2.1"
+version = "1.3.2"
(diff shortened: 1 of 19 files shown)


tomaka:
Can be done after https://github.com/paritytech/smoldot/pull/1417

PR #1491, opened 2021-10-20
Fix a few unwrap TODO by extracting parse_bootnode

liuchengxu:


liuchengxu:
@tomaka Reverted to the previous panicking behavior.

PR #1493, opened 2021-10-20
Verify correctness of downloaded block bodies and fix all the block bodies format confusions

tomaka:


tomaka:
Any full node database that existed before this PR is now invalid, but we don't care about that at the moment.

PR #1561, opened 2021-10-31
Implement Aura block production in the full node

tomaka:


tomaka:
This PR is pretty useless without support for Babe.

PR #1567, opened 2021-11-01
Allow passing a trie root hash instead of the genesis storage

### LINKED ISSUE #1566: Support a variant of chain specs with genesis trie root hash

tomaka:
Close #1566


tomaka:
Everything is working except the `runtime_service.rs`

PR #1569, opened 2021-11-02
Implement Babe blocks authorship

tomaka:
To do before https://github.com/paritytech/smoldot/pull/1561

### DIFF
--- a/src/author.rs
+++ b/src/author.rs
@@ -18,5 +18,6 @@
 // TODO: doc
 
 pub mod aura;
+pub mod babe;
 pub mod build;
 pub mod runtime;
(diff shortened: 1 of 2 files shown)


kwanCCC:
@tomaka , I'm a new to substrate and I found `smoldot` is a good way to understand the background of `substrate`. I have a question about #1569. Is there any reason why `BABE` has not been implemented:question: Would you tell me If it's convenient for you. By the way, thanks to your great work:rocket:

tomaka:
It's not been implemented because nobody implemented it. It's as simple as that.
Full node and block authoring capabilities are not the primary goal of smoldot, so they're low priority.

tomaka:
This PR is missing the code that actually determines when a Babe slot is available.

PR #1574, opened 2021-11-04
Fix behavior and return value of clear_prefix_version_2

tomaka:
Should first be tested by fully syncing a node from scratch.


tomaka:
> Should first be tested by fully syncing a node from scratch.

Actually no because the `clear_prefix_version_2` function wasn't working properly anyway, and the full node has never synced to any chain that uses it.

PR #1666, opened 2021-11-23
Don't use parity_scale_codec in the crypto recover host functions

tomaka:
Need to sync a chain before merging, to make sure no mistake was made.


tomaka:
Synced to block ~200k without any problem

PR #1672, opened 2021-11-24
Add a storage overlay struct

tomaka:


tomaka:
Seems to work, as the node syncs blocks successfully. Need to test syncing a long chain before merging.

PR #1710, opened 2021-11-30
smoldot light improvements

josepot:
As I was trying to understand why there is a test that's not working on the PR that tries to [upgrade `smoldot-light` on `substrate-connect`](https://github.com/paritytech/substrate-connect/pull/555), I've noticed 2 issues on `smoldot-light`:

- The type definitions exported on the `index.d.ts` file don't actually match what's being exported. This must have been the case for quite some time now, since [substrate-connect is abusing `any` in order to use `smoldot-light`](https://github.com/paritytech/substrate-connect/blob/master/projects/extension/src/background/ConnectionManager.ts#L219), which pretty much defeats the purpose of the typings file. The first commit addresses that.
- The second commit improves the behavior of `client.terminate`, because the current behavior causes weird memory-leaks and race conditions on some tests (because jest has to force-quite some tasks). That was happening mainly because `worker.terminate()` behaves differently between NodeJS (where it returns a Promise) and the DOM (where it returns `undefined`). I've also taken that opportunity to cleanup a possible timeout callback.

### DIFF
--- a/bin/wasm-node/javascript/src/compat-browser.js
+++ b/bin/wasm-node/javascript/src/compat-browser.js
@@ -21,5 +21,6 @@ export const net = null;
 export const Worker = typeof window != 'undefined' ? window.Worker : null;
 export const workerOnMessage = (worker, callback) => { worker.onmessage = (event) => callback(event.data) };
 export const workerOnError = (worker, callback) => { worker.onerror = callback; };  // TODO: unclear if the parameter of the callback is same as with NodeJS
+export const workerTerminate = (worker) => { worker.terminate(); return Promise.resolve(); };
 export const postMessage = (msg) => self.postMessage(msg);
 export const setOnMessage = (callback) => { self.onmessage = (event) => callback(event.data) };
--- a/bin/wasm-node/javascript/src/compat-nodejs.js
+++ b/bin/wasm-node/javascript/src/compat-nodejs.js
@@ -26,5 +26,6 @@ export { Worker } from 'worker_threads';
 
 export const workerOnMessage = (worker, callback) => worker.on('message', callback);
 export const workerOnError = (worker, callback) => worker.on('error', callback);
+export const workerTerminate = (worker) => worker.terminate().then(() => {});
 export const postMessage = (msg) => parentPort.postMessage(msg);
 export const setOnMessage = (callback) => parentPort.on('message', callback);
--- a/bin/wasm-node/javascript/src/index.js
+++ b/bin/wasm-node/javascript/src/index.js
@@ -15,7 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-import { Worker, workerOnError, workerOnMessage } from './compat-nodejs.js';
+import { Worker, workerOnError, workerOnMessage, workerTerminate } from './compat-nodejs.js';
 export * from './health.js';
 
 export class AlreadyDestroyedError extends Error {
@@ -230,9 +230,14 @@ export function start(config) {
       return chainAddedPromise;
     },
     terminate: () => {
-      worker.terminate();
-      if (!workerError)
-        workerError = new AlreadyDestroyedError();
+      if (workerError)
+        return Promise.reject(workerError)
+      workerError = new AlreadyDestroyedError();
+
+      if (livenessTimeout !== null)
+        clearTimeout(livenessTimeout)
+
+      return workerTerminate(worker)
     }
   }
 }
--- a/bin/wasm-node/javascript/src/index.d.ts
+++ b/bin/wasm-node/javascript/src/index.d.ts
@@ -80,7 +80,7 @@ export interface Client {
    * @throws {AlreadyDestroyedError} If the client has already been terminated earlier.
    * @throws {CrashError} If the background client has crashed.
    */
-  terminate(): void;
+  terminate(): Promise<void>;
 }
 
 /**
--- a/bin/wasm-node/javascript/src/index.d.ts
+++ b/bin/wasm-node/javascript/src/index.d.ts
@@ -18,21 +18,21 @@
 /**
  * Thrown in case of a problem when initializing the chain.
  */
-declare class AddChainError extends Error {
+export class AddChainError extends Error {
   constructor(message: string);
 }
 
 /**
  * Thrown in case the API user tries to use a chain or client that has already been destroyed.
  */
-declare class AlreadyDestroyedError extends Error {
+export class AlreadyDestroyedError extends Error {
 }
 
 /**
  * Thrown when trying to send a JSON-RPC message to a chain whose JSON-RPC system hasn't been
  * enabled.
  */
-declare class JsonRpcDisabledError extends Error {
+export class JsonRpcDisabledError extends Error {
 }
 
 /**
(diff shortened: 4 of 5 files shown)


tomaka:
`index.test-d.ts` also needs an update.

bin/wasm-node/javascript/src/index.js
```diff
@@ -230,9 +230,17 @@ export function start(config) {
...
+      clearTimeout(livenessTimeout)
+
+      // On NodeJS `worker.terminate()` returns a Promise, while on the DOM
+      // it returns `undefined`. In order to have a consistent API we will
+      // always return a Promise
+      const terminationPromise = worker.terminate() || Promise.resolve()
+
+      return terminationPromise.then(() => {
+        if (!workerError)
+          workerError = new AlreadyDestroyedError();
```
tomaka, line 242: This needs to be done before we call `worker.terminate()`. The goal is to prevent people from using the `Client` after they have called `terminate()`, which they could still do if you put this in the `then()`.

bin/wasm-node/javascript/src/index.js
```diff
@@ -230,9 +230,17 @@ export function start(config) {
...
     terminate: () => {
-      worker.terminate();
-      if (!workerError)
-        workerError = new AlreadyDestroyedError();
+      clearTimeout(livenessTimeout)
+
+      // On NodeJS `worker.terminate()` returns a Promise, while on the DOM
+      // it returns `undefined`. In order to have a consistent API we will
+      // always return a Promise
+      const terminationPromise = worker.terminate() || Promise.resolve()
```
tomaka, line 238: Could you do this in `compat-node.js` and `compat-browser.js` instead?
Something like `export const workerTerminate = ...` similar to what there already is.

tomaka (changes requested):

bin/wasm-node/javascript/src/index.js
```diff
@@ -230,9 +230,17 @@ export function start(config) {
       return chainAddedPromise;
     },
     terminate: () => {
-      worker.terminate();
-      if (!workerError)
-        workerError = new AlreadyDestroyedError();
+      clearTimeout(livenessTimeout)
```
tomaka, line 233:
```suggestion
      if (livenessTimeout !== null)
        clearTimeout(livenessTimeout)
```

I know if shouldn't change anything, but I'm not a fan of calling functions with bad parameters because the function says it's ok.

bin/wasm-node/javascript/src/index.js
josepot: Yeah! That's what I thought, but the previous code was doing it after termination, so I didn't want to change too much... But I will make sure that get's fixed, too. Thanks!

bin/wasm-node/javascript/src/index.js
josepot: mmmm, sure let me give that a shot, that makes sense!

bin/wasm-node/javascript/src/index.js
josepot: The [spec says it's ok, though](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-cleartimeout)... But I don't mind changing that.

bin/wasm-node/javascript/src/index.js
josepot: done!

bin/wasm-node/javascript/src/index.js
josepot: done!

bin/wasm-node/javascript/src/index.js
```diff
@@ -230,9 +230,11 @@ export function start(config) {
       return chainAddedPromise;
     },
     terminate: () => {
-      worker.terminate();
-      if (!workerError)
-        workerError = new AlreadyDestroyedError();
+      if (workerError) return Promise.reject(workerError)
+      workerError = new AlreadyDestroyedError();
+
+      if(livenessTimeout) clearTimeout(livenessTimeout)
```
tomaka, line 236:
```suggestion
      if (livenessTimeout !== null)
        clearTimeout(livenessTimeout)
```

bin/wasm-node/javascript/src/index.js
```diff
@@ -230,9 +230,11 @@ export function start(config) {
       return chainAddedPromise;
     },
     terminate: () => {
-      worker.terminate();
-      if (!workerError)
-        workerError = new AlreadyDestroyedError();
+      if (workerError) return Promise.reject(workerError)
```
tomaka, line 233:
```suggestion
      if (workerError)
        return Promise.reject(workerError)
```

bin/wasm-node/javascript/src/compat-nodejs.js
```diff
@@ -26,5 +26,10 @@ export { Worker } from 'worker_threads';
 
 export const workerOnMessage = (worker, callback) => worker.on('message', callback);
 export const workerOnError = (worker, callback) => worker.on('error', callback);
+
+// On NodeJS `worker.terminate()` returns a Promise, while on the DOM
+// it returns `undefined`. In order to have a consistent API we will
+// always return a Promise<void>
+export const workerTerminate = (worker) => worker.terminate().then(() => {});
```
tomaka, line 33:
```suggestion
export const workerTerminate = (worker) => worker.terminate().then(() => {});
```

bin/wasm-node/javascript/src/compat-browser.js
```diff
@@ -21,5 +21,13 @@ export const net = null;
...
 export const workerOnMessage = (worker, callback) => { worker.onmessage = (event) => callback(event.data) };
 export const workerOnError = (worker, callback) => { worker.onerror = callback; };  // TODO: unclear if the parameter of the callback is same as with NodeJS
+
+// On NodeJS `worker.terminate()` returns a Promise, while on the DOM
+// it returns `undefined`. In order to have a consistent API we will
+// always return a Promise<void>
+export const workerTerminate = (worker) => {
+  worker.terminate();
+  return Promise.resolve();
+};
```
tomaka, line 31:
```suggestion
export const workerTerminate = (worker) => { worker.terminate(); return Promise.resolve(); };
```

tomaka (approved):
Nits :rainbow: 
Code style is a bit arbitrary and doesn't matter too much, but since it's mostly me working on this, I think it makes sense if uses the style I'm used to.

josepot:
> Nits rainbow Code style is a bit arbitrary and doesn't matter too much, but since it's mostly me working on this, I think it makes sense if uses the style I'm used to.

There are very few things in life that I hate more than arguing over code-style on a PR, I've reached a point in my life where I don't give a f**k about code style... Well, I actually do, but not to the point to argue over it on a PR.

That's why I love using [prettier](https://prettier.io/) in all my projects, I just code and then prettier handles the code-style for me... It also makes me a lot more productive not to have to worry about code-style as I'm coding. Do I like what prettier does to my code? Of course not, no one likes what prettier does to our code... However, we all love what it does to other's ppl code, and that's way more important :smile:, and on top of that we avoid bikeshedding on PRs :wink: 

Can I send a PR adding prettier to `smoldot-light`? :pray:

tomaka:
> Can I send a PR adding prettier to smoldot-light? pray

Yeah, go head.

PR #1729, opened 2021-12-02
Compile the Wasm node with -Oz

tomaka:
Uses new feature from Rust 1.57.0


tomaka:
Despite the report showing +671kiB, this results in almost one megabyte less after passing the wasm through `wasm-opt`.

tomaka:
The `cargo-deny` Docker image is now up-to-date with Cargo 1.57

PR #1746, opened 2021-12-05
Test whether the twiggy diff actually works

tomaka:
This PR isn't intended to be merged.
I want to test whether the twiggy diff actually works by removing almost all the code of the node.


tomaka:
Seems to work

PR #1749, opened 2021-12-06
Small documentation fix in tx validation

tomaka:


tomaka:
How lengthening an error message can decrease the size of the output is a bit beyond me.

PR #1758, opened 2021-12-08
Restore timestamps after checking out

tomaka:


tomaka:
Let's give up. This doesn't seem to actually be necessary.

PR #1763, opened 2021-12-08
Remove the dependencies on multiaddr and multihash

tomaka:


tomaka:
Size decrease: 5.1 MiB uncompressed / 2.3 MiB compressed to 4.9 MiB uncompressed / 2.2 MiB compressed.

PR #1766, opened 2021-12-08
Publish v0.5.7

tomaka:


tomaka:
How the hell does bumping the version number decrease the size.

Issue #1769, opened 2021-12-10
Add finalized block runtime code and heap_pages to wasm node database


tomaka:
The wasm node database should contain the finalized block runtime code and heap pages.
Additionally, we should start by querying the hash of the runtime code from the network, instead of the value, and compare it against the hash stored in the database.
This would save the user from downloading the runtime if it hasn't changed since last time.

tomaka:
Thanks to https://github.com/paritytech/smoldot/pull/1773 we can now be sure that adding the runtime to the database will not make it exceed the maximum allowed size for local storage.

tomaka:
> querying the hash of the runtime code from the network, instead of the value

This is the major remaining item in this issue.

PR #1787, opened 2021-12-15
Add RuntimeService::block_runtime_code_heap_pages

tomaka:
cc https://github.com/paritytech/smoldot/issues/1769

This PR will fail CI because of an unused function warning, but it paves the way for #1769.

### DIFF
--- a/bin/light-base/src/runtime_service.rs
+++ b/bin/light-base/src/runtime_service.rs
@@ -495,6 +495,41 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
         subscription.next().await.unwrap()
     }
 
+    /// Returns the value of the `:code` and `:heap_pages` storage items of the given block.
+    ///
+    /// Returns `None` if the runtime of that block isn't known by the runtime service, for
+    /// example because the block itself isn't known or because the runtime hasn't been downloaded
+    /// yet.
+    pub async fn block_runtime_code_heap_pages(
+        &self,
+        block_hash: &[u8; 32],
+    ) -> Option<(Option<Vec<u8>>, Option<Vec<u8>>)> {
+        let guarded = self.guarded.lock().await;
+        let guarded = &*guarded;
+
+        let runtime_index = match &guarded.tree {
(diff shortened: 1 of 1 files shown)


tomaka:
Don't ask me why adding an unused method increases the size of the binary.

bin/light-base/src/runtime_service.rs
```diff
@@ -495,6 +495,41 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
         subscription.next().await.unwrap()
     }
 
+    /// Returns the value of the `:code` and `:heap_pages` storage items of the given block.
+    ///
+    /// Returns `None` if the runtime of that block isn't known by the runtime service, for
```
melekes, line 500: looks like this fn always returns `Some(...)`

bin/light-base/src/runtime_service.rs
```diff
@@ -495,6 +495,41 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
...
+                tree: Some(tree), ..
+            } => *tree
+                .input_iter_unordered()
+                .find(|block| block.user_data.hash == *block_hash)
+                .and_then(|block| block.async_op_user_data)?,
+            GuardedInner::FinalizedBlockRuntimeUnknown { tree: Some(tree) } => tree
+                .input_iter_unordered()
+                .find(|block| block.user_data.hash == *block_hash)
+                .and_then(|block| block.async_op_user_data)?
+                .unwrap(),
```
melekes, line 525: is it always safe to unwrap here?

bin/light-base/src/runtime_service.rs
tomaka: There are two `?`s within the body of the function that can make it return `None`

melekes: Oh, right. Sorry

tomaka:
I'm going to close this. It's only 40 lines of code with a merge conflict. While these lines of code may be useful later, it's better to clean up the list of PRs at this point.

PR #1788, opened 2021-12-15
Allocate Wasm memory lazily

tomaka:
This PR fixes the current mess with `heap_pages`.
I now have a better understanding of what heap pages are for.

It also adds a new feature: the memory of the heap is not completely allocated upfront, but lazily when the allocator allocates from it.
In other words, if your heap pages value is for example 10, we start with 0 (and check that 10 would fit) and dynamically increase the actual allocated memory. 10 is only the limit given to the allocator after which it returns an error, but doesn't correspond to an actual amount of allocated memory anymore.


tomaka:
Polkadot block 993 fails with an OOM error.

PR #1791, opened 2021-12-15
Enforce no-std compatibility in CI

### LINKED ISSUE #106: Make the library compile for no_std
Everything in the title.

For obvious reasons, only `wasmi` would be supported.

tomaka:
Close https://github.com/paritytech/smoldot/issues/106

This is not passing, but this PR can be used as a way to track progress.


Cargo.toml
```diff
@@ -94,7 +94,7 @@ soketto = { version = "0.7.1", optional = true }
 # TODO:
 parity-scale-codec = { version = "2.3.1", features = ["derive"] } # TODO: a lot of unnecessary overhead in terms of memory allocations
 
-[target.'cfg(target_arch = "x86_64")'.dependencies]
+[target.'cfg(target_arch = "x86_63")'.dependencies]
```
tomaka, line 97: Can't figure out why `wasmtime` is compiled even though we pass `--no-default-features`.
Should be figured out and removed.

gilescope:
Is it possible to compile smoldot for wasm32-unknown-unknown? Would be great to get that enforced in the CI too. I wanted to try and use it from a wgpu rendered bevy client on the browser but when I pull in the dep:
```
smoldot = { git= "https://github.com/paritytech/smoldot.git", default-features = false }
```
it grumbles about sql light which is odd as I told it no default features...

tomaka:
`wasm32-unknown-unknown` has a `std`, so this is off topic.
But yes, the `smoldot` and `smoldot-light-base` libraries compile for Wasm, that is checked by CI.

Issue #189, opened 2020-11-13
Support nested calls in virtual machine


tomaka:
In the Substrate sandboxing API, the runtime can call a function in a separate VM (the guest), and this guest can in return call functions in its supervisor. These "callbacks" happen while the `invoke` function (that invokes the guest) is still in progress.

In order to be able to implement this, the virtual machine API needs to support being able to call a function while an existing one is in an interrupted state.

tomaka:
No longer needed after https://github.com/paritytech/substrate/pull/9592.
We should remove all the corresponding host functions.

tomaka:
After discussion, it is planned to reintroduce usage of these functions of Substrate, so let's not remove them.

PR #1838, opened 2022-01-04
Remove all sandboxing-related host functions

tomaka:
Close https://github.com/paritytech/smoldot/issues/189


tomaka:
CI fails because it breaks compatibility with older runtimes.

We have two possibilities:

- Leave the functions as they are.
- Allow the runtime to have unresolved imports as long as the unresolved function isn't called.

Need to ask experts what is correct.

tomaka:
Closing, see https://github.com/paritytech/smoldot/issues/189#issuecomment-1039124703

PR #1842, opened 2022-01-05
Transition JavaScript code to TypeScript

tomaka:


bin/wasm-node/javascript/src/compat/index.d.ts
```diff
@@ -0,0 +1,45 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+// TODO: remove this reference thing /!\ it's completely wrong
+/// <reference lib="dom" />
```
tomaka, line 19: Going to leave this hack in, for the moment. It's not possible to use WebAssembly without the `dom` library at the moment.

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+                mem.writeUInt32LE(ptr, error_ptr_ptr);
+                mem.writeUInt32LE(len, error_ptr_ptr + 4);
+                mem.writeUInt8(isBadAddress ? 1 : 0, error_ptr_ptr + 8);
+                return 1;
+            }
+        },
+
+        // Must close and destroy the connection object.
+        connection_close: (id: number) => {
+            let connection = connections[id] as TcpWrapped | WebSocketWrapped;
```
josepot, line 258: the `as TcpWrapped | WebSocketWrapped` is not needed (and a bit misleading b/c someone could think that what's in `connections` is not that type), so I would remove it. If you wanted to be explicit about the type of `connection` you could do:

```
let connection: TcpWrapped | WebSocketWrapped = connections[id]
```

which is unnecessary, but at least is not misleading :slightly_smiling_face:

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+export default function (config: Config): compat.WasmModuleImports {
+    // Used below to store the list of all connections.
+    // The indices within this array are chosen by the Rust code.
+    let connections: Record<number, TcpWrapped | WebSocketWrapped> = {};
+
+    return {
+        // Must exit with an error. A human-readable message can be found in the WebAssembly
+        // memory in the given buffer.
+        panic: (ptr: number, len: number) => {
+            const instance = config.instance as SmoldotWasmInstance;
```
josepot, line 50: Perhaps a better way of doing this would be:

```ts
const intance = config.instance!
```

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+
+import { Buffer } from 'buffer';
+import Websocket from 'websocket';
+import * as compat from '../compat/index.js';
+import type { SmoldotWasmInstance } from './bindings.js';
+
+import type { Socket as TcpSocket } from 'net';
+
+export interface Config {
+    instance?: SmoldotWasmInstance,
```
josepot, line 31: why is it optional?

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
josepot: or just
```ts
const { instance } = config
```
if the Config interface defines the instance as a non-optional field.

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+
+                } else if (tcpParsed != null) {
+                    // `net` module will be missing when we're not in NodeJS.
+                    if (!compat.isTcpAvailable() || config.forbidTcp) {
+                        throw new ConnectionError('TCP connections not available');
+                    }
+
+                    const socket = compat.createConnection({
+                        host: tcpParsed[2],
+                        port: parseInt(tcpParsed[3] as string, 10),
```
josepot, line 204:
```suggestion
                        port: parseInt(tcpParsed[3]!, 10),
```

bin/wasm-node/javascript/src/worker/worker.ts
```diff
@@ -57,9 +55,9 @@ const injectMessage = (instance, message) => {
     // Write the potential relay chains into memory.
     const potentialRelayChainsLen = message.potentialRelayChains.length;
     const potentialRelayChainsPtr = instance.exports.alloc(potentialRelayChainsLen * 4) >>> 0;
-    for (let idx in message.potentialRelayChains) {
+    for (let idx = 0; idx < message.potentialRelayChains.length; ++idx) {
       Buffer.from(instance.exports.memory.buffer)
-        .writeUInt32LE(message.potentialRelayChains[idx], potentialRelayChainsPtr + idx * 4);
+        .writeUInt32LE(message.potentialRelayChains[idx] as number, potentialRelayChainsPtr + idx * 4);
```
josepot, line 60: Using `forEach` i s a bit more declarative and then there is no need to deal with possibly undefined values:

```ts
message.potentialRelayChains.forEach((potentialRelayChain, idx) => {       
  Buffer.from(instance.exports.memory.buffer)                              
    .writeUInt32LE(potentialRelayChain, potentialRelayChainsPtr + idx * 4);
})                                                                         
```

Alternatively: `message.pontentialRelayChains[idx]!` is less verbose and clearer IMO.

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
josepot: Ups! I was wrong, when I wrote that ^^ I didn't realize that you were using the `noUncheckedIndexedAccess` option, sorry about that. Well, then I would go with: `connections[id]!`.

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+                    });
+
+                } else {
+                    throw new ConnectionError('Unrecognized multiaddr format');
+                }
+
+                connections[id] = connection;
+                return 0;
+
+            } catch (error) {
```
wirednkod, line 239:
```suggestion
            } catch (error: Error | ConnectionError) {
```

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+                // Note: peers can decide of the content of `addr`, meaning that it shouldn't be
+                // trusted.
+                const wsParsed = addr.match(/^\/(ip4|ip6|dns4|dns6|dns)\/(.*?)\/tcp\/(.*?)\/(ws|wss)$/);
+                const tcpParsed = addr.match(/^\/(ip4|ip6|dns4|dns6|dns)\/(.*?)\/tcp\/(.*?)$/);
+
+                if (wsParsed != null) {
+                    let proto = 'wss';
+                    if (wsParsed[4] == 'ws') {
+                        proto = 'ws';
+                    }
```
wirednkod, line 157:
```suggestion
                    const proto = (wsParsed[4] == 'ws') ? 'ws' : 'wss';
```

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
```diff
@@ -0,0 +1,310 @@
...
+                    ) {
+                        throw new ConnectionError('Connection type not allowed');
+                    }
+
+                    let url: string;
+                    if (wsParsed[1] == 'ip6') {
+                        url = proto + "://[" + wsParsed[2] + "]:" + wsParsed[3];
+                    } else {
+                        url = proto + "://" + wsParsed[2] + ":" + wsParsed[3];
+                    }
```
wirednkod, line 171:
```suggestion
                    const url = (wsParsed[1] == 'ip6') ?
                        url = proto + "://[" + wsParsed[2] + "]:" + wsParsed[3] :
                        url = proto + "://" + wsParsed[2] + ":" + wsParsed[3];
```

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
tomaka: We need to pass the bindings as parameter to `WebAssembly.instantiate`, so there's absolutely no way to pass an `instance` here. The way the code works is we create the bindings, then add the `instance` to the config after it is created.

bin/wasm-node/javascript/src/worker/bindings-smoldot-light.ts
tomaka: Doesn't compile. Catch clause can only be `any` or `unknown`

tomaka:
Addressed review, thanks!

PR #1845, opened 2022-01-06
Fix bogus variable name in environ_get

tomaka:


bin/wasm-node/javascript/src/bindings-wasi.js
```diff
@@ -132,27 +132,27 @@ export default (config) => {
...
-            argv_buf >>>= 0;
+            argvBuf >>>= 0;
 
             const mem = Buffer.from(config.instance.exports.memory.buffer);
 
             let argvPos = 0;
             let argvBufPos = 0;
 
-            envVars.forEach(env_var => {
-                let envVarLen = Buffer.byteLength(e, 'utf8');
```
tomaka, line 148: `e` doesn't exist here

PR #1846, opened 2022-01-06
Implement blocks pinning in runtime service

tomaka:
In order to implement the new JSON-RPC API, and in order to make the runtime service more deterministic, the runtime service now has a pinned blocks system as well.

Every block that is reported through its public API is automatically "pinned" and needs to be unpinned by the receiver.
While a block is pinned, the runtime service is guaranteed to keep in its state the runtime of that block.

This solves many ambiguities regarding the current API of the runtime service.


bin/light-base/src/runtime_service.rs
```diff
@@ -41,6 +41,19 @@
...
+//!
+//! Blocks that are reported through [`RuntimeService::subscribe_all`] are automatically *pinned*.
+//! If multiple subscriptions exist, each block is pinned once per subscription.
+//!
+//! As long as a block is pinned, the [`RuntimeService`] is guaranteed to keep in its internal
+//! state the runtime of this block and its properties.
+//!
+//! Blocks must be manually unpinned by calling TODO.
+//! Failing to do so is effectively a memory leak. If the number of pinned blocks becomes too
+//! large, the subscription is force-killed by the [`RuntimeService`].
```
tomaka, line 55: Remains to do

bin/light-base/src/json_rpc_service.rs
```diff
@@ -2476,6 +2494,7 @@ impl<TPlat: Platform> Background<TPlat> {
         };
 
         let mut subscribe_all = if runtime_updates {
+            // TODO: must unpin blocks
```
tomaka, line 2497: Remains to do

PR #1847, opened 2022-01-07
Report pruned blocks in runtime service notifications

tomaka:
So far my attitude has been "the notifications subscriber can just maintain their own tree of blocks", but it turns out that this is really complicated and annoying to do. Instead, the pruned blocks are also sent by the background task.

### DIFF
--- a/bin/light-base/src/runtime_service.rs
+++ b/bin/light-base/src/runtime_service.rs
@@ -284,6 +284,7 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
                         Some(Notification::Finalized {
                             hash,
                             best_block_hash,
+                            ..
                         }) => {
                             finalized_hash = hash;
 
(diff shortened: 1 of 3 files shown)


bin/light-base/src/runtime_service.rs
```diff
@@ -836,18 +837,21 @@ pub struct SubscribeAll {
...
+        /// Blake2 hash of the header of the block that has been finalized.
         ///
         /// A block with this hash is guaranteed to have earlier been reported in a
         /// [`BlockNotification`], either in [`SubscribeAll::non_finalized_blocks_ancestry_order`]
         /// or in a [`Notification::Block`].
         ///
-        /// It is, however, not guaranteed that this block is a child of the previously-finalized
-        /// block. In other words, if multiple blocks are finalized at the same time, only one
+        /// It is also guaranteed that this block is a child of the previously-finalized block. In
+        /// other words, if multiple blocks are finalized at the same time, only one
```
tomaka, line 847: This was actually already the case. The difference with this PR is that it is now a guarantee and not an implementation detail.

PR #1848, opened 2022-01-07
Implement finalized subscriptions upon subscribe_all

tomaka:


tomaka:
Realized that the stream has to be infinite.

PR #1873, opened 2022-01-12
Merge the two chain maps together

tomaka:


tomaka:
It's unclear to me why CI fails. Everything works locally.

tomaka:
CI failure was due to Node version being low. Bumped to Node v14, which was released in 2020.

PR #1976, opened 2022-02-01
Change the way bootnodes are passed to networking

tomaka:
There's a small problem before this PR: a user can add a chain with an empty list of bootnodes, then add another chain with a non-empty list of bootnodes. Because the networking service is created only the first time, the bootnodes of that other chain will never be added, and the networking of the second chain will not have any peer.

### DIFF
--- a/src/network/service.rs
+++ b/src/network/service.rs
@@ -909,6 +909,9 @@ where
                 debug_assert!(!kbuckets_addrs.get_mut().is_empty());
             }
         }
+
+        // As we have new potential peers, ask the user to try open new connections.
+        self.start_connect_needed.notify_additional(1);
     }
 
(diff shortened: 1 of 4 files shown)


tomaka:
I think the code in this PR is correct, but a problem in the networking makes connecting to new peers quite slow.

src/network/service.rs
```diff
@@ -909,6 +909,9 @@ where
                 debug_assert!(!kbuckets_addrs.get_mut().is_empty());
             }
         }
+
+        // As we have new potential peers, ask the user to try open new connections.
+        self.start_connect_needed.notify_additional(1);
```
tomaka, line 914: This is a bugfix

tomaka:
Will merge after https://github.com/paritytech/smoldot/pull/1980, to maximize the chances of detecting a bug after this PR is merged and before it's released.

PR #1990, opened 2022-02-03
Make start_slot_number mandatory for past epochs

tomaka:
It turns out there is some fiasco regarding `start_slot_number`.

Currently it is an `Option` and the documentation mentions that it must be none iff `epoch_index` is 0.
This is however only true for future epochs. If the current epoch is epoch 0, then its starting slot can be known and must be known.
This PR modifies the check for the validity of chain information to allow passing a slot number for epoch 0 if it is the current epoch. This fixes a problem during warp syncing to a block within epoch 0.

Since the `blocks_tree` module would always leave `start_slot_number` equal to None for epoch 0 when syncing from scratch, it would now lead to an invalid `ChainInformation` when serializing the state when within epoch 0. This has been updated too.

Additionally, when fetching the Babe epoch information at the end of warp syncing, the start slot number is now ignored if we fetch the future epoch and obtain epoch index 0 (in which case the runtime returns an unspecified value).

### DIFF
--- a/src/chain/blocks_tree/verify.rs
+++ b/src/chain/blocks_tree/verify.rs
@@ -372,8 +372,29 @@ impl<T> VerifyContext<T> {
                 VerifyConsensusSpecific::Babe { .. },
                 FinalizedConsensus::Babe { .. },
                 Some(BlockConsensus::Babe { next_epoch, .. }),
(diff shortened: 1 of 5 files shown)


tomaka:
Tested this PR with both syncing from scratch and warp syncing.

PR #2005, opened 2022-02-08
Implement wake_by_ref instead of wake in wasm-node executor

tomaka:


tomaka:
This PR is wrong.
See https://github.com/paritytech/smoldot/pull/2006

PR #2012, opened 2022-02-08
Refactor JSON-RPC service initialization

### LINKED ISSUE #1935: JSON-RPC server non responsive if connected to early on
How to reproduce:

- Open two PolkadotJS tabs connected to the demo.
- Restart the demo.

One of the two tabs will never receive any response from the node:

image

### LINKED ISSUE #1961: `wasm-node-checks` CI intermittent failures
Example: https://github.com/paritytech/smoldot/runs/4983825371?check_suite_focus=true

Should be figured out.

Can't reproduce locally.

tomaka:
Refactor the JSON-RPC service initialization to fix the problem outlined [here](https://github.com/paritytech/smoldot/issues/1961#issuecomment-1032988876).

Fix #1961
Fix #1935

More precisely, this PR splits the JSON-RPC service initialization in two: first we create some kind of channel-like, then later we finish initialization. Thanks to this, the JSON-RPC service itself supports sending requests immediately even before it is fully initialized. Right now it's `lib.rs` that does this, and does it poorly.

### DIFF
--- a/bin/light-base/src/json_rpc_service.rs
+++ b/bin/light-base/src/json_rpc_service.rs
@@ -57,21 +57,171 @@ use smoldot::{
 use std::{
     collections::HashMap,
     iter,
-    marker::PhantomData,
     num::NonZeroU32,
     str,
     sync::{atomic, Arc},
     time::Duration,
 };
 
-/// Configuration for a JSON-RPC service.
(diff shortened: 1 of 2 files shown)


tomaka:
Can confirm that this fixes the test failures in https://github.com/paritytech/smoldot/pull/2001

PR #2026, opened 2022-02-10
Implement generating checkpoints in smoldot

tomaka:


tomaka:
Closing. We need a proper checkpoint format before implement that.
cc https://github.com/paritytech/substrate/issues/11184

Issue #1305, opened 2021-09-15
Storage changes overlay is rather slow


tomaka:
After https://github.com/paritytech/smoldot/pull/1304, the data structures representing a storage diff are `BTreeMap`s, which is quite slower compared to a `HashMap`.

A solution should be found, and more generally we should create a separate data structure dedicated to storage diffs.

PR #2051, opened 2022-02-16
Optimize storage diff by having both a hashmap and btree

tomaka:
Close https://github.com/paritytech/smoldot/issues/1305

Also renames `diff_iter()` to `diff_iter_unordered()` because the ordering is now broken. This clarifies that no ordering is guaranteed.

Opening as draft because it seems that this might actually slow down syncing. Need to do some precise measurements.

### DIFF
--- a/bin/full-node/src/run/consensus_service.rs
+++ b/bin/full-node/src/run/consensus_service.rs
@@ -1077,7 +1077,7 @@ impl SyncBackground {
                                     .as_ref()
                                     .unwrap()
                                     .storage_top_trie_changes
-                                    .diff_iter()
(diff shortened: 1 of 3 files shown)


tomaka:
I believe that in my initial testing I had some download speed issue, because after more testing it looks something like 2-3 times faster than without this PR.

This is very much a guess-timate, but it is the result I would expect for not having to traverse a binary tree at every lookup.

tomaka:
My machine is syncing at ~500 blocks/sec after this PR. Before this PR the verification time was very erratic, ranging from ~50 blocks per second to ~200 blocks per second.

PR #2054, opened 2022-02-17
Don't notify transaction inclusions after download

tomaka:
As this is done when we call `set_best_block`.

Notifying right after the download is a mistake.


tomaka:
This is wrong. The code is actually correct.

PR #2057, opened 2022-02-17
Use cached cargo-install

TriplEight:
I assumed @tomaka  didn't like [`cargo-nextest`](https://nexte.st/index.html) so not suggesting it here.
Here you go, 1sec instead of 2 min
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,6 +70,11 @@ jobs:
       with:
         node-version: '14'
     - run: cargo install twiggy  # TODO: somehow cache this in order to not compile twiggy every single time
+    - name: Install twiggy with cache
+        uses: baptiste0928/cargo-install@v1
+        with:
+          crate: cargo-nextest
+          version: 0.9
     - run: git checkout ${{ github.event.pull_request.base.sha }}
     - run: cd bin/wasm-node/javascript && npm ci && npm run-script build
     - run: cp ./target/wasm32-wasi/min-size-release/smoldot_light_wasm.wasm ./.ci-parent-build.wasm  # TODO: maybe get the path from the `npm build` output or something?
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -69,7 +69,10 @@ jobs:
     - uses: actions/setup-node@v2.5.1
       with:
         node-version: '14'
-    - run: cargo install twiggy  # TODO: somehow cache this in order to not compile twiggy every single time
+    - uses: baptiste0928/cargo-install@v1
```
tomaka, line 72:
```suggestion
    - uses: baptiste0928/cargo-install@v1  # This action ensures that the twiggy compilation is cached.
```

tomaka (approved):
Thanks

tomaka:
> I assumed @tomaka didn't like [cargo-nextest](https://nexte.st/index.html) so not suggesting it here.

That's not true. I'm cautious against being overly reliant on tools like this, not avoiding using them. For example parsing the programmatic output of `cargo nextest` wouldn't be a good idea.

TriplEight:
just nagging you there :)

> For example parsing

[yes](https://github.com/nextest-rs/nextest/issues/41), [indeed](https://github.com/nextest-rs/nextest/issues/20), but do you parse it?

tomaka:
What I mean is that if `cargo nextest` happened to be abandoned over night, we should be able to switch back to `cargo test` easily. If we were analyzing the output of `cargo nextest` and performing important actions depending on the result, we would lose the possibility to easily switch back to `cargo test`.

TriplEight:
yeah, this is a normal concern, but considering how rarely we end up parsing i.e. test outputs I think it's worth discussing this probability when we're about to parse some outputs.

PR #2072, opened 2022-02-19
Remove execute_block module as too thin

tomaka:
This PR merges the `execute_block` module into `header_body`.

The `execute_block` module basically did nothing except provide the correct function name and parameters, and check whether the output is empty.


tomaka:
Tested to be working

Issue #227, opened 2020-11-20
Figure out check_inherents


tomaka:
It seems that we need to also call `check_inherents` in order to verify blocks.

tomaka:
Apparently this only needs to be done on validator nodes.

PR #2073, opened 2022-02-19
Implement calling check_inherents

tomaka:
Close https://github.com/paritytech/smoldot/issues/227

~~Opening as draft because not fully implemented.~~

### DIFF
--- a/src/verify/header_body.rs
+++ b/src/verify/header_body.rs
@@ -299,7 +299,12 @@ pub fn verify(
 
     // Start the virtual machine with `BlockBuilder_check_inherents`.
     let check_inherents_process = {
-        // The second parameter of `BlockBuilder_check_inherents` is a list of inherents. 
+        // The second parameter of `BlockBuilder_check_inherents` contains the current timestamp
+        // and current Aura or Babe slot number. The Aura/Babe slot number is no longer used in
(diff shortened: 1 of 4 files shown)


src/verify/header_body.rs
```diff
@@ -274,25 +285,85 @@ pub fn verify(
...
+                // TODO: consider using a function to get current slot from aura/babe module
+                ConfigConsensus::Aura { slot_duration, .. } => {
+                    inherents::InherentDataConsensus::Aura {
+                        slot_number: config.now_from_unix_epoch.as_secs() * 1000
+                            / slot_duration.get(),
+                    }
+                }
+                ConfigConsensus::Babe { .. } => {
+                    inherents::InherentDataConsensus::Babe {
+                        slot_number: u64::max_value(), // TODO: hack /!\ but we don't really care as we're planning to remove this with https://github.com/paritytech/smoldot/issues/2096
```
tomaka, line 339: I wanted to be lazy and implement this is hack, because we're going to remove this code with https://github.com/paritytech/smoldot/issues/2096 anyway
Unfortunately, this hack causes the runtime of older blocks to return errors.
I think I'm going to do https://github.com/paritytech/smoldot/issues/2096 first, and then rebase this PR on top.

Issue #1623, opened 2021-11-12
Treat clippy warnings as errors


tomaka:
First, all the clippy warnings need to be fixed, which is a lot of work.

PR #2098, opened 2022-02-24
run `cargo clippy --fix`

melekes:
Refs #1623

### DIFF
--- a/src/identity/seed_phrase.rs
+++ b/src/identity/seed_phrase.rs
@@ -215,7 +215,7 @@ pub fn bip39_to_seed(phrase: &str, password: &str) -> Result<[u8; 32], Bip39ToSe
 
     // These rules are part of the seed phrase format "specification" and have been copy-pasted
     // from the Substrate code base.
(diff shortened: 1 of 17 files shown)


src/identity/seed_phrase.rs
```diff
@@ -215,7 +215,7 @@ pub fn bip39_to_seed(phrase: &str, password: &str) -> Result<[u8; 32], Bip39ToSe
 
     // These rules are part of the seed phrase format "specification" and have been copy-pasted
     // from the Substrate code base.
-    if entropy_len < 16 || entropy_len > 32 || entropy_len % 4 != 0 {
+    if !(16..=32).contains(&entropy_len) || entropy_len % 4 != 0 {
```
tomaka, line 218: Not a fan of this change in isolation, but since it fulfills the goal of making clippy happy it's fine

tomaka (approved):

src/identity/seed_phrase.rs
melekes: yeah, don't like this too

Issue #2096, opened 2022-02-24
Backport `aura, babe: remove slot inherent data`


tomaka:
https://github.com/paritytech/substrate/pull/10906

~~The slot number for inherent data can be removed from the block authorship code, but should stay for the blocks verification code for backwards compatibility reasons.~~

The slot number can be removed altogether, and we ignore errors when verifying older blocks.

PR #2099, opened 2022-02-24
Remove from InherentData the Aura/Babe slot number

tomaka:
Close https://github.com/paritytech/smoldot/issues/2096

What the comment says: we no longer pass the Aura/Babe slot number to the runtime when authoring blocks, because they're unused anyway. Authoring blocks using older runtime versions is intentionally not supported.

Verifying the inherents of blocks isn't implemented in smoldot yet, and will be covered by https://github.com/paritytech/smoldot/pull/2073


melekes (approved):
👍

PR #2101, opened 2022-02-24
Use new std::thread::available_parallelism function

tomaka:
New function added in Rust 1.59.

There's no policy about minimum Rust version in smoldot at the moment. We don't have to merge this now but can wait a bit if necessary.


melekes (approved):
👍

tomaka:
https://github.com/paritytech/smoldot/pull/2132

Issue #2088, opened 2022-02-22
Parachain (Westmint) connectivity (without non-ws bootnodes) results to panicked smoldot


wirednkod:
While trying to connect to Westmint parachain with substrate-connect, I realized that 0 peers were returned.
After @tomaka 's suggestion the non-ws bootnodes were removed and connectivity succeeded - but with smoldot panicking.

Attached can be found the [smoldot's logs](https://github.com/paritytech/smoldot/files/8117051/panicked-smoldot.log) when panic occurred, and the [browser console's logs](https://github.com/paritytech/smoldot/files/8117050/panicked-smoldot-console-errors.log) of the error.

tomaka:
Problem most likely caused by the fact that we seem to report the same block three times:

> client.js:77 [runtime-westmint] Worker => OutputNewBlock(hash=0xd790…81b2, is_new_best=true)
> client.js:77 [runtime-westmint] Worker => OutputNewBlock(hash=0xd790…81b2, is_new_best=false)
> client.js:77 [runtime-westmint] Worker => OutputNewBlock(hash=0xd790…81b2, is_new_best=false)

tomaka:
What I don't understand is that it's supposed to be handled by this:
https://github.com/paritytech/smoldot/blob/2a8ff45a2c9704d8a2f4c97af64714b06b977f12/bin/light-base/src/sync_service/parachain.rs#L259-L261

tomaka:
I believe the problem is that the three relay blocks aren't children of each other but forks (unfortunately the logs don't mention this).
All three relay chain forks include the same parahead that is different from the parahead in their relay parent, leading us to report the same parahead three times.

tomaka:
Only proper way to fix this is to add a `ForkTree` for the parachain blocks, to track what we have reported or not.

PR #2105, opened 2022-02-25
Log the block's parent in block announces

tomaka:
When investigating https://github.com/paritytech/smoldot/issues/2088 I realized that we were missing this critical info.


melekes (approved):
👍

PR #2119, opened 2022-03-04
Update adz parachain spec

tomaka:
The "adz" parachain has migrated from Westend to Rococo and has a brand new chain specification.


tomaka:
There's a genesis mismatch. The chain spec isn't correct.

tomaka:
Now working

PR #2120, opened 2022-03-04
implement ecdsa sign, verify, sign_prehashed and verify_prehashed

### LINKED ISSUE #1968: Backport "expose ecdsa_sign_prehashed in sp-io"
https://github.com/paritytech/substrate/pull/10119

melekes:
Closes #1968

Note this PR implements `ecdsa_sign` and introduces `ecdsa_verify` function, which are both out of the scope of the issue above.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,6 @@ hashbrown = { version = "0.12.0", default-features = false, features = ["serde"]
 hex = { version = "0.4.3", default-features = false }
 hmac = { version = "0.12.1", default-features = false }
 itertools = { version = "0.10.3", default-features = false }
-# TODO: don't activate `static-context` by default; see https://github.com/paritytech/libsecp256k1/issues/77
 libsecp256k1 = { version = "0.7.0", default-features = false, features = ["static-context", "hmac"] }
 # Reminder: `log` is forbidden
 merlin = { version = "3.0", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -58,7 +58,7 @@ hex = { version = "0.4.3", default-features = false }
 hmac = { version = "0.12.1", default-features = false }
 itertools = { version = "0.10.3", default-features = false }
 # TODO: don't activate `static-context` by default; see https://github.com/paritytech/libsecp256k1/issues/77
-libsecp256k1 = { version = "0.7.0", default-features = false, features = ["static-context"] }
+libsecp256k1 = { version = "0.7.0", default-features = false, features = ["static-context", "hmac"] }
 # Reminder: `log` is forbidden
 merlin = { version = "3.0", default-features = false }
 nom = { version = "7.1.0", default-features = false, features = ["alloc"] }
--- a/src/executor/host.rs
+++ b/src/executor/host.rs
@@ -1114,32 +1114,50 @@ impl ReadyToRun {
             }
             HostFunction::ext_crypto_ecdsa_generate_version_1 => host_fn_not_implemented!(),
             HostFunction::ext_crypto_ecdsa_sign_version_1 => {
-                let data = [0; 32];
+                let mut data = [0; 32];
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -58,7 +58,7 @@ hex = { version = "0.4.3", default-features = false }
 hmac = { version = "0.12.1", default-features = false }
 itertools = { version = "0.10.3", default-features = false }
 # TODO: don't activate `static-context` by default; see https://github.com/paritytech/libsecp256k1/issues/77
```
melekes, line 60: @tomaka can we remove this comment since a) the issue is closed b) the reason we're not enabling default features is no-std, no?

src/executor/host.rs
```diff
@@ -1113,8 +1113,64 @@ impl ReadyToRun {
                 })
             }
             HostFunction::ext_crypto_ecdsa_generate_version_1 => host_fn_not_implemented!(),
-            HostFunction::ext_crypto_ecdsa_sign_version_1 => host_fn_not_implemented!(),
+            HostFunction::ext_crypto_ecdsa_sign_version_1 => {
+                let data = [0; 32];
+                data.copy_from_slice(
```
melekes, line 1118: is there a way to avoid copying here? or it will be unsafe? 🤔

src/executor/host.rs
```diff
@@ -1113,8 +1113,64 @@ impl ReadyToRun {
...
+                    blake2_rfc::blake2b::blake2b(32, &[], expect_pointer_size!(0).as_ref())
+                        .as_bytes(),
+                );
+                let message = libsecp256k1::Message::parse(&data);
+                let (sig, ri) =
+                    libsecp256k1::sign(&message, &expect_pointer_constant_size!(1, 32).into());
+
+                // NOTE: the function returns 2 slices: signature and recovery ID (AS A SLICE)
```
melekes, line 1126: should we return a single slice instead? signature (64 bytes) + 1 recovery byte

melekes:
@tomaka can I also implement ecdsa verity (should be straightforward now that we have `verify_prehashed`)?

src/executor/host.rs
tomaka: Yeah, you can call `as_bytes` on what `blake2b` returns: https://docs.rs/blake2-rfc/latest/blake2_rfc/blake2b/struct.Blake2bResult.html

tomaka:
> can I also implement ecdsa verity (should be straightforward now that we have verify_prehashed)?

Yeah go ahead

Cargo.toml
tomaka: Yes

tomaka: I don't think a new version has been released though

src/executor/host.rs
melekes: `as_bytes` returns `&[u8]`, I need `[u8; 32]`

tomaka (approved):
Looks good, modulo avoiding the copy!

src/executor/host.rs
tomaka: Ah I see. You can do `<[u8; 32]>::try_from(....as_bytes()).unwrap()`

tomaka:
Note that I haven't properly checked if the behavior is the same as in Substrate, but it _seems_ good.

Unfortunately none of the crypto functions in `host.rs` are tested because writing unit tests for that is a big effort.

I know that a team from w3f is currently writing a test framework where they give runtimes to a client and check if said client behaves properly, but it's not ready yet.

melekes:
> Unfortunately none of the crypto functions in host.rs are tested because writing unit tests for that is a huge effort.

yeah, I figured

> I know that a team from w3f is currently writing a test framework where they give runtimes to a client and check if said client behaves properly, but it's not ready yet.

this sounds like exactly what is needed!

PR #2121, opened 2022-03-06
Make set_parent_hash mark entire chain as bad

tomaka:
If a certain block is bad, all its descendants are meant to be marked as bad as well.
When you call `set_parent_hash`, and the parent is bad, the immediate child is set as bad. This PR also sets the entire chain of descendants as bad.


melekes (approved):
👍

tomaka:
Wait for https://github.com/paritytech/smoldot/pull/2122 to merge

Issue #1572, opened 2021-11-04
AllSync block user data confusing


tomaka:
The block user data of the `AllSync` isn't in a usable state.

PR #2123, opened 2022-03-07
Refactor AllForksSync::finish_ancestry_search

tomaka:
This paves the way towards https://github.com/paritytech/smoldot/issues/1572

Before this PR, you would call `finish_ancestry_search` and pass an iterator to the list of blocks contained in the response.
The problem is that some of these blocks might already be known while not others, making it impossible for the user to easily update block user data.

This PR refactors this function: it now takes ownership of the state machine, and lets you add blocks one by one.

This new API is also more explicit, as the user knows at which point the response has an issue.


tomaka:
I'm intentionally using a `loop` so that I can do `break value` and use it as a `goto`, but clippy doesn't seem to like that :-/

src/sync/all_forks.rs
```diff
@@ -976,13 +883,153 @@ impl<TBl, TRq, TSrc> ops::IndexMut<SourceId> for AllForksSync<TBl, TRq, TSrc> {
...
+    /// On success, the block is inserted in the state machine.
+    ///
+    /// If an error is returned, the [`FinishAncestrySearch`] is turned back again into a
+    /// [`AllForksSync`], but all the blocks that have already been added are retained.
+    pub fn add_block(
+        mut self,
+        scale_encoded_header: impl AsRef<[u8]>,
+        scale_encoded_justifications: impl Iterator<Item = ([u8; 4], impl AsRef<[u8]>)>,
+    ) -> Result<Self, (AncestrySearchResponseError, AllForksSync<TBl, TRq, TSrc>)> {
+        let scale_encoded_header = scale_encoded_header.as_ref();
```
melekes, line 922: isn't `scale_encoded_header` already a ref?

melekes (approved):
👍

src/sync/all_forks.rs
tomaka: `impl AsRef<[u8]>` allows you pass things such as `Vec<u8>` and such. It doesn't strictly have to be a `&[u8]`.
I've originally used `impl AsRef<[u8]>` because when we're talking about the items yielded by an iterator it's sometimes the only solution. For example if you have an iterator that generates `Vec<u8>`s, you can't turn it into an iterator that generates `&[u8]`s without some additional buffering.

However, now that the iterator is gone, I'm going to turn this into a `&[u8]`.

PR #2129, opened 2022-03-09
Refactor all forks a bit

tomaka:
Continuation of https://github.com/paritytech/smoldot/pull/2123 and https://github.com/paritytech/smoldot/issues/1572

Instead of having `FinishAncestrySearch::add_block` immediately insert the block, we now instead return a new `AddBlock` enum that indicates whether the block has already been inserted before, and that lets you actually do the block insertion.

The diff is moderately large, but this PR should be easy to review commit by commit.

It still doesn't finish this whole "user data saga". The next steps will be to update `block_announce` and `add_source`, as they also manipulate blocks. I didn't want to bloat this single PR too much.


melekes (approved):
lgtm

PR #1938, opened 2022-01-21
Experiment with a WebRTC client

tomaka:
cc https://github.com/paritytech/smoldot/issues/1712

Opening this PR as a draft, mostly to keep the overview that I wrote down.

The code in this PR makes the browser open a TCP connection to 127.0.0.1:41000.
I haven't experimented with a DTLS+SCTP server yet.

Issue #1712, opened 2021-11-30
wasm-node: Add support for WebRTC


tomaka:
See https://github.com/libp2p/specs/ for how to implement this.

Two things need to be done:

- [x] Modify the networking code to support transport protocols that natively support substreams.
- [x] Modify `bindings-smoldot-light.js` to support WebRTC.

tomaka:
As a small note, WebRTC is unfortunately not available from within workers (despite WebSockets being available, don't ask me), so we would have to do some messages exchanges between outside and inside the worker to make that work.

tomaka:
Considering that we might need to add another encryption layer within the DTLS+SCTP stream, it makes sense to me to simply open one SCTP channel for everything and apply Noise+Yamux to it.

As such, I've crossed out the first point.

> Modify the networking code to support transport protocols that natively support substreams.

Maybe it turns out that in the end we'll need to do it, but for now it's not necessary.

tomaka:
WebRTC should solve a problem related to blacklisted IP addresses, which I won't go into details to avoid someone "exploiting" this (though it's not really an exploit, just a way to be annoying).

melekes:
What's left from the transport perspective:

- [ ] submit a spec for WebRTC libp2p transport
- [ ] finish https://github.com/libp2p/rust-libp2p/pull/2622
- [ ] use the new transport in smoldot

tomaka:
After https://github.com/paritytech/smoldot/pull/2352:

- [ ] Add support for a `/webrtc` (or similar) protocol in `Multiaddr`.
- [ ] Actually add support for WebRTC in the smoldot JS code.
- [ ] Finish the areas where https://github.com/paritytech/smoldot/pull/2352 has left todos, mainly `wasm-node/rust/src/platform`. This will be more easily done once the JS supports WebRTC, so that we can actually test if things work.

tomaka:
After https://github.com/paritytech/smoldot/pull/2579, what remains to do is:

- [ ] https://github.com/paritytech/smoldot/pull/2759
- [ ] Implement the meta-headers in each message.
- [ ] Stabilize the whole thing by removing the `enableExperimentalWebRTC` flag.

tomaka:
Another thing to do is fix compatibility with Firefox.
[...]

PR #2135, opened 2022-03-10
WebRTC client & server

melekes:
Previous work by @tomaka: https://github.com/paritytech/smoldot/pull/1938
Refs #1712 

**TODO**:
- [x] read SDP spec https://datatracker.ietf.org/doc/html/rfc3264
- [x] modify SDP in webrtc server (and check one in client)
- [x] working connection
- [x] make necessary changes to `webrtc` rs library
- [x] sometimes DTLS handshake does not occur => debug & figure out why. maybe this is connected to `tls-id` attribute? no. it's because of the issue with local candidates (see https://github.com/melekes/libp2p-webrtc-direct/blob/ed829dc035b7cff1c155bfbd402261fbadf86d72/src/transport.rs#L531-L540)
- [x] implement https://docs.rs/libp2p/latest/libp2p/core/transport/trait.Transport.html
- [ ] submit a spec for WebRTC libp2p transport
- [ ] finish https://github.com/libp2p/rust-libp2p/pull/2622
- [ ] use the new transport in smoldot

```
ICE conn state: checking
main.js:1 ICE conn state: connected
main.js:1 conn state: connecting
main.js:1 'data' opened
main.js:1 conn state: connected
main.js:1 new message on 'data': 'WZaAkRgSkjiGwBw'
main.js:1 new message on 'data': 'dcgSNLysAvsyMJf'
main.js:1 new message on 'data': 'OZvQJpcHmpSdApM'
```


melekes:
TODO:
- [ ] read SDP spec https://datatracker.ietf.org/doc/html/rfc3264
- [ ] modify SDP in webrtc server (and check one in client)
- [ ] working connection

Questions:
1. do we need multiple UDP ports for ICE traffic on the server?
2. is it possible to disable DTLS client's certificate verification? or all clients will have the same certificate too similar to servers?

tomaka:
> do we need multiple UDP ports for ICE traffic on the server?

No

> is it possible to disable DTLS client's certificate verification? or all clients will have the same certificate too similar to servers?

I would prefer to disable DTLS client verification.
Hacks such as using the same certificate everywhere are necessary because we can't control the browser. However we're totally in control of the code on the server side, so we should do what is the most appropriate.

> modify SDP in webrtc server (and check one in client)

I'm not sure that's the right approach. If I'm not mistaken, the server doesn't need to manipulate SDP at all, unless it makes it easier to interact with the API of the `webrtc` library.

melekes:
> I would prefer to disable DTLS client verification.

trying to figure out how to do it right now. It's possible in Go library, but not in webrtc-rs, which suggest we might need to tweak it.

> I'm not sure that's the right approach. If I'm not mistaken, the server doesn't need to manipulate SDP at all, unless it makes it easier to interact with the API of the webrtc library.

Yes for the most part. I was talking about modifying some fields like IP (of the offerer, ie. client) and media's timestamp

bin/wasm-node/javascript/src/webrtc/index.ts
```diff
@@ -210,20 +210,30 @@ export default function(targetIp: string, protocol: 'tcp' | 'udp', targetPort: n
             // The maximum SCTP user message size (in bytes) (RFC8841)
             "a=max-message-size:100000" + "\n" +
             // A transport address for a candidate that can be used for connectivity checks (RFC8839).
-            "a=candidate:0 1 " + (protocol == 'tcp' ? "TCP" : "UDP") + " 2113667327 " + targetIp + " " + targetPort + " typ host" + "\n";
+            "a=candidate:1 1 " + (protocol == 'tcp' ? "TCP" : "UDP") + " 2113667327 " + targetIp + " " + targetPort + " typ host" + "\n" +
+            // Indicate that there will be no further candidates.
+            "a=end-of-candidates" + "\n";
```
tomaka, line 215: From what I remember, I believe that technically you need to change the `a=ice-options:` attribute to the version that supports this

melekes: right, it seems that it requires `ice-options:trickle`. removed

bin/webrtc-server/README.md
```diff
@@ -0,0 +1,51 @@
+# WebRTC server
+
+## Install
+
+The certificate in `./static` directory **MUST** be marked as trusted (Chrome:
+'Security and privacy' -> 'Security' -> 'Manage certificates').
```
wemeetagain, line 6: Why is this necessary?

melekes:
Closing. The work will be continued in https://github.com/libp2p/rust-libp2p/pull/2622

bin/webrtc-server/README.md
melekes: You can discard that. It turned out to be another issue with the library I'm using.

PR #2138, opened 2022-03-13
limit codegen-units to get the maximum optimization

sabify:
Hello,

There might be a performance regression due to enabling lto without limiting the number of code generation units (codegen-units = 1).

References:
https://github.com/rust-lang/rust/issues/47745
https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,6 +135,9 @@ codegen-units = 1
 inherits = "release"
 lto = true
 opt-level = "z"
+# https://github.com/rust-lang/rust/issues/47745
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1
 
 [[bench]]
 name = "header"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,6 +126,9 @@ opt-level = 3
 [profile.release]
 panic = "abort"
 lto = true
+# https://github.com/rust-lang/rust/issues/47745
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1
 #strip = "symbols"      # TODO: uncomment once stable
 
 [profile.min-size-release]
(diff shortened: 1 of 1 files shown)


melekes (commented):
the change makes sense to me. thanks for contributing 🙏

Cargo.toml
```diff
@@ -126,6 +126,9 @@ opt-level = 3
 [profile.release]
 panic = "abort"
 lto = true
+# https://github.com/rust-lang/rust/issues/47745
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1
 #strip = "symbols"      # TODO: uncomment once stable
 
 [profile.min-size-release]
```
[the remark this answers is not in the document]
sabify: Done.

Cargo.toml
```diff
@@ -126,12 +126,18 @@ opt-level = 3
...
 inherits = "release"
 lto = true
 opt-level = "z"
+# https://github.com/rust-lang/rust/issues/47745
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1
```
tomaka, line 140: That's normally not needed, since we have `inherits=release` above

Cargo.toml
```diff
@@ -126,12 +126,18 @@ opt-level = 3
 [profile.release]
 panic = "abort"
 lto = true
+# https://github.com/rust-lang/rust/issues/47745
```
tomaka, line 129: Could you add a small text?

```suggestion
# codegen-units set to 1 to avoid performance regressions when combined with LTO
# See https://github.com/rust-lang/rust/issues/47745
```

Cargo.toml
sabify: I've added the duplication upon @melekes request and I just removed it. I also removed the `lto` that was a duplication of release profile.

Cargo.toml
sabify: I just applied your suggestion.

tomaka (approved):
Thank you!

PR #2140, opened 2022-03-14
Refactor block announces to allow for user datas

tomaka:
This is the continuation of https://github.com/paritytech/smoldot/pull/2129 and https://github.com/paritytech/smoldot/issues/1572

The `block_announce` method now returns an enum, and in the enum you will find other objects that let you insert or update the block.
The changes are very similar to the ones in https://github.com/paritytech/smoldot/pull/2129


melekes (approved):
👍

PR #2142, opened 2022-03-15
Refactor AllForksSync::add_source

tomaka:
Continuation of https://github.com/paritytech/smoldot/pull/2140

The changes are the same: the function only looks at the state of the block, then returns an enum that allows inserting the block.

### DIFF
--- a/src/sync/all.rs
+++ b/src/sync/all.rs
@@ -372,14 +372,19 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
                 let outer_source_id_entry = self.shared.sources.vacant_entry();
                 let outer_source_id = SourceId(outer_source_id_entry.key());
 
-                let source_id = all_forks.prepare_add_source(
+                let source_id = match all_forks.prepare_add_source(
                     AllForksSourceExtra {
                         user_data,
                         outer_source_id,
                     },
                     best_block_number,
                     best_block_hash,
-                );
+                ) {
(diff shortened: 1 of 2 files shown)


src/sync/all.rs
```diff
@@ -372,14 +372,25 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
...
-                    best_block_hash,
-                );
+                let source_user_data = AllForksSourceExtra {
+                    user_data,
+                    outer_source_id,
+                };
+
+                let source_id =
+                    match all_forks.prepare_add_source(best_block_number, best_block_hash) {
+                        all_forks::AddSource::BestBlockAlreadyVerified(b) => {
```
melekes, line 382: can we combine these 3 branches?

src/sync/all.rs
```diff
@@ -2275,14 +2286,23 @@ impl<TRq> Shared<TRq> {
...
-            );
+            let source_user_data = AllForksSourceExtra {
+                user_data: source.user_data,
+                outer_source_id: source.outer_source_id,
+            };
+
+            let updated_source_id = match all_forks
+                .prepare_add_source(source.best_block_number, source.best_block_hash)
+            {
+                all_forks::AddSource::BestBlockAlreadyVerified(b) => b.add_source(source_user_data),
```
melekes, line 2297: same

melekes (approved):
👍

src/sync/all.rs
tomaka: Can't combine all three, as the types are different

Issue #2139, opened 2022-03-13
Add frontier support


xcaptain:
Currently smoldot doesn't support decode frontier block, I got a 
```
ProtocolError(error=Failed to decode a block announcement) 
```

The frontier concensus engine id is FRON, https://github.com/paritytech/frontier/blob/master/primitives/consensus/src/lib.rs#L28

tomaka:
For now I will just add the possibility for smoldot to parse Frontier blocks, and keep this issue open to decide on a proper solution later.

@xcaptain Do you have a Frontier block header available somewhere by any chance, that I could use for a unit test?

xcaptain:
@tomaka I think we can't just ignore the fron item. The frontier consensus algorithm may be a replacement of babe(not sure).
Our chain uses frontier, an example block can be found at  https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fmainnet-deeper-chain.deeper.network#/explorer/query/3234675 
its raw block file was generated using the `inspect` subcommand can be found at
https://gist.github.com/xcaptain/84cd84a979fe4da96b22916174f76612

I also found that moonbeam is using frontier, give its block 608077 as an example, the consensue engine id is also fron not babe
![image](https://user-images.githubusercontent.com/4054836/158363705-0c573b63-c0e7-4a84-a6c3-a12d948f0837.png)

tomaka:
Just want to point out that `nmbs` isn't supported. From a brief look into it, I don't think it's really light-client-compatible, but I haven't really digged much.

PR #2150, opened 2022-03-17
Add support for the "fron" log item

tomaka:
Close https://github.com/paritytech/smoldot/issues/2139

I wanted to add a unit test, but we have no good way to obtain a SCALE-encoded header from a Substrate node.
Since the code is just copy-pasted from the one of the Beefy and Polkadot log items, it should work.


xcaptain:
Thanks, tested and it works.

melekes (approved):
👍

PR #2151, opened 2022-03-17
Add ClientOptions.cpuRateLimit

tomaka:
Adding an option to limit the amount of CPU consumed by smoldot.

Warp syncing a long chain such as Kusama can use a lot of CPU for up to 10-15 seconds, and we'd rather not have a web page or the substrate-connect extension use up an entire CPU for this long, as it tends to cause systems to freeze.
After this change is merged and published, I'd cap the CPU to 50% in substrate-connect.

### DIFF
--- a/bin/wasm-node/rust/src/cpu_rate_limiter.rs
+++ b/bin/wasm-node/rust/src/cpu_rate_limiter.rs
@@ -1,0 +1,103 @@
+// Smoldot
+// Copyright (C) 2019-2022  Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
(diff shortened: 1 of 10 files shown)


bin/wasm-node/rust/src/cpu_rate_limiter.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context) -> Poll<Self::Output> {
+        let mut this = self.project();
+
+        // Note that `crate::times::Delay` is a `FusedFuture`, making it ok to call `poll` even
+        // if it is possible for the `Delay` to already be resolved.
+        // We add a small zero-cost shim to ensure at compile time that this is indeed the case.
+        fn enforce_fused<T: futures::future::FusedFuture>(_: &T) {}
+        enforce_fused(&this.prevent_poll_until);
+        match Future::poll(this.prevent_poll_until.as_mut(), cx) {
```
melekes, line 68: nitpik: since we're interested in only one arm

```suggestion
        if let Poll::Pending = Future::poll(this.prevent_poll_until.as_mut(), cx) {
```

melekes (approved):
👍

PR #2154, opened 2022-03-21
Pass a string in chainAddedErr to remove an as

tomaka:
Very insubstantial change with the objective of removing an `as` in TypeScript


melekes (approved):
👍

PR #2155, opened 2022-03-21
Add a user_data for pending blocks in all_forks.rs

tomaka:
Continuation of https://github.com/paritytech/smoldot/pull/2142 and https://github.com/paritytech/smoldot/issues/1572
This is the last step of the modifications to `all_forks.rs`

This now stores a "user data" for pending blocks in `all_forks.rs` as well, and adjusts all the APIs.

I haven't touched any of the API of `all.rs`. Instead I've made it so that `all.rs` stores a `None` in the `AllForksSync` when the block is still pending, and replaces it with a `Some` when it no longer is.


melekes (approved):
👍

PR #2156, opened 2022-03-21
Fix hack in the block announces format

tomaka:
At the moment the decoded block announces struct contains at the same time `header` (decoded) and `scale_encoded_header` (encoded)

The reason why there are two is because the block announces format concatenates the header with the other fields, meaning that it is not possible to decode block announces without decoding the header. In other words, if a block announces has been successfully decoded, then the header is necessarily decodable successfully as well.
However, this fact should be seen as a quirk in the block announces format. It would be better if you could decode the announce without decoding the header, and it's not completely impossible that we do this in the long future.

This PR removes `header` and commits to the fact that `scale_encoded_header` might be undecodable. It is still the case that the header is always decodable, but it is not longer guaranteed in the API.

### DIFF
--- a/bin/light-base/src/sync_service/parachain.rs
+++ b/bin/light-base/src/sync_service/parachain.rs
@@ -483,16 +483,15 @@ pub(super) async fn start_parachain<TPlat: Platform>(
                                 let decoded_header_hash = header::hash_from_scale_encoded_header(
                                     &decoded.scale_encoded_header
                                 );
-                                let decoded_header_number = decoded_header.number;
                                 sync_sources.add_known_block(
                                     local_id,
-                                    decoded_header_number,
+                                    decoded_header.number,
                                     decoded_header_hash
                                 );
                                 if decoded.is_best {
                                     sync_sources.add_known_block_and_set_best(
                                         local_id,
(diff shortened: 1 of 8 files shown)


bin/light-base/src/sync_service/parachain.rs
```diff
@@ -479,10 +479,23 @@ pub(super) async fn start_parachain<TPlat: Platform>(
...
                             let decoded = announce.decode();
-                            let decoded_header_hash = decoded.header.hash();
-                            sync_sources.add_known_block(local_id, decoded.header.number, decoded_header_hash);
-                            if decoded.is_best {
-                                sync_sources.add_known_block_and_set_best(local_id, decoded.header.number, decoded_header_hash);
+                            if let Ok(decoded_header) = header::decode(&decoded.scale_encoded_header) {
+                                let decoded_header_hash = header::hash_from_scale_encoded_header(
+                                    &decoded.scale_encoded_header
+                                );
+                                let decoded_header_number = decoded_header.number;
```
melekes, line 486: nitpik: not sure we need this. a) same length b) no speed / other benefits

melekes (approved):
👍

PR #2165, opened 2022-03-22
Give more access to user datas of blocks in all_forks.rs

tomaka:
cc https://github.com/paritytech/smoldot/issues/1572

### DIFF
--- a/src/sync/all_forks.rs
+++ b/src/sync/all_forks.rs
@@ -91,7 +91,7 @@ use alloc::{
     borrow::ToOwned as _,
     vec::{self, Vec},
 };
-use core::{num::NonZeroU32, ops, time::Duration};
+use core::{mem, num::NonZeroU32, ops, time::Duration};
 
 mod disjoint;
 mod pending_blocks;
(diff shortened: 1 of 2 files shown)


src/sync/all_forks.rs
```diff
@@ -995,9 +995,36 @@ pub struct AddBlockOccupied<TBl, TRq, TSrc> {
...
 impl<TBl, TRq, TSrc> AddBlockOccupied<TBl, TRq, TSrc> {
+    /// Gives access to the user data of the block.
+    pub fn user_data_mut(&mut self) -> &mut TBl {
+        if self.is_verified {
+            &mut self
+                .inner
+                .inner
+                .chain
+                .non_finalized_block_by_hash(&self.inner.expected_next_hash)
+                .unwrap()
```
melekes, line 1006: safe to unwrap?

melekes (approved):
👍

src/sync/all_forks.rs
tomaka: Yeah, a `AddBlockOccupied` struct is only constructed only for blocks that exist.

PR #2170, opened 2022-03-23
Print warning when GrandPa warp syncing takes a long time

tomaka:
One of the main consequence of issues right now is that the GrandPa warp syncing never finishes, for example because we have no connectivity to any peer or because we fail to decode some data structure or something.

This PR prints a warning every 15 seconds in order to remind users of the fact that smoldot is still trying to do something.


melekes (approved):
👍

PR #2178, opened 2022-03-27
Some tweaks to finality block reports in the sync_service

tomaka:
Avoids reporting the finalized block twice, and clarifies the use of two variables.


melekes (approved):
👍

PR #2180, opened 2022-03-28
Simplifications and clean ups to the runtime_service

tomaka:
See individual commits.

This attempts at cleaning up the code of the service and making it easier to read and understand.


bin/light-base/src/runtime_service.rs
```diff
@@ -593,98 +208,102 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
...
+            (
+                tree.finalized_async_user_data().clone(),
+                *decoded_finalized_block.state_root,
+                decoded_finalized_block.number,
+            ),
+        );
+
+        let non_finalized_blocks_ancestry_order: Vec<_> = tree
+            .input_iter_ancestry_order()
+            .filter_map(|block| {
```
melekes, line 259: why do we need `filter` if we always return `Some`?

bin/light-base/src/runtime_service.rs
```diff
@@ -593,98 +208,102 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
...
+                        || tree
+                            .input_iter_ancestry_order()
+                            .any(|b| parent_hash == b.user_data.hash
+                                && b.async_op_user_data.is_some())
+                );
 
-                guarded_lock.pinned_blocks.insert(
-                    (subscription_id, finalized_block.hash),
+                let decoded_header = header::decode(&block.user_data.scale_encoded_header).unwrap();
+                pinned_blocks.insert(
```
melekes, line 280: for some reason mutating the state while collecting items feels bad, but can't find a good reason why. feel free to ignore

bin/light-base/src/runtime_service.rs
```diff
@@ -765,30 +371,35 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
...
+                .is_none()
             {
-                panic!("block already unpinned");
-            } else {
-                return;
-            }
-        };
+                // Cold path.
+                if all_blocks_subscriptions.contains_key(&subscription_id.0) {
+                    panic!("block already unpinned");
```
melekes, line 388: is there a guidance around the use of panics in the code? Can one use them even if they are not terminal (i.e. state isn't corrupted | program can continue to run)?

bin/light-base/src/runtime_service.rs
```diff
@@ -1920,88 +1565,67 @@ impl<TPlat: Platform> Background<TPlat> {
                             }
                         }
                         for to_remove in to_remove {
```
melekes, line 1567: can we extract this into a function since it's duplicated here and above?

bin/light-base/src/runtime_service/sub_utils.rs
```diff
@@ -0,0 +1,408 @@
...
+    ///
+    /// The stream is infinite. In other words it is guaranteed to never return `None`.
+    pub async fn subscribe_runtime_version(
+        &self,
+    ) -> (
+        Result<executor::CoreVersion, RuntimeError>,
+        stream::BoxStream<'static, Result<executor::CoreVersion, RuntimeError>>,
+    ) {
+        let mut master_stream = stream::unfold(self.guarded.clone(), |guarded| async move {
+            let subscribe_all = Self::subscribe_all_inner(&guarded, 16, 32).await;
```
melekes, line 40: magic numbers?

bin/light-base/src/runtime_service.rs
tomaka: It's not always `Some`, the line just below the `filter_map` has a `?`

bin/light-base/src/runtime_service.rs
tomaka: Ah I agree, that's an accident (wasn't the case before the PR) and I'll change that.

bin/light-base/src/runtime_service.rs
tomaka: The guidance is to use them when the state of the program would be corrupted. And that's the case here: while the state of the `runtime_service` itself isn't corrupted, the state of the stuff on top most likely is. Unpinning the same block twice is a bit like deleting the same object twice.

bin/light-base/src/runtime_service.rs
tomaka: I tried, actually. The code you've highlighted, however, has a bit more logic than the code above, and that extra logic uses variables that are only in this scope. I think that any attempt at de-duplicating this would actually make the code way more complicated.

melekes: Okay

bin/light-base/src/runtime_service/sub_utils.rs
tomaka: Yeah, I agree it's not great. These functions are only used in order to respond to some JSON-RPC methods, and so we have no choice but to use magic numbers anyway. The question is just where to put these magic numbers.

I think these functions should be moved to the `json_rpc_service`. I'll do that in a follow-up PR if you don't mind.

PR #2106, opened 2022-02-25
Don't notify the same parahead twice

tomaka:
Fix #2088 

See #2088 for context.
What this PR fixes is that we no longer notify about a new parachain block if the same parachain block was already in another relay chain fork.

PR #2182, opened 2022-03-28
Fix #2088 again

### LINKED ISSUE #2088: Parachain (Westmint) connectivity (without non-ws bootnodes) results to panicked smoldot

### LINKED ISSUE #2147: Block already unpinned error
An error appeared during zombienet tests in substrate connect
```
Error: panicked at 'block already unpinned', /__w/smoldot/smoldot/bin/light-base/src/runtime_service.rs:779:17
```
Logs can be found [here](https://github.com/paritytech/substrate-connect/runs/5571696271?check_suite_focus=true#step:7:911)

tomaka:
Fix #2088 again
Fix #2147 as well

This is a fix of https://github.com/paritytech/smoldot/pull/2106, which was the previous attempt at fixing #2088 

What https://github.com/paritytech/smoldot/pull/2106 didn't do is check whether the block to report is equal to the finalized block.

What happens is:

- When the parachain sync service initializes, it starts to fetch the parahead corresponding the latest relay chain finalized block and to fetch the parahead corresponding to each relay chain non-finalized block
- The fetch of the relay chain finalized block succeeds, and we report this block. We not only reports it as new block but also as finalized.
- The fetch of one of the relay chain non-finalized blocks succeeds, and the code that this PR modifies runs. But because the code before this PR only looks at non-finalized blocks, it doesn't detect that we have already reports this new block, and thus we report it again.

As a result, we have two different blocks with the same hash, which isn't supposed to happen, and triggers a panic in the state machines on top.

### DIFF
--- a/bin/light-base/src/sync_service/parachain.rs
+++ b/bin/light-base/src/sync_service/parachain.rs
@@ -247,11 +247,12 @@ pub(super) async fn start_parachain<TPlat: Platform>(
                         // past. This covers situations where the parahead is identical to the
                         // relay chain's parent's parahead, but also situations where multiple
                         // sibling relay chain blocks have the same parahead.
-                        if async_tree
-                            .input_iter_unordered()
-                            .filter(|item| item.id != block_index)
-                            .filter_map(|item| item.async_op_user_data)
-                            .any(|item| item.as_ref() == Some(&scale_encoded_header))
+                        if finalized_parahead == scale_encoded_header
+                            || async_tree
+                                .input_iter_unordered()
+                                .filter(|item| item.id != block_index)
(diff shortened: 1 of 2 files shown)


melekes (approved):
👍

PR #2184, opened 2022-03-29
Add default-features=false to parity-scale-codec

tomaka:
This is an oversight, and might reduce the Wasm binary size.


melekes (approved):
👍

PR #2188, opened 2022-03-31
Cap the maximum value passed to setTimeout

tomaka:
Yet another surprise factor from JavaScript:

- https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#maximum_delay_value
- https://nodejs.org/api/timers.html#settimeoutcallback-delay-args


melekes (approved):
👍 

js never stops to amuse me

PR #2189, opened 2022-03-31
Correctly implement the cpuRateLimit feature

tomaka:
Correctly implements https://github.com/paritytech/smoldot/pull/2151

The code currently does `u32::max_value().checked_sub(rate_limit)`. This is wrong, it should be `checked_div`.
I remember that when I was testing https://github.com/paritytech/smoldot/pull/2151 I was doing `u32::max_value() / rate_limit`, and then at the last second before opening the PR I replaced this with the `checked_` function in order to handle the situation where `rate_limit` is 0.

In addition to fixing this mistake, I've also changed the code to use floating points instead of integers.
Right now the precision is actually very low: because we round the division to the nearest integer, the possible values for the CPU bound can only be 100%, 50%, 25%, 12.5%, etc. Using floating points fixes this.

### DIFF
--- a/bin/wasm-node/javascript/demo/demo.js
+++ b/bin/wasm-node/javascript/demo/demo.js
@@ -38,6 +38,7 @@ const client = smoldot.start({
     forbidWs: false,
     forbidNonLocalWs: false,
     forbidWss: false,
+    cpuRateLimit: 0.5,
     logCallback: (level, target, message) => {
         // As incredible as it seems, there is currently no better way to print the current time
         // formatted in a certain way.
--- a/bin/wasm-node/rust/src/cpu_rate_limiter.rs
+++ b/bin/wasm-node/rust/src/cpu_rate_limiter.rs
@@ -32,7 +32,7 @@ use core::{
 pub struct CpuRateLimiter<T> {
     #[pin]
     inner: T,
-    max_divided_by_rate_limit: u32,
+    max_divided_by_rate_limit_minus_one: f64,
 
     /// Prevent `self.inner.poll` from being called before this `Delay` is ready.
     #[pin]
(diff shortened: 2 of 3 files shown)


bin/wasm-node/javascript/demo/demo.js
```diff
@@ -38,6 +38,7 @@ const client = smoldot.start({
     forbidWs: false,
     forbidNonLocalWs: false,
     forbidWss: false,
+    cpuRateLimit: 0.5,
```
tomaka, line 41: Note: I've intentionally left this in the demo so that it mimics what substrate-connect will now do (https://github.com/paritytech/substrate-connect/pull/900)

bin/wasm-node/rust/src/cpu_rate_limiter.rs
```diff
@@ -81,13 +87,21 @@ impl<T: Future> Future for CpuRateLimiter<T> {
...
-                .saturating_sub(poll_duration);
+                // The base equation here is: `(after_poll_sleep + poll_duration) * rate_limit == poll_duration * u32::max_value()`.
+                // Because `Duration::mul_f64` and `Duration::from_secs_f64` panic in case of
+                // overflow, we need to do the operation of multiplying and checking the bounds
+                // manually.
+                let mut after_poll_sleep =
+                    poll_duration.as_secs_f64() * *this.max_divided_by_rate_limit_minus_one;
+                debug_assert!(after_poll_sleep >= 0.0 && !after_poll_sleep.is_nan());
+                let max_duration_float: f64 = Duration::MAX.as_secs_f64(); // TODO: turn this into a `const` once `as_secs_f64` is `const`
+                if !(after_poll_sleep < max_duration_float) {
```
melekes, line 98:
```suggestion
                if after_poll_sleep >= max_duration_float {
```

bin/wasm-node/rust/src/cpu_rate_limiter.rs
```diff
@@ -44,11 +44,17 @@ impl<T> CpuRateLimiter<T> {
     /// `u32::max_value()` represents "one CPU". For example passing `rate_limit / 2` represents
     /// "50% of one CPU".
     pub fn new(inner: T, rate_limit: u32) -> Self {
+        let max_divided_by_rate_limit_minus_one =
+            (f64::from(u32::max_value()) / f64::from(rate_limit)) - 1.0;
+        // Note that it is legal for `max_divided_by_rate_limit_minus_one` to be +infinite
+        debug_assert!(
```
melekes, line 50: is there a guidance on when to use debug_asserts? don't you think they tend to clutter the code?

melekes (approved):
👍

bin/wasm-node/rust/src/cpu_rate_limiter.rs
tomaka: I don't think I've wrote this down, but I use them for internal-bug-detecting purposes.
In other words, what you put in the `debug_assert!` is something that, if false, indicates that there's a bug in the internal logic of the current module.
However if it's an incorrect API usage, then it should be `assert!` instead.

PR #2194, opened 2022-04-01
Tweak runtime_service::RuntimeLock::specification

tomaka:
The function doesn't need to take `&mut self`.
Also changes the error type. The error right now is only used to be turned into a string and sent back as a JSON-RPC error in case the runtime is invalid.


melekes (approved):
👍

PR #2201, opened 2022-04-04
Fix `when_known` not notified

tomaka:
Fixed a bug introduced in https://github.com/paritytech/smoldot/pull/2180

Before https://github.com/paritytech/smoldot/pull/2180, we were detecting when the runtime is ready by subscribing to notifications about the chain, and then reading the state again when something happens on this subscription.
This is hacky, so I replaced it with a `when_known: EventListener` that gets notified when the runtime is ready.

Unfortunately, I forgot to notify this `EventListener` in case the sync service notification resets.

The problematic situation is:

- You call `RuntimeService::subscribe_all` at the very beginning. The runtime service is in the `FinalizedBlockRuntimeUnknown` state, and we listen for events on the `when_ready`.
- The sync service finishes its warp sync, and resets all its notifications subscribers.
- This causes the `RuntimeService` to reset and to re-subscribe to the sync service.
- Because the sync service has a runtime already ready, we immediately transition to `FinalizedBlockRuntimeKnown`, but without notifying the `when_ready`. This is what this PR fixes.

Because some JSON-RPC functions indirectly rely on `RuntimeService::subscribe_all`, this bug causes these JSON-RPC functions to never produce any answer if they are called at the very beginning.


melekes (approved):
👍

Issue #2204, opened 2022-04-04
Panicked at 'called `Result::unwrap()` on an `Err` value: InvalidPort'


wirednkod:
When trying to connect to a custom relay chain and parachain with substrate-connect (smoldot 0.6.10) the following error is received:
```tsx
bindings-smoldot-light.jss:33 Uncaught (in promise) Error: panicked at 'called `Result::unwrap()` on an `Err` value: InvalidPort', /__w/smoldot/smoldot/bin/light-base/src/lib.rs:453:70
    at panic (bindings-smoldot-light.jss:33:1)
    at 00ec1f46:0x2b667
    at 00ec1f46:0x4344
    at 00ec1f46:0x248937
    at 00ec1f46:0x248019
    at 00ec1f46:0x24860c
    at 00ec1f46:0x253dde
    at 00ec1f46:0x254e91
    at 00ec1f46:0x1579b
    at 00ec1f46:0x137ee
```

Attached can be found the [extension logs](https://github.com/paritytech/smoldot/files/8410023/Error-logs.txt)
and the specs used for relay ([rococo-local-0.9.16 - relay specs.txt](https://github.com/paritytech/smoldot/files/8410011/rococo-local-0.9.16.-.relay.specs.txt)) and parachain
([Staging-latest-parachain-1.7.4 - parachain specs.txt…]())

PR #2207, opened 2022-04-05
Do not panic when passed an invalid bootnode address

tomaka:
Close https://github.com/paritytech/smoldot/issues/2204

We no longer panic when the chain spec contains an invalid bootnode address.

As I've commented in the code, it is not possible to be sure that a bootnode address is actually invalid. We can only be sure that smoldot cannot parse it, but that's not the same as saying that it's invalid.
For this reason, invalid bootnode addresses are simply ignored in the light client.

### DIFF
--- a/bin/light-base/src/lib.rs
+++ b/bin/light-base/src/lib.rs
@@ -452,20 +452,35 @@ impl<TChain, TPlat: Platform> Client<TChain, TPlat> {
         };
 
         // Build the list of bootstrap nodes ahead of time.
-        // TODO: either leave this here if it's fallible, or move below if it's not fallible
-        let bootstrap_nodes = {
-            let mut list = Vec::with_capacity(chain_spec.boot_nodes().len());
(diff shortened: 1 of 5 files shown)


bin/light-base/src/lib.rs
```diff
@@ -452,20 +452,35 @@ impl<TChain, TPlat: Platform> Client<TChain, TPlat> {
...
-                    address.pop();
-                    list.push((peer_id, vec![address]));
-                } else {
-                    panic!() // TODO:
+                match node {
+                    chain_spec::Bootnode::Parsed { multiaddr, peer_id } => {
+                        if let Ok(multiaddr) = multiaddr.parse::<multiaddr::Multiaddr>() {
+                            let peer_id = peer_id::PeerId::from_bytes(peer_id).unwrap();
+                            valid_list.push((peer_id, vec![multiaddr]));
+                            continue;
```
melekes, line 470: do we really need continue here?

melekes (approved):
👍

PR #2210, opened 2022-04-05
Properly fix the regression again

tomaka:
This is https://github.com/paritytech/smoldot/pull/2201 again

This time I've made sure that `when_known` was indeed notified before every time we do `lock.tree = ...`.


tomaka:
Merging as insubstantial

Issue #2209, opened 2022-04-05
Parachains can take a long time to initialize if relay chain sync service is busy


tomaka:
The runtime service calls [this line](https://github.com/paritytech/smoldot/blob/1baa69b6161e9f6fd5b7f8f59a693ee770ed92cb/bin/light-base/src/runtime_service.rs#L126) to initialize.

This has the effect of queuing a message on the channel from the sync service frontend to backend.
Unfortunately, if the backend is busy verifying things such as warp sync fragments, it will not process the messages in this channel, leading to the call to `is_near_head_of_chain_heuristic` not being answered until the verifications finish.

PR #2213, opened 2022-04-06
Interleave CPU-heavy operations with I/O events in the sync service

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2209

Right now, if the sync service is verifying a warp sync proof, in the meanwhile it doesn't respond to any event anymore.
This causes queues of operations to grow in size, and thus in latency, which causes other kind of small problems to manifest.

This PR makes sure that the sync service processes events at a certain minimum rate by interleaving CPU-heavy operations with responding to events.


bin/light-base/src/sync_service/standalone.rs
```diff
@@ -94,16 +94,47 @@ pub(super) async fn start_standalone_chain<TPlat: Platform>(
     let mut from_network_service = from_network_service.fuse();
 
     // Main loop of the syncing logic.
+    //
+    // This loop contains some CPU-heavy operations (e.g. verifying justifications and warp sync
```
melekes, line 98: maybe stupid question, but would it make sense to use threads to separate verification and foreground ops? That way we won't have to play scheduler ourselves

bin/light-base/src/sync_service/standalone.rs
```diff
@@ -389,307 +433,290 @@ struct Task<TPlat: Platform> {
...
-                                header: request_headers,
-                                body: request_bodies,
-                                justifications: request_justification,
-                            },
+        // Before inserting the request back to the syncing state machine, clamp the number
+        // of blocks to the number of blocks we expect to receive.
+        // This constant corresponds to the maximum number of blocks that nodes will answer
+        // in one request. If this constant happens to be inaccurate, everything will still
+        // work but less efficiently.
+        request_detail.num_blocks_clamp(NonZeroU64::new(64).unwrap());
```
melekes, line 458: can we extract this into a const?

bin/light-base/src/sync_service/standalone.rs
```diff
@@ -389,307 +433,290 @@ struct Task<TPlat: Platform> {
...
+                                    proof: outcome.iter().map(|nv| &nv[..]),
+                                    requested_key: key.as_ref(),
+                                    trie_root_hash: &state_trie_root,
                                 })
-                                .collect::<Result<Vec<_>, ()>>()
-                        } else {
-                            Err(())
-                        }
-                    };
+                                .map_err(|_| ())
```
melekes, line 558: 😮

bin/light-base/src/sync_service/standalone.rs
```diff
@@ -389,307 +433,290 @@ struct Task<TPlat: Platform> {
...
+    fn start_next_request(&mut self) -> bool {
+        // `desired_requests()` returns, in decreasing order of priority, the requests
+        // that should be started in order for the syncing to proceed. The fact that multiple
+        // requests are returned could be used to filter out undesired one. We use this
+        // filtering to enforce a maximum of one ongoing request per source.
+        let (source_id, _, mut request_detail) = match self
+            .sync
+            .desired_requests()
+            .find(|(source_id, _, _)| self.sync.source_num_ongoing_requests(*source_id) == 0)
+        {
```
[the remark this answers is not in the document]
tomaka: This appears in the diff only because I've changed an indentation level. I don't want to touch this code.

bin/light-base/src/sync_service/standalone.rs
tomaka: Same remark. I've just changed the indentation level.

bin/light-base/src/sync_service/standalone.rs
tomaka: Wasm is always single threaded. Additionally, the complexity of everything would probably be multiplied by 10 if you introduced concurrency in this code.

melekes (approved):

PR #2214, opened 2022-04-06
Move sub_utils under json_rpc_service/state_chain

tomaka:
As promised in https://github.com/paritytech/smoldot/pull/2180, this moves the `sub_utils` module to `json_rpc_service/state_chain` where it is more appropriate.

Can be easily reviewed commit by commit.


melekes (approved):
🚀

Issue #1920, opened 2022-01-19
Infinite allocating loop


tomaka:
Occasionally, smoldot seems to become unresponsive, and then a few seconds later panic because of an allocation error.
The node normally doesn't use more than 100 MiB, so we really shouldn't reach OOM.
This lets me think that there might be an infinite loop somewhere that constantly allocates, for example by repeatedly pushing to a `Vec` or something like that.

tomaka:
Obviously unable to reproduce after adding more debugging

tomaka:
It's when growing a `Vec` some place within the `tx-service`

tomaka:
Seeing the timing, was most likely introduced in https://github.com/paritytech/smoldot/pull/1892

tomaka:
Also happened within `json-rpc-...` task. Probably actually something in the sync or runtime service.

tomaka:
Just had this in `tx-service` despite no transaction being submitted, so this is likely not caused for example by some network request.

tomaka:
Could maybe be fixed by https://github.com/paritytech/smoldot/pull/2030, but it's not sure.

tomaka:
Just happened again.

PR #2215, opened 2022-04-06
Ignore the maximum number of pinned blocks when irrelevant

tomaka:
Fix https://github.com/paritytech/smoldot/issues/1920

This PR tweaks the calls to `RuntimeService::subscribe_all` to pass `usize::max_value()` for the maximum number of pinned blocks in situations where it is appropriate.
As a second minor change, this PR also adjusts all the buffer sizes to 32 in these situations.

## Context

The runtime service has a mechanism of "pinned blocks".
When the runtime service reports a new block, this block stays pinned in the memory of the runtime service and must later be unpinned. While a block is pinned, one can start runtime calls against this block.

The same logic of pinned blocks exists in the publicly-exposed JSON-RPC API. When a JSON-RPC client subscribes to blocks, these blocks stay pinned to guarantee that the client can access information about them, and must then be manually unpinned by the client.

This leads to a problem: a JSON-RPC client could simply never unpin blocks and grow the memory of the node forever.
To counter-act this, each subscription in the runtime service has a maximum number of pinned blocks. If any subscription grows over this limit, this subscription gets killed. In the case of the JSON-RPC API, this subscription being killed leads to a notification of type "dead" being sent to the client so that the client must resubscribe.

But this maximum number of pinned blocks has a flaw in its design: if you start a subscription with a maximum of N pinned blocks, and there's already more than N non-finalized blocks to report, then the subscription gets killed immediately.
Some pieces of code, such as the transactions service, react to a subscription being killed by resubscribing immediately. But doing so doesn't solve the problem, as this new subscription will also get killed immediately, and the service will be stuck in an infinite loop.

Because this maximum number of pinned blocks is a mechanism against malicious behavior, and that our own code is by definition not malicious (it properly unpins blocks), we're now passing `usize::max_value()` for the maximum number of pinned blocks, effectively removing any maximum.

There is still a problem, though: when a JSON-RPC client subscribes, the maximum number of pinned blocks is set to an arbitrary constant of 48. Of course this limit shouldn't be decided by the JSON-RPC client, as the point of this limit is precisely to prevent the client from doing stupid things. The same problem as described above can then happen: if there are already more than 48 non-finalized blocks, the subscription will be killed immediately. Fortunately this will not result in an infinite loop, but is still kind of crappy.
For this I've added a TODO and will figure out later how to adjust the API. I think that `subscribe_all` could automatically add the current number of non-finalized blocks to the provided limit.
In any case, passing `usize::max_value()` would still be the correct thing to do for non-malicious code.

### DIFF
--- a/bin/light-base/src/json_rpc_service.rs
+++ b/bin/light-base/src/json_rpc_service.rs
@@ -591,9 +591,14 @@ impl<TPlat: Platform> Background<TPlat> {
 
                     // Subscribe to new runtime service blocks in order to push them in the
                     // cache as soon as they are available.
+                    // The buffer size should be large enough so that, if the CPU is busy, it
+                    // doesn't become full before the execution of this task resumes.
+                    // The maximum number of pinned block is ignored, as this maximum is a way to
+                    // avoid malicious behaviors. This code is by definition not considered
+                    // malicious.
                     let mut subscribe_all = me
                         .runtime_service
-                        .subscribe_all(8, cache.recent_pinned_blocks.cap())
+                        .subscribe_all(32, cache.recent_pinned_blocks.cap())
                         .await;
 
                     cache.subscription_id = Some(subscribe_all.new_blocks.id());
--- a/bin/light-base/src/json_rpc_service.rs
+++ b/bin/light-base/src/json_rpc_service.rs
@@ -598,7 +598,7 @@ impl<TPlat: Platform> Background<TPlat> {
                     // malicious.
                     let mut subscribe_all = me
                         .runtime_service
-                        .subscribe_all(32, cache.recent_pinned_blocks.cap())
+                        .subscribe_all(32, usize::max_value())
                         .await;
 
                     cache.subscription_id = Some(subscribe_all.new_blocks.id());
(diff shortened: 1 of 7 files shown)


tomaka:
Note that the buffer size (which this PR sets to 32) represents the number of notifications that can be buffered *after* the current already-existing non-finalized blocks.
This buffer size must simply be large enough to not be exceeded between the moment a task yields and the moment it is waken up again and can process this buffer.

melekes:
Thanks for the write-up! 

> and must then be manually unpinned by the client.

what about the case where the client disconnects? does he need to unpin blocks before that or it's done automatically?

> these blocks stay pinned to guarantee that the client can access information about them

can't we load them on demand? or this will be too slow?

> if you start a subscription with a maximum of N pinned blocks, and there's already more than N non-finalized blocks to report,

maybe N should mean "the number of finalized pinned blocks", not the sum of non-finalized and finalized?

melekes (approved):
lgtm, but see the above questions

tomaka:
> what about the case where the client disconnects? does he need to unpin blocks before that or it's done automatically?

That's done automatically.

> can't we load them on demand? or this will be too slow?

"Loading" a block that we no longer have in memory requires a network request, which can take a long time, use bandwidth, and can fail.
It's for this reason it's the other way around: blocks don't leave the memory until they're no longer needed.
This guarantees the JSON-RPC client that for example getting the header of a pinned block is always an instantaneous request that is always successful.

> maybe N should mean "the number of finalized pinned blocks", not the sum of non-finalized and finalized?

Yeah that's an option

tomaka:
I opened https://github.com/paritytech/smoldot/issues/2217 as a follow up

PR #2216, opened 2022-04-06
Yield between each request in the JSON-RPC service

tomaka:
In situations where there are many JSON-RPC requests queued up, the JSON-RPC service will currently process them all in a row without giving the chance to other tasks to do some work.

This PR changes that by yielding between each request.

Note that since there's a maximum number of requests that can be queued up, and that while the node is busy no new request can be queued up (new requests are queued up by pulling from a queue, which can't happen if we don't yield), this is not a way to DoS the node.
What this PR does is merely make sure that the tasks are properly interleaved and that the node can't "freeze" for long periods of time.


melekes (approved):
👍 🥇 🐎

PR #2219, opened 2022-04-07
Backport change to checkpoints format

tomaka:
https://github.com/paritytech/substrate/pull/9284 accidentally modified the checkpoints format. This PR backports this change.

We maintain compatibility with older checkpoints by not actually adding the new field but instead using `decode` instead of `decode_all`. Contrary to `decode_all`, `decode` succeeds even if there is data remaining after the decoded value, meaning that the additional field will not lead to an error.

This PR also updates the checkpoint in the Polkadot spec. I tested this PR with Westend (that uses an old checkpoint) and Polkadot (that uses a new checkpoint).


tomaka:
I've opened https://github.com/paritytech/substrate/issues/11184

melekes (approved):
👍

PR #2220, opened 2022-04-07
Publish v0.6.15

tomaka:


tomaka:
Tested all 4 well-known chains as they currently are in this repo, and as they are in the substrate-connect repo, and they all work.

PR #2232, opened 2022-04-12
Update to Rust 1.60

tomaka:


melekes (approved):

tomaka:
+109kiB is a bit much. I'm wondering what happened here.

tomaka:
So this CI comparison compares the .wasm output of the Rust compiler.

When we take into account the fact that this .wasm is then passed to `wasm-opt`, we have:
Before this PR: 2.612 MiB
After this PR: 2.599 MiB

So it's a win. Let's merge.

PR #2240, opened 2022-04-19
Don't copy the Rust .wasm output if not necessary

tomaka:
At the moment, if the user compiles in debug or if `wasm-opt` isn't installed, we copy the .wasm file generated by the Rust compiler to a temporary location and later delete that temporary location.

This is completely useless. The file is only read and not modified. This pull request removes this copy and makes us directly use the Rust output.


melekes (approved):
👍

PR #2247, opened 2022-04-25
Don't trigger the liveness timeout after a crash

tomaka:
The smoldot binary runs in a worker.
Every few seconds, the worker sends a "ping" message to the outside. If the outside doesn't detect a ping for a long time, it prints a big warning message in the logs, as this typically indicates that we're stuck in an infinite loop of some sort.
This mechanism is called the liveness timeout.

In parallel to this, if there is a panic in smoldot, the worker crashes and a big warning message is printed in the logs as well.
But then, later, the liveness timeout detects that there's no ping (which is normal, since the worker has stopped), and prints its own message.
This is weird and confusing. As such, this PR doesn't print the liveness timeout warning if it knows that a crash happened earlier.

An alternative would have been to disable the liveness timeout when a crash happens, but this was more complicated than what this PR does.


tomaka:
This size diff doesn't make sense.

melekes (approved):
👍

PR #2266, opened 2022-05-04
Rename wrongly-named function

tomaka:
This is purely a mistake. All the documentation already says "Merkle value".


tomaka:
Merging as insubstantial

Issue #1967, opened 2022-01-31
Backport "Inner hashing of value in state trie (runtime versioning)"


tomaka:
https://github.com/paritytech/substrate/pull/9732

tomaka:
After #1970 and #1971, what remains is the implementation of the actual changes in the trie. And that's going to be complicated, especially without any specification.

tomaka:
Some note for myself:

In the Substrate client, `ext_storage_root_version_2` always passes as parameter the same value as in `RuntimeSpec::state_version`. It is `RuntimeSpec::state_version` that is authoritative as to which version to use for the storage writes. In smoldot I'm going to panic if `ext_storage_root_version_2` gets passed a different value.

tomaka:
The most pressing change to make at the moment is implementing the V1 trie format so that we can generate the proper genesis hash root even if `state_version` is equal to 1, which is important for the light client to be able to connect to some chains.
Implementing `ext_storage_root_version_2` and all is more than secondary.

tomaka:
https://spec.polkadot.network/#defn-node-header

PR #2270, opened 2022-05-05
Optimize building ChainInformation and return runtime

tomaka:
In order to build the `ChainInformation` (containing the information about the finalized state of a chain) from a chain spec, we need to compile the runtime of the genesis block.

Before this PR, this is done in a very inefficient way: we build the runtime and use it to obtain the Aura config, then build the runtime again and use it to get the Babe config, then sometimes build the runtime again and use it to get the GrandPa config.

This PR optimizes this by having each step return the runtime that has been built so that it can be passed to the next step.

Furthermore, the entire `ChainInformation` building process now returns the genesis runtime as well.
This is actually the main reason why I open this PR, as I will need the genesis runtime in order to do https://github.com/paritytech/smoldot/issues/1967.


tomaka:
Merging as I need these changes in order to progress with https://github.com/paritytech/smoldot/issues/1967

PR #2277, opened 2022-05-06
Implement trie version 1

tomaka:
cc https://github.com/paritytech/smoldot/issues/1967

The first version of trie is called version 0. In version 1 of the trie, values that are longer than 32 bytes are hashed. Additionally, the header of each trie element has some additional bit patterns that indicate whether the value is hashed or not.

In order to know whether to use v0 or v1 of the trie, the runtime spec (found in the runtime) has an additional field called `state_version` (which was already parsed by smoldot since https://github.com/paritytech/smoldot/pull/1971 but ignored) indicating the version.

For this reason, it is no longer appropriate to have a standalone `calculate_genesis_block_header` function. Calculating the genesis block header requires knowing information from the genesis runtime, and building the genesis runtime is a very expensive operation. Instead, the genesis block header can now be obtained by higher-level code by building the genesis `ChainInformation`.

Note that technically each individual item of the trie is either v0 or v1. In other words, a trie can be in "hybrid" mode with some items in v0 and some items in v1. This happens if the chain was using v0 in the past and has switched to v1. When a chain switches to v1 (by modifying the `state_version` field), the updates are done lazily when items are written to the storage.
This means that Merkle proofs can also contain a hybrid of v0 and v1 items.

I have not implemented the full-node-related changes, as they are not straight forward and require some changes to the database. At the moment, the full node will error if you use it on a chain with a v1 trie or try to switch to v1. The places that need an update have been appropriately marked with `TODO`s.

### DIFF
--- a/src/executor/host.rs
+++ b/src/executor/host.rs
@@ -791,8 +791,8 @@ impl ReadyToRun {
         macro_rules! expect_state_version {
             ($num:expr) => {{
                 match &params[$num] {
-                    vm::WasmValue::I32(0) => 0,
-                    vm::WasmValue::I32(1) => 1,
+                    vm::WasmValue::I32(0) => trie::TrieEntryVersion::V0,
+                    vm::WasmValue::I32(1) => trie::TrieEntryVersion::V1,
                     v => {
                         return HostVm::Error {
                             error: Error::WrongParamTy {
@@ -932,9 +932,10 @@ impl ReadyToRun {
             HostFunction::ext_storage_root_version_2 => {
                 let state_version = expect_state_version!(0);
                 match state_version {
-                    0 => HostVm::ExternalStorageRoot(ExternalStorageRoot { inner: self.inner }),
-                    1 => host_fn_not_implemented!(), // TODO: https://github.com/paritytech/smoldot/issues/1967
-                    _ => unreachable!(),
+                    trie::TrieEntryVersion::V0 => {
+                        HostVm::ExternalStorageRoot(ExternalStorageRoot { inner: self.inner })
+                    }
+                    trie::TrieEntryVersion::V1 => host_fn_not_implemented!(), // TODO: https://github.com/paritytech/smoldot/issues/1967
                 }
             }
             HostFunction::ext_storage_changes_root_version_1 => {
@@ -1481,13 +1482,12 @@ impl ReadyToRun {
             HostFunction::ext_sandbox_get_global_val_version_1 => host_fn_not_implemented!(),
             HostFunction::ext_trie_blake2_256_root_version_1
             | HostFunction::ext_trie_blake2_256_root_version_2 => {
-                if matches!(host_fn, HostFunction::ext_trie_blake2_256_root_version_2) {
-                    match expect_state_version!(1) {
-                        0 => {}
-                        1 => host_fn_not_implemented!(), // TODO: https://github.com/paritytech/smoldot/issues/1967
-                        _ => unreachable!(),
-                    }
-                }
+                let state_version =
+                    if matches!(host_fn, HostFunction::ext_trie_blake2_256_root_version_2) {
+                        expect_state_version!(1)
+                    } else {
+                        trie::TrieEntryVersion::V0
+                    };
 
                 let result = {
                     let input = expect_pointer_size!(0);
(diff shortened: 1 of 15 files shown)


src/trie/proof_verify.rs
```diff
@@ -219,7 +219,11 @@ pub fn trie_node_info<'a, 'b>(
         } else {
             // The current node (as per `proof_iter`) exactly matches the requested key.
             return Ok(TrieNodeInfo {
-                storage_value: decoded_node_value.storage_value,
+                storage_value: match decoded_node_value.storage_value {
+                    proof_node_decode::StorageValue::Hashed(_) => todo!(), // TODO: /!\
```
tomaka, line 223: This should be finished before the PR is ready.

src/trie/node_value.rs
```diff
@@ -198,13 +213,24 @@ where
...
+            let stored_value = config.stored_value.unwrap();
+
             // Doing something like `merkle_value_sink.update(stored_value.encode());` would be
             // quite expensive because we would duplicate the storage value. Instead, we do the
             // encoding manually by pushing the length then the value.
-            merkle_value_sink
-                .update(util::encode_scale_compact_usize(stored_value.as_ref().len()).as_ref());
-            merkle_value_sink.update(stored_value.as_ref());
+            if !hash_stored_value {
+                merkle_value_sink
```
melekes, line 223: this can be moved into `else` block below

src/trie/node_value.rs
```diff
@@ -222,13 +248,23 @@ where
...
+        let stored_value = config.stored_value.unwrap();
+
         // Doing something like `merkle_value_sink.update(stored_value.encode());` would be
         // quite expensive because we would duplicate the storage value. Instead, we do the
         // encoding manually by pushing the length then the value.
-        merkle_value_sink
-            .update(util::encode_scale_compact_usize(stored_value.as_ref().len()).as_ref());
-        merkle_value_sink.update(stored_value.as_ref());
+        if !hash_stored_value {
+            merkle_value_sink
```
melekes, line 258: same

src/trie/proof_node_decode.rs
```diff
@@ -19,19 +19,36 @@ use super::nibble;
...
 pub fn decode(mut node_value: &[u8]) -> Result<Decoded, Error> {
     if node_value.is_empty() {
         return Err(Error::Empty);
     }
 
-    let has_children = (node_value[0] & 0x80) != 0;
-    let has_storage_value = (node_value[0] & 0x40) != 0;
+    let (has_children, storage_value_hashed, pk_len_first_byte_bits) = match node_value[0] >> 6 {
+        0b00 => {
+            if (node_value[0] >> 5) == 0b001 {
```
melekes, line 31: why not `match`?

src/trie/proof_node_decode.rs
```diff
@@ -19,19 +19,36 @@ use super::nibble;
 use core::{fmt, iter, slice};
 
 /// Decodes a node value found in a proof into its components.
+///
+/// This can decode nodes no matter their version.
 pub fn decode(mut node_value: &[u8]) -> Result<Decoded, Error> {
```
melekes, line 24: would be good to link the spec here too

melekes (commented):
👍

src/trie/proof_node_decode.rs
tomaka: You'd need to write something like:

```
match node_value[0] {
    v if (v >> 5) == 0b001 => ...
    v if (v >> 4) == 0b0001 => ...
    0 => ...
    _ => 
}
```

And I just find the ifs more readable

src/executor/host.rs
```diff
@@ -1481,13 +1482,12 @@ impl ReadyToRun {
...
             | HostFunction::ext_trie_blake2_256_root_version_2 => {
-                if matches!(host_fn, HostFunction::ext_trie_blake2_256_root_version_2) {
-                    match expect_state_version!(1) {
-                        0 => {}
-                        1 => host_fn_not_implemented!(), // TODO: https://github.com/paritytech/smoldot/issues/1967
-                        _ => unreachable!(),
-                    }
-                }
+                let state_version =
+                    if matches!(host_fn, HostFunction::ext_trie_blake2_256_root_version_2) {
```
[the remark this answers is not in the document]
tomaka: This is a standalone function that builds a trie from the function parameter and returns the value.

It's not the "root_version" that is version 2, it's the function as a whole that is version 2, and the function is called "ext_trie_<hashalgo>_root"

`ext_trie_blake2_256_root_version_2` can be passed a `state_version` to use for the calculation, while `ext_trie_blake2_256_root_version_1` can't and implicitly assumes state_version = 0.

melekes (approved):
💪 🏋 🥇

vivekvpandya:
> @tomaka once completely switched to V1 , is it possible to go back by setting state_version to 0?

Any thoughts on this?

Issue #2290, opened 2022-05-17
Error on zombienet setup


bernardoaraujor:
I want to do some tests on a local setup and stumbled on an error.
End goal is to connect to a custom parachain.

I'm using `zombienet-linux`, `polkadot` executable is `v0.9.20` and `parachain-collator` is built from [substrate-parachain-template (polkadot-v0.9.20)](https://github.com/substrate-developer-hub/substrate-parachain-template/releases/tag/polkadot-v0.9.20)

This is zombienet's `toml` file:
```
[settings]
timeout = 1000

[relaychain]
chain = "rococo-local"
default_command = "./polkadot"

  [[relaychain.nodes]]
  name = "alice"
  validator = true
  extra_args = [ "--alice", "-lparachain=debug" ]

  [[relaychain.nodes]]
  name = "bob"
  validator = true
  extra_args = [ "--bob", "-lparachain=debug" ]

[[parachains]]
id = 100
addToGenesis = true
cumulus_based = true

  [parachains.collator]
  name = "collator01"
  command = "./parachain-collator"
  args = ["-lparachain=debug"]
```
After I start the setup, a chainspec is generated, which I copy as `customSpec.json`:
```
$ cp /tmp/zombie-a16cd203076fe4c2a430f70f553d747f_-498293-GE4jWyFwXO2g/rococo-local-100.json customSpec.json 
```

This is my Substrate Connect javascript:
```
import { ScProvider } from '@polkadot/rpc-provider/substrate-connect';
import { ApiPromise } from '@polkadot/api';
import customSpec from './customSpec.json' assert {type: "json"};
// Create the provider for the custom chain
const provider = new ScProvider(customSpec);
// Stablish the connection (and catch possible errors)
await provider.connect()
// Create the PolkadotJS api instance
const api = await ApiPromise.create({ provider });
await api.rpc.chain.subscribeNewHeads((lastHeader) => {
console.log(lastHeader.hash);
});
await api.disconnect();
```

And finally, this is the error message I run into:
```
$ node subconn.js 
(node:499792) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
[smoldot] Smoldot v0.6.15
Smoldot has panicked. This is a bug in smoldot. Please open an issue at https://github.com/paritytech/smoldot/issues with the following message:
[...]

PR #2291, opened 2022-05-17
Improve error message when passing non-string as chain spec

tomaka:
cc https://github.com/paritytech/smoldot/issues/2290


melekes (approved):
👍

Issue #2295, opened 2022-05-17
Panic in Peers::start_request


tomaka:
https://github.com/paritytech/smoldot/blob/e7768f532440a1d7a51e23e51ffa1ee42bed24b6/src/libp2p/peers.rs#L1154

```
Error: panicked at 'called `Option::unwrap()` on a `None` value', /home/tomaka/Projets/smoldot/src/libp2p/peers.rs:1154:72
(demo) Error during JSON-RPC request: Error: Error: panicked at 'called `Option::unwrap()` on a `None` value', /home/tomaka/Projets/smoldot/src/libp2p/peers.rs:1154:72
```

Happened while I was testing https://github.com/paritytech/smoldot/issues/1475, roughly when my Internet connection came back

PR #2297, opened 2022-05-17
Make sure we're connected before starting requests

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2295

After https://github.com/paritytech/smoldot/pull/2264, we now panic if starting a request on a peer we're not connected to because this is an invalid usage of the API.
The reason why this wouldn't panic before is because there was always a chance of a race condition where a connection started being closed in parallel of the request being started.
We simply do the check for connectivity in the network service, as it is now possible to do so.

### DIFF
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -1180,8 +1180,7 @@ where
 
         self.connections_by_peer
             .range(
-                (peer_index, ConnectionId::min_value())
-                    ..=(peer_index, ConnectionId::max_value()),
+                (peer_index, ConnectionId::min_value())..=(peer_index, ConnectionId::max_value()),
             )
             .any(|(_, established)| *established)
     }
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -1172,8 +1172,8 @@ where
     }
 
     /// Returns `true` if there exists an established connection with the given peer.
-    pub fn has_established_connection(&self, peer: &PeerId) -> bool {
-        let peer_index = match self.peer_indices.get(peer) {
+    pub fn has_established_connection(&self, peer_id: &PeerId) -> bool {
+        let peer_index = match self.peer_indices.get(peer_id) {
             Some(idx) => *idx,
             None => return false,
         };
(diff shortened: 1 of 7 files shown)


src/libp2p/peers.rs
```diff
@@ -1171,6 +1171,20 @@ where
         self.inner.respond_in_request(id.0, response)
     }
 
+    /// Returns `true` if there exists an established connection with the given peer.
+    pub fn has_established_connection(&self, peer: &PeerId) -> bool {
```
melekes, line 1175:
```suggestion
    pub fn has_established_connection(&self, peer_id: &PeerId) -> bool {
```

src/libp2p/peers.rs
```diff
@@ -1171,6 +1171,20 @@ where
...
+        let peer_index = match self.peer_indices.get(peer) {
+            Some(idx) => *idx,
+            None => return false,
+        };
+
+        self.connections_by_peer
```
melekes, line 1181: why would we want to have multiple connections to the same peer?

src/network/service.rs
```diff
@@ -1919,6 +1919,11 @@ where
         None
     }
 
+    /// Returns `true` if there exists an established connection with the given peer.
+    pub fn has_established_connection(&self, peer: &PeerId) -> bool {
```
melekes, line 1923:
```suggestion
    pub fn has_established_connection(&self, peer_id: &PeerId) -> bool {
```

src/libp2p/peers.rs
tomaka: This can happen accidentally if two peers connect to each other at the same time.

Another situation is if you connect to an IP address and expect to find peer A but actually connect to peer B, even though you're already connected to B.
In that case, from the point of view of peer B, maybe you have accidentally connected to it (the example that I just gave), but maybe the first connection you had with B is actually dead and you're reopening a new one.

The best way to handle these situations is to allow multiple connections between peers, and automatically close unused ones.

melekes (approved):
👍

PR #2298, opened 2022-05-18
Remove redundant field in service.rs

tomaka:
Before https://github.com/paritytech/smoldot/pull/2264, we were maintaining an additional redundant list of connections in `service.rs` in order to avoid having to lock a mutex.
After the big refactor, this redundant list is no longer needed as we can directly access the list in the underlying data structure.


melekes (approved):
👍

PR #2299, opened 2022-05-19
Fix a couple of panics in new networking

tomaka:
I have bundled 5 networking bug fixes in 5 commits in this single PR.

These bugfixes have made me realize that some implementation details need to be changed quite a bit to adjust to the new design, which I'll do in further PRs.


tomaka:
Merging as I really need this to continue work.

PR #2300, opened 2022-05-19
Make notification_substream_queued_bytes take a &self

tomaka:


src/libp2p/connection/yamux.rs
```diff
@@ -895,6 +904,48 @@ pub struct Config {
...
+        self.id
+    }
+
+    /// Returns the user data associated to this substream.
+    pub fn user_data(&self) -> &T {
+        &self.substream.user_data
+    }
+
+    /// Returns the user data associated to this substream.
+    pub fn into_user_data(self) -> &'a T {
```
melekes, line 926: _don't see much difference between this and above fn except `'a` lifetime._

melekes (approved):
👍

src/libp2p/connection/yamux.rs
tomaka: Yeah but they're fundamentally different.

These are methods on an object that borrows the connection.
`user_data(&self) -> &T` returns a `&T` that borrows the object that borrows the connection, while `into_user_data(self) -> &'a T` kills the object and returns a `&T` that borrows the connection.

melekes: Oh, right. Thanks

PR #2306, opened 2022-05-20
Don't store in peers.rs whether connections are established

tomaka:
Replaces a `BTreeMap` with a `BTreeSet`. The value that was in the `BTreeMap` is now queried from the underlying layer.

I've also added a TODO because this change made me realize a problem with the code that would be off-topic for this PR.

### DIFF
--- a/src/libp2p/collection.rs
+++ b/src/libp2p/collection.rs
@@ -397,6 +397,21 @@ where
         self.connections.len()
     }
 
+    /// Returns the state of the given connection.
+    ///
+    /// # Panic
+    ///
+    /// Panics if the identifier is invalid or corresponds to a connection that has already
+    /// entirely shut down.
+    ///
+    pub fn connection_state(&self, connection_id: ConnectionId) -> ConnectionState {
+        let connection = self.connections.get(&connection_id).unwrap();
+        ConnectionState {
+            established: !connection.handshaking,
+            shutting_down: connection.shutting_down,
+        }
+    }
+
     /// Returns the Noise key originalled passed as [`Config::noise_key`].
     pub fn noise_key(&self) -> &NoiseKey {
         &self.noise_key
@@ -906,6 +921,7 @@ where
                 }
                 ConnectionToCoordinatorInner::ShutdownFinished => {
                     debug_assert!(connection.shutting_down);
+                    let was_established = !connection.handshaking;
                     let user_data = self.connections.remove(&connection_id).unwrap().user_data;
                     self.messages_to_connections.push_back((
                         connection_id,
@@ -913,6 +929,7 @@ where
                     ));
                     Event::Shutdown {
                         id: connection_id,
+                        was_established,
                         user_data,
                     }
                 }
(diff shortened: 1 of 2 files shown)


src/libp2p/peers.rs
```diff
@@ -328,7 +323,12 @@ where
                                 (actual_peer_index, collection::ConnectionId::min_value())
                                     ..=(actual_peer_index, collection::ConnectionId::max_value()),
                             )
-                            .filter(|(_, established)| **established)
+                            .filter(|(_, connection_id)| {
+                                // Note that connections that are shutting down are still counted,
+                                // as we report the disconnected event only at the end of the
+                                // shutdown.
+                                self.inner.connection_state(*connection_id).established
```
melekes, line 330: is it possible that this code panics when connection existed during range, but was shutdown before filter => panic?

src/libp2p/collection.rs
```diff
@@ -397,6 +397,21 @@ where
...
 
+    /// Returns the state of the given connection.
+    ///
+    /// # Panic
+    ///
+    /// Panics if the identifier is invalid or corresponds to a connection that has already
+    /// entirely shut down.
+    ///
+    pub fn connection_state(&self, connection_id: ConnectionId) -> ConnectionState {
+        let connection = self.connections.get(&connection_id).unwrap();
```
melekes, line 408: in my experience it's better to return an error or Option rather than panic.

src/libp2p/peers.rs
tomaka: No, the local list of connections is always strictly in sync with the underlying layer.

src/libp2p/collection.rs
tomaka: I know many people say that, but in my opinion it's not.

If you take for example the `queue_notification` function, which can fail because the queue is full, and imagine that instead the function returns an `Err` either if the queue is full or if the `ConnectionId` is wrong.

If you call `queue_notification` and get an error, what do you do? The queue being full is a legitimate problem that you can handle by buffering the notification to send it later for example. But the `ConnectionId` being invalid can't be handled properly. The only reason why you would pass an invalid `ConnectionId` is a bug in the code, and in that situation the right thing to do is to panic.

Now, in principle I agree that it's the code that calls `queue_notification` that should match the error and handle some kind of `ConnectionIdInvalid` error and call `panic!()`, but in practice this is extremely tedious. If you specifically filter out the `ConnectionIdInvalid` error, then you have to redefine another error type that contains all the variants except for `ConnectionIdInvalid`. In practice what happens is that programmers don't make the effort or simply don't think about it and propagate the `ConnectionIdInvalid` error to upper layers of the code, which is non-sensical.
Basically, it is much easier for everyone if it's `queue_notification` that panics if the `ConnectionId` is invalid.

My implicit rule is that if the only reason why an error can happen is a bug in the code, then there's no reason to properly return an error and panicking is completely appropriate.

melekes: Sorry if my comment was too generic. I agree that in cases of bugs we should panic. But in cases such as queue being full we should instead provide back-pressure or drop incoming messages and/or figure out if we can speed up processing (whatever is more appropriate in a particular case). That being said, I think we're on the same page generally.

PR #2308, opened 2022-05-23
Decode the block number directly into a `u64` into of going through a `usize`

tomaka:
Fixes a TODO in `header.rs`

Can be reviewed commit by commit


melekes (approved):
👍

PR #2309, opened 2022-05-23
Remove obsolete TODO in chain_information

tomaka:
The TODO refers to something that is already done. See lines 379-381 and 395-399.


melekes (approved):
👍

PR #2310, opened 2022-05-23
Remove a TODO in finalized serialized decoding code

tomaka:


melekes (approved):
👍

melekes:
Does the db we're using checks the data integrity itself? like adding checksums and checking them after reading data from disk?

tomaka:
This code just encodes/decodes the state into JSON, it doesn't do any disk reading/writing.
This JSON is then saved into the browser's local storage.

PR #2319, opened 2022-05-27
Use proper randomness for outgoing pings

tomaka:
When sending an outgoing ping, we're supposed to send a random payload. Before this PR, this payload was hardcoded to `[0xff; 32]`. This fixes it.


melekes (approved):
👍

PR #2321, opened 2022-05-27
Fix a panic if Noise frame length is too low

tomaka:
The value in `expected_len` is read from an incoming networking packet. The networking packet is invalid if `expected_len < 16`, however it is unwise to do `expected_len - 16` as a malicious peer might send an invalid value on purpose.

Rust only detects underflow errors in debug mode. However, in release mode this will lead to a very high value being passed to `resize`, which will lead to an allocation error.

This bug is technically a vulnerability, as anyone can crash a smoldot light client if it connects to their node. However, since we're not widely used at the moment and that an attacker has nothing to gain except being a nuisance, I'm going to assume that nobody is going to exploit this and skip the non-disclosure period.

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Fix a panic in case of a Noise message with an invalid length. ([#2219](https://github.com/paritytech/smoldot/pull/2219))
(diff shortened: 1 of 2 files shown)


tomaka:
Note that this changes the order of operations. Before this PR, we add `len_before` and `expected_len`, then subtract 16. After this PR, we first subtract 16. This is intentional and not very important.

melekes (approved):
👍

PR #2322, opened 2022-05-27
Fix a debug assertion in Noise handshake when expected_len == 0

tomaka:
This is similar to https://github.com/paritytech/smoldot/pull/2321, but since it's only a `debug_assert!` it's not a big deal.


melekes (approved):
👍

PR #2323, opened 2022-05-27
Add tests to established connections

tomaka:
Adds a test to the established connections.

I initially just wanted to write regression tests, unfortunately the first test I wrote doesn't pass because of [this](https://github.com/paritytech/smoldot/blob/8fceefb18f8936e4324b640e0e58caeb132d8ef7/src/libp2p/connection/established.rs#L447-L462) generally messy piece of code.
I'm going to fix that generally messy piece of code before we can merge this PR.

---

Update:

Working on the tests revealed a couple of bugs that I've fixed.

Additionally, I've also made a change where we now pass a timeout to `open_notifications_substream` to allow configuring it and to make it similar to `add_request`.


tomaka:
Tests are now passing.

tomaka:
This PR is definitely a bit embarrassing: it revealed many unimplemented things, which now are.

melekes (approved):

PR #2327, opened 2022-05-30
Re-enable incremental compilation

tomaka:
I had disabled it [9 months ago](https://github.com/paritytech/smoldot/pull/1311) due to bugs.

In [the blog post of Rust 1.60](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#incremental-compilation-status) (note that we're at 1.61 right now), it is indicated that incremental compilation should work much better now, so let's re-enable it for debug mode.

It does indeed considerably speed up compilation times.


melekes (approved):
👍

PR #2337, opened 2022-06-06
Fix LEB128-related panic again

tomaka:
Similar to https://github.com/paritytech/smoldot/pull/2326#issuecomment-1141318206

We actually have two LEB128 decoding codes: one when the data is complete and one when the data is maybe incomplete, and I forgot about this.


melekes (approved):
👍

PR #2340, opened 2022-06-06
Fix #2339

### LINKED ISSUE #2339: Panic in decode_block_request
Repro: `[26, 10]`

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2339

It turns out that `finish` panics if the data is incomplete (https://docs.rs/nom/latest/nom/trait.Finish.html#tymethod.finish)
In order to fix that, I've added the `complete` combinator everywhere we use `finish`, making sure that `Incomplete` errors are turned into actual errors.


melekes (approved):
👍

PR #2342, opened 2022-06-06
Add fuzzing targets

tomaka:
This PR adds fuzzing binaries to the smoldot repo.
I've added one for every fuzzable thing that I could think of.

I've only let them run maybe 15 minutes each, so there's room for running things over night to detect crashes.

Many of them are failing because of https://github.com/paritytech/smoldot/pull/2340

The `network-connection-encrypted` fuzzing test is failing a lot, but that's good, and I'll have to fix all the issues that arise.


melekes (approved):
🚀 🌔

PR #2344, opened 2022-06-06
Optimize range() usage in storage_diff and consensus service

tomaka:
Removes a few calls to `to_vec()` by passing a different trait implementation to `range()`.


melekes (approved):
👍

PR #2345, opened 2022-06-06
Properly log the block requests in the full node

tomaka:


bin/full-node/src/run/network_service.rs
```diff
@@ -586,13 +615,39 @@ impl NetworkService {
...
             rx
         };
 
-        rx.await.unwrap().map_err(BlocksRequestError::Request)
+        let result = rx.await.unwrap().map_err(BlocksRequestError::Request);
+
+        // Requet has finished. Print the log and prevent the cancellation message from being
+        // printed.
+        mem::forget(_log_if_cancel);
+        // TODO: print something if the request is cancelled?
```
melekes, line 630: but we've just dropped the cancellation struct, why the need for todo?

tomaka: Mistake

melekes (approved):
👍

PR #2349, opened 2022-06-06
Better Display impl for ProtocolError

tomaka:
The derivation with `derive_more::Display` completely erases the description when there is a single field. This is a bit annoying.


melekes (approved):
👍

PR #2351, opened 2022-06-07
Properly cache the fuzzing artifacts

tomaka:
Right now the action caches the `./target` at the root, but that's not where the artifacts that are relevant to us are located.


tomaka:
Merging as insubstantial.

PR #2354, opened 2022-06-08
Improve Display implementation of most errors

tomaka:
As mentioned in https://github.com/paritytech/smoldot/pull/2349, in situations where there's a single field, the `Display` trait automatic derivation automatically delegates the displaying to it.

This is sometimes good, but sometimes the inner error is missing a lot of context. For example if you make a runtime call, you might get an error saying "Data size exceeds allowed maximum". It's missing the fact that this is while decoding the response of a storage request on the network.

I've added proper context to all error messages.
Sometimes I've deliberately put `#[display(fmt = "{}", _0)]`, which simply prints the single field. This is to make explicit the fact that this is what we want.


bin/full-node/src/run/network_service.rs
```diff
@@ -665,6 +665,10 @@ pub enum InitError {
     #[display(fmt = "I/O error when creating listener for {}: {}", _0, _1)]
     ListenerIo(Multiaddr, io::Error),
     /// A listening address passed through the configuration isn't valid.
+    #[display(
```
melekes, line 668: [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) only applies to errors in opposite to `derive_more` which is more general and provides more functionality, but they are more or less the same thing, right?

melekes (approved):
👍

bin/full-node/src/run/network_service.rs
tomaka: Yes. `thiserror` doesn't support no-std platforms though.

PR #2355, opened 2022-06-08
Add more missing calls to nom::combinator::complete

tomaka:
Similar to https://github.com/paritytech/smoldot/pull/2340
There's another code pattern where I don't call `finish` but I manually match the error and use `unreachable!()`. This PR fixes it as well.


melekes (approved):
👍

PR #2356, opened 2022-06-08
Pin the nightly version in fuzzing-binaries-compile

tomaka:
The caching still doesn't work, this time because of the Rust compiler changing version every day.


melekes (approved):
👍

PR #2357, opened 2022-06-08
Detect infinite loops in protobuf decoding

tomaka:
I didn't encounter any infinite loop, however I think it's a good idea to detect them.
The `nom` library (that we use) does the same for its own low-level primitives.


melekes (approved):
👍

PR #2363, opened 2022-06-09
Add Acala and Karura to the chain specs list

tomaka:
Just like the other chain specs in this repo, it's just to make it easier to test.


melekes (approved):
👍

PR #2364, opened 2022-06-09
Improve full node error message when JSON-RPC port is already busy

tomaka:
Right now the error just says "Addr already occupied" without any context, which is a bit hard to understand.


melekes (approved):
👍

PR #2365, opened 2022-06-09
Add experimental warning at full node initialization

tomaka:


bin/full-node/src/run.rs
```diff
@@ -103,6 +103,13 @@ pub async fn run(cli_options: cli::CliOptionsRun) {
             .expect("Failed to decode chain specs")
     };
 
+    // This warning message should be removed if the full node becomes mature.
```
melekes, line 106:
```suggestion
    // This warning message should be removed if/when the full node becomes mature.
```

?)

melekes (approved):
👍

PR #2367, opened 2022-06-09
Update README content regarding Wasm demo

tomaka:
This small text was written when the demo only connected to Westend.


melekes (approved):
👍

Issue #2362, opened 2022-06-09
panic when trying to sync with Acala


xlc:
```
$ RUST_BACKTRACE=1 cargo run -- run --chain=./bin/acala-dist.json
    Finished dev [optimized + debuginfo] target(s) in 0.24s
     Running `target/debug/full-node run --chain=./bin/acala-dist.json`
path : "./bin/polkadot.json"
2022-06-09T03:40:58.216213+00:00  INFO full_node::run: successful-initialization local_peer_id=12D3KooWDrz9ZxfAfEb6owhmtkk4ak11TXAHa3LAnmrQ3yq6w1wu database_is_new=true finalized_block_hash=0xfc41…a64c finalized_block_number=0
2022-06-09T03:41:07.211219+00:00  WARN full_node::run::consensus_service: failed-block-verification hash=0x2543…1ebb height=3 error=Error while executing Wasm VM: Trap: MemoryAccessOutOfBounds                                                                                                                                    ] #1190311 (🔗 24) (🌐  24)   
""
2022-06-09T03:41:07.880869+00:00  WARN full_node::run::consensus_service: failed-block-verification hash=0x417d…efa9 height=1 error=Error while executing Wasm VM: Trap: ElemUninitialized                                                                                                                                          ] #1190312 (🔗 24) (🌐  24)   
""
2022-06-09T03:41:08.172604+00:00  WARN full_node::run::consensus_service: failed-block-verification hash=0x417d…efa9 height=1 error=Error while executing Wasm VM: Trap: ElemUninitialized                                                                                                                                          ] #1190312 (🔗 24) (🌐  24)   
""
thread 'tasks-pool-3' panicked at 'called `Option::unwrap()` on a `None` value', /Users/xiliangchen/projects/smoldot/src/sync/optimistic.rs:598:18                                                                                                                                                                                  ] #1190312 (🔗 26) (🌐  26)   
stack backtrace:
   0: rust_begin_unwind
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
[...]

PR #2368, opened 2022-06-09
Fix OptimisticSync::remove_source not removing obsolete requests

tomaka:
Fixes the first panic of https://github.com/paritytech/smoldot/issues/2362

When we disconnect from a so-called "source" (i.e. a peer), `remove_source` returns the list of all on-going requests so that the API user knows that these requests no longer exist in the syncing state machine.

However, before this PR `remove_source` didn't return the requests whose outcome no longer interests us. This leads to a state mismatch.
Later, when the request gets answered, this triggers a panic because the state machine doesn't expect the request to be there.

This was marked by a TODO, and this PR fixes this TODO.


melekes (approved):
👍

Issue #2361, opened 2022-06-09
thread 'tasks-pool-3' panicked at 'called `Option::unwrap()` on a `None` value', 


xlc:
```
 $ RUST_BACKTRACE=1 cargo run -- run
    Finished dev [optimized + debuginfo] target(s) in 0.25s
     Running `target/debug/full-node run`
2022-06-09T03:36:12.139134+00:00  INFO full_node::run: successful-initialization local_peer_id=12D3KooWQ7sbPyxMfoMXpp8YjA6oEt3dDqfKAjzmp2SXYpU6YvhF database_is_new=false finalized_block_hash=0x91b1…90c3 finalized_block_number=0
    Polkadot   #0 [>                                                                                                                                                                                                                                                                                                                           ] ? (🔗 16) (thread 'tasks-pool-4' panicked at 'called `Option::unwrap()` on a `None` value', /Users/xiliangchen/projects/smoldot/src/libp2p/peers.rs:1178:72                                                                                                                                                                                               ] ? (🔗 16) (🌐  16)   
stack backtrace:
   0: rust_begin_unwind
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/b17226fcc11587fed612631be372a5b4cb89988a/library/core/src/option.rs:752:21
   4: smoldot::libp2p::peers::Peers<TConn,TNow>::start_request
             at ./src/libp2p/peers.rs:1178:36
   5: smoldot::network::service::ChainNetwork<TNow>::start_kademlia_find_node_inner
             at ./src/network/service.rs:1821:18
   6: smoldot::network::service::ChainNetwork<TNow>::start_kademlia_discovery_round
             at ./src/network/service.rs:1772:17
   7: full_node::run::network_service::NetworkService::new::{{closure}}::{{closure}}
             at ./bin/full-node/src/run/network_service.rs:323:44
   8: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
[...]

PR #2369, opened 2022-06-10
Add fake out requests system in peers.rs

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2361

This PR solves a tricky problem in `peers.rs`.

A connection can be in three phases: handshaking, established, or shutting down (and, later, completely shut down, but that doesn't count as a phase because when the connection is shut down it no longer exists).
Once a connection starts its shutdown phase, it is forbidden to start new requests on this connection, but incoming notifications can still arrive.

This is not a problem in `connection.rs` because we simply indicate when a connection starts shutting down, and the upper layers of the code are aware of the fact that they shouldn't start any new request but can still receiving incoming notifications.

But in `peers.rs` this is problematic, because `peers.rs` groups together multiple connections belonging to the same peer and exposes them as just one. It is for example possible to have one connection to a peer shutting down but another connection to the same peer is still established. For this reason, `peers.rs` doesn't report shut downs, and instead simply says "we are connected to a peer" or "we are no longer connected to a peer".

But this causes another problem: if you have only one connection left and it is shutting down, should you still tell the upper API layers that you're connected? The answer is yes, otherwise it would be weird potentially receive incoming notifications from peers you are not connected to.
But if the upper API layers think that you are connected to a peer despite having only one connection that is shutting down, what happens if these upper API layers start new requests? This is the problem that this PR solves.

This PR introduces a fake requests system in `peers.rs`. When a request is started and all the connections to that peer are shutting down, we create a fake request and later report as event that the request has failed.

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Fix another panic in case of a carefully-crafted LEB128 length. ([#2337](https://github.com/paritytech/smoldot/pull/2337))
 - Fix a panic when decoding a block header containing a large number of Aura authorities. ([#2338](https://github.com/paritytech/smoldot/pull/2338))
 - Fix multiple panics when decoding network messages in case where these messages were truncated. ([#2340](https://github.com/paritytech/smoldot/pull/2340), [#2355](https://github.com/paritytech/smoldot/pull/2355))
+- Fix panic when the Kademlia random discovery process initiates a request on a connection that has just started shutting down. ([#2369](https://github.com/paritytech/smoldot/pull/2369))
 
 ## 0.6.17 - 2022-05-31
 
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -150,6 +150,13 @@ pub struct Peers<TConn, TNow> {
     /// function.
     // TODO: explain why it can't grow unbounded
     pending_desired_out_notifs: VecDeque<(DesiredOutNotificationId, usize, usize)>,
+
+    /// List of so-called "fake outgoing requests" for which an event about their failure should
+    /// be reported as soon as possible.
+    fake_out_requests_to_report: VecDeque<u64>,
+
+    /// Identifier to assign to the next fake outgoing requests. Increased one by one.
+    next_fake_out_request_id: u64,
 }
 
 /// See [`Peers::peers_notifications_out`].
@@ -183,6 +190,14 @@ struct Connection<TConn> {
     /// - If the handshake is in progress and the connection is inbound, contains `None`.
     peer_index: Option<usize>,
 
+    /// List of so-called fake outgoing requests that [`Peers::start_request`] has created
+    /// regarding this connection. Always empty unless the connection is currently shutting down.
+    /// The API user thinks that the connection is currently executing this request, but in
+    /// reality it is not. When the connection finishes its shutdown, these fake requests are
+    /// pushed to [`Peers::fake_out_requests_to_report`] in order to be reported as having failed.
+    fake_out_requests: Vec<u64>,
+
+    /// Opaque data decided by the API user.
     user_data: TConn,
 }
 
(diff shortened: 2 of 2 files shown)


tomaka:
I've also opened https://github.com/paritytech/smoldot/issues/2370 which is an alternative approach.

bin/wasm-node/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Fix another panic in case of a carefully-crafted LEB128 length. ([#2337](https://github.com/paritytech/smoldot/pull/2337))
 - Fix a panic when decoding a block header containing a large number of Aura authorities. ([#2338](https://github.com/paritytech/smoldot/pull/2338))
 - Fix multiple panics when decoding network messages in case where these messages were truncated. ([#2340](https://github.com/paritytech/smoldot/pull/2340), [#2355](https://github.com/paritytech/smoldot/pull/2355))
+- Fix panic when the Kademlia random discovery process initiates a request on a connection that has just started shutting down. ([#2369](https://github.com/paritytech/smoldot/pull/2369))
```
tomaka, line 14: I'm mentioning Kademlia because I've only ever seen this problem happen with the Kademlia discovery. In theory it can happen with other protocols, but in practice it doesn't seem to, probably because we accidentally conveniently never start any other kind of requests during shut down.

melekes:
> if you have only one connection left and it is shutting down, should you still tell the upper API layers that you're connected? The answer is yes, otherwise it would be weird potentially receive incoming notifications from peers you are not connected to.

It's not immediately obvious to me why the answer is yes. If we've reported that connection is shutting down in `peers.rs` and upper layers are aware, then it's okay for them to receive notifications (same as in `connection.rs`).

> But in peers.rs this is problematic, because peers.rs groups together multiple connections belonging to the same peer and exposes them as just one.

I think we've talked about this before a little bit, but this bug again is a "reminder" that the underlying design is too complex and buggy, and it would be far simpler to have a single connection to a peer instead of `peers.rs` pretending to have a single connection when in fact there are many.

tomaka:
> If we've reported that connection is shutting down in peers.rs and upper layers are aware, then it's okay for them to receive notifications (same as in connection.rs).

Well, that's https://github.com/paritytech/smoldot/issues/2370

> it would be far simpler to have a single connection to a peer

This opens us to many issues and even an attack:

- Node A and B are connected to each other.
- Node C then tells node A that there exists a node D with a certain IP address.
- Node A tries to connect to node D. It turns out that this IP address connects to node B, and this hypothetical node D is actually B. Node A finds itself with two connections to node B.

So C managed to force A to connect twice to B.

If add some code that closes one connection when there are two, the question is which connection do you close? The older one or the newer one?
Closing the older one opens the door for an attack where C can force-disconnect A and B.
Closing the newer one means that if node A has a sudden disconnect and changes its IP address, then until node B realizes that the connection is unresponsive (which will take ~40 seconds) node A will be unable to reconnect.

See also https://github.com/paritytech/substrate/issues/4272#issuecomment-573796930

melekes:
> It turns out that this IP address connects to node B, and this hypothetical node D is actually B. Node A finds itself with two connections to node B.

Can't `A` drop the second connection once it finds out that the IP provided by `C` == `B`'s IP?

melekes:
> See also https://github.com/paritytech/substrate/issues/4272#issuecomment-573796930

Thanks for the link! Didn't know there were some issues with `unique` connection before. It may be that multiple connections are lesser evil, I was just pointing out that when bugs like https://github.com/paritytech/smoldot/issues/2361  start to appear it could be the sign of improper design (i.e. bugs that force you to lay additional abstractions like fake requests so the code work).

tomaka:
> That's what peer IDs are for, correct? And the same process (C relaying data) can't be performed with peer IDs because the B's certificate contains B's IP, right?

I don't know which certificates you're mentioning.

The original problem is that C can force A to connect to B. If A was already connected to B, then C will have forced A to open a second connection.
There's also no way for A to have the guarantee in advance that it will not connect to a peer it is already connected to.

PR #2371, opened 2022-06-10
Add experimental WebRTC support to multiaddr

tomaka:
cc @mxinden if you're interested

The payload of `/certhash` isn't documented in https://github.com/libp2p/specs/pull/412.
I suspect that ultimately people will want to make `/certhash` super generic and future-proof and over-engineered and all (in case in the vague future it might eventually be used for something else, even though we all know that this will likely never happen), but for now I went with a fixed 32 bytes payload that is hexadecimal-encoded without the `0x` prefix. The certificate hash is put as hexadecimal in the SDP packet, so an hexadecimal payload will make it a bit easier to debug if things work.

This is all marked as `experimental-` until we finish implementing it, so that there's no possible confusion with versions that are incompatible with each other and that kind of things.

### DIFF
--- a/src/libp2p/multiaddr.rs
+++ b/src/libp2p/multiaddr.rs
@@ -207,6 +207,11 @@ pub enum ProtocolRef<'a> {
     Wss,
     // TODO: unclear what the payload is; see https://github.com/multiformats/multiaddr/issues/127
     Memory(u64),
+    // TODO: these protocols are inventions for prototyping, see https://github.com/libp2p/specs/pull/412
+    ExperimentalWebRtc,
+    // TODO: these protocols are inventions for prototyping, see https://github.com/libp2p/specs/pull/412
+    /// Contains the hash of the TLS certificate.
+    ExperimentalCertHash([u8; 32]),
 }
 
 impl<'a> ProtocolRef<'a> {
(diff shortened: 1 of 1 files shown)


melekes (approved):
👍

mxinden:
> I suspect that ultimately people will want to make `/certhash` super generic and future-proof and over-engineered and all (in case in the vague future it might eventually be used for something else, even though we all know that this will likely never happen), but for now I went with a fixed 32 bytes payload that is hexadecimal-encoded without the `0x` prefix. The certificate hash is put as hexadecimal in the SDP packet, so an hexadecimal payload will make it a bit easier to debug if things work.

I would be in favor of being consistent with the way `/certhash` is encoded in the WebTransport specification. The current draft suggests:

> a list of multihashes of the certificates

https://github.com/libp2p/specs/pull/404/

@tomaka would you consider a multihash as "super generic" already? :angel:

tomaka:
A browser client, in order to connect, must know specifically the SHA256 hash of the certificate.
In other words, all multiaddrs of the form `.../webrtc/certhash/<hash>` are undialable unless `<hash>` is SHA256.
Which means that the byte in the multihash that says "this multihash is SHA256" is redundant. And since a SHA256 hash is always 32 bytes, the length is also redundant.
So yes I would consider a multihash too generic.

Issue #2217, opened 2022-04-07
Figure out pinned block limit mechanism in chainHead_unstable_follow


tomaka:
See explanations in this PR: https://github.com/paritytech/smoldot/pull/2215

PR #2373, opened 2022-06-12
Change max number of pinned blocks to max number of finalized pinned blocks

tomaka:
Close https://github.com/paritytech/smoldot/issues/2217

As suggested here: https://github.com/paritytech/smoldot/pull/2215#issuecomment-1091100870

Since the user can't control the number of non-finalized blocks, it makes sense for the non-finalized blocks to not be counted towards the limit. It is entirely smoldot's fault if the number of non-finalized blocks becomes too high (well, it might be because of a problem with the chain itself, but basically it's smoldot's problem and not the JSON-RPC client's problem).

Note that this doesn't require any change to https://github.com/paritytech/json-rpc-interface-spec/, as the number of pinned blocks is at the discretion of the implementation.

### DIFF
--- a/bin/light-base/src/json_rpc_service.rs
+++ b/bin/light-base/src/json_rpc_service.rs
@@ -57,7 +57,7 @@ use smoldot::{
 use std::{
     collections::HashMap,
     iter,
-    num::NonZeroU32,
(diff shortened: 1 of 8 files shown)


melekes (approved):
👍

PR #2376, opened 2022-06-13
DRY in RequestsSubscriptions::add_client

tomaka:
Merges two similar functions `add_client` and `add_client_mut` into one.


melekes (approved):
👍

PR #2378, opened 2022-06-13
Remove obsolete comment in service.rs

tomaka:
This was true because it was possible to call the two mentioned methods simultaneously, which is no longer possible.


melekes (approved):
👍

PR #2386, opened 2022-06-14
Fix panic with unpinning pruned blocks

tomaka:
Fixes a panic introduced in https://github.com/paritytech/smoldot/pull/2373

Imagine the following situation:
Block A has two children: block B and block C. The latest finalized block is currently block A. All three blocks are pinned.

Now, imagine that block B gets finalized. Block A, B, and C all stay pinned. The number of pinned blocks remaining is decreased by 1.

When the user then unpins block C, we check `height(block_C) <= height(current_finalized_block)`, which returns `true`, and we add 1 to the number of pinned blocks remaining.
When the user then unpins block A, the same happens, and we add 1 again to the number of pinned blocks remaining.

This leads to an overflow when that number of pinned blocks is `usize::max_value()`.

This PR fixes this problem by tracking in `pinned_blocks` whether the block counts towards the limit.

### DIFF
--- a/bin/light-base/src/runtime_service.rs
+++ b/bin/light-base/src/runtime_service.rs
@@ -310,8 +310,7 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
             0 | 1
         ));
 
-        all_blocks_subscriptions
-            .insert(subscription_id, (tx, max_pinned_blocks.get() - 1));
+        all_blocks_subscriptions.insert(subscription_id, (tx, max_pinned_blocks.get() - 1));
 
         SubscribeAll {
             finalized_block_scale_encoded_header: finalized_block.scale_encoded_header.clone(),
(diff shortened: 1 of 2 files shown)


bin/light-base/src/runtime_service.rs
```diff
@@ -1489,25 +1486,37 @@ impl<TPlat: Platform> Background<TPlat> {
...
-                                *finalized_pinned_remaining -= 1;
-                            } else {
+                            if sender.try_send(all_blocks_notif.clone()).is_err() {
                                 to_remove.push(*subscription_id);
+                                continue;
+                            }
+
+                            *finalized_pinned_remaining -= count_limit;
+
+                            // Mark the finalized and pruned blocks as finalized or non-canonical.
```
melekes, line 1510:
```suggestion
                            // Mark the finalized and pruned blocks as finalized and non-canonical.
```

?

tomaka: By "non-canonical", I mean "not part of the finalized chain". So by definition "finalized" means "canonical".

melekes (approved):
👍

PR #2388, opened 2022-06-14
Use block in handshake to initialize network_known_best

tomaka:
In the full node, whenever we receive a block announce, we update the value of the `network_known_best` variable that contains the number of the best block seen on the network.

When we connect to a peer, this peer announces what its best block is.

This PR makes it so we also update the value of this variable whenever we connect to a peer.


melekes (approved):
👍

Issue #2392, opened 2022-06-15
Failed to verify warp sync fragment on Rococo


tomaka:
![image](https://user-images.githubusercontent.com/1412254/173781322-a47272cc-f019-4eba-8acb-5fd1aef7f73d.png)

PR #2395, opened 2022-06-15
Update Rococo chain spec

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2392

We did some sudo operation on Rococo recently that overwrites the list of authorities.
This operation is on chain but it breaks the warp syncing algorithm.
There's unfortunately not much we can do here except generate a new chain spec in order to start syncing after the change.


melekes (approved):
👍

PR #2397, opened 2022-06-16
Reduce algorithmic complexity of verifying commits and justifications

tomaka:
Reduces the complexity of checking duplicate public keys by storing the public keys that we have seen before in a hash map rather than iterating again every time.
Considering that on Kusama there are 900 public keys for example, I think that this is a good idea.

I've left TODOs about using `SipHasher`. This will be fixed at the same time as other TODOs in this function by passing a randomness value to the verification process.


melekes (approved):
👍

Issue #515, opened 2021-02-15
Figure out the situation with forced GrandPa authorities changes


tomaka:
A forced GrandPa authorities change consists in forcing a change in the GrandPa authorities. It can only be done with the sudo key.

There are two problems here:

- When a force change occurs, it doesn't get signed by the current authorities. GrandPa warp sync is unfortunately based upon the assumption that each set finalizes the change to the next set. We would have to include more information in the GrandPa warp sync payload, notable the headers of the blocks that lead to the forced change.

- Light clients only verify block headers and not block bodies, meaning that we don't have any indication that a forced change is actually legitimate. A malicious validator could emit an invalid block, and light clients can't detect that it's invalid with only the header. We would have to execute this block in particular, but the storage of the parent might no longer be available on the network.

tomaka:
The pragmatic solution at the moment is to generate a checkpoint from a full node after the forced change.

PR #2401, opened 2022-06-17
Only use database if it's more recent than checkpoint

tomaka:
At a chain initialization, we provide a chain spec and, generally, an existing database. The chain spec can contain a checkpoint where we start syncing from.

Before this PR, the database always took precedence. This means that if database contains block 10 but the chain spec contains block 50000, we'll start syncing at block 10.
After this PR, the database is ignored if its block is below the checkpoint's.

This is especially important in situations where some sudo operation happened on a chain which breaks the warp syncing, like it happened on Rococo a week ago. Basically, smoldot is incapable of syncing over the blocks that force-change the list of validators. Before this PR, even if we ship a more recent checkpoint to bypass this issue, users would still use their database and be stuck at the block in question.

### DIFF
--- a/bin/light-base/src/lib.rs
+++ b/bin/light-base/src/lib.rs
@@ -409,16 +409,35 @@ impl<TChain, TPlat: Platform> Client<TChain, TPlat> {
                 }),
                 finalized_serialize::decode_chain(config.database_content),
             ) {
-                (Ok(Ok(genesis_ci)), _, Ok((database, _))) => {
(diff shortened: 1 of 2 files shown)


tomaka:
> Basically, smoldot is incapable of syncing over the blocks that force-change the list of validators.

This is tracked by https://github.com/paritytech/smoldot/issues/515

melekes (approved):
👍

PR #2402, opened 2022-06-17
Clarify parameter index in error messages

tomaka:
We recently had two errors in substrate-connect saying `Parameter #1 is invalid when calling ...`.
Without thinking about it, I intuitively thought that this was the first parameter. But no, it's the second parameter because it's indexed from 0.

This PR makes in my opinion the error messages a bit more clear.


melekes (commented):
👍 

> Without thinking about it, I intuitively thought that this was the first parameter. 

I would choose to show `{param_num+1}` instead of adding "index of". Index is programming term.

tomaka:
Unfortunately the `derive_more` library doesn't let you do `param_num+1`. I'd have to implement the `Display` trait manually, which is a bit annoying.

melekes (approved):

PR #2406, opened 2022-06-20
Make storage proofs and call proofs more zero cost

tomaka:
This PR does three things:

- Decoding a Merkle proof no longer returns a `Vec<Vec<u8>>` but a `Vec<&[u8]>`, meaning that we no longer copy the entire proof somewhere else in memory when we decode it.
- The `network/service` module, which receives the undecoded Merkle proofs, now verifies that the proof is correct then return a `EncodedMerkleProof` instead of a `Vec<Vec<u8>>`. This `EncodedMerkleProof` contains the undecoded proof, and has a `decode()` method that does basically `decode(self.0).unwrap()`. The `EncodedMerkleProof` is just an object here for strong typing purposes.
- Since their responses are in the same format anyway, I've merged the `call_proof` and `storage_proof` modules into one.

Thanks to these changes, we never copy the content of the Merkle proof after receiving it.


tomaka:
Oops, sorry for merging before a review

Issue #2407, opened 2022-06-20
Smoldot Crash during statemine/statemint tests


wirednkod:
While talking with @PierreBesson , he mentioned that during various tests that he was performing, for connecting light client with Statemine (using Substrate connect) the following errors appeared:

![smoldot crash](https://user-images.githubusercontent.com/5408605/174625057-32b66b91-8e71-4d34-900d-8972ee50ae30.png)

tomaka:
What I think happens is:

- We queue a parahead fetching operation, but the operation doesn't start immediately.
- The block whose parahead we are fetching gets pruned. The code currently unpins the block.
- The fetching operation gets executed. The panics happens because the block is unpinned.

wirednkod:
> What I think happens is:
> 
> * We queue a parahead fetching operation, but the operation doesn't start immediately.
> * The block whose parahead we are fetching gets pruned. The code currently unpins the block.
> * The fetching operation gets executed. The panics happens because the block is unpinned.

Fwiw this is an error thrown during multiple local tests, Pierre was trying to connect to statemint/statemine

tomaka:
My diagnostic is wrong. The pruning is done after the parahead fetch operation.

tomaka:
I have no idea what causes this, but https://github.com/paritytech/smoldot/pull/2631 improves the error message.
It might also have been accidentally fixed by a couple of PRs since then.
I'm going to close this issue, please open a new one if you have a similar error.

PR #2408, opened 2022-06-20
Indicate caller in backtrace on panic in pinned_block_runtime_lock

tomaka:
cc https://github.com/paritytech/smoldot/issues/2407


tomaka:
Ah, I forgot that `#[track_caller]` doesn't work on async functions.

PR #2416, opened 2022-06-21
Improve connection state flags in collection.rs

tomaka:
This PR replaces the multiple boolean flags with an enum.
This enum clarifies which state is valid or not.

This PR also fixes some debug_assert!s panicking because there was an ambiguity as to the meaning of `shutting_down`, especially because there's another flag named `start_shutdown_called`.
Some parts of the code set it to true even if the shutdown was initiated locally, while other parts of the code assume it means "shutdown initiated remotely".

### DIFF
--- a/src/libp2p/collection.rs
+++ b/src/libp2p/collection.rs
@@ -237,20 +237,28 @@ pub struct Network<TConn, TNow> {
 
 // TODO: consider refactoring this struct to use an enum
 struct Connection<TConn> {
-    /// `true` if the connection is still in its handshake phase.
-    handshaking: bool,
-
-    /// `true` if [`Network::start_shutdown`] has been called on this connection.
-    ///
-    /// Even if the remote starts the shutdown at the same time, from an API perspective if this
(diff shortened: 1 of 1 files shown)


src/libp2p/collection.rs
```diff
@@ -237,20 +237,28 @@ pub struct Network<TConn, TNow> {
 
 // TODO: consider refactoring this struct to use an enum
```
melekes, line 238: remove?

src/libp2p/collection.rs
```diff
@@ -413,10 +417,17 @@ where
     ///
     pub fn start_shutdown(&mut self, connection_id: ConnectionId) {
         let connection = self.connections.get_mut(&connection_id).unwrap();
-        assert!(!connection.start_shutdown_called);
-        assert!(!connection.shutting_down);
-        connection.start_shutdown_called = true;
-        connection.shutting_down = true;
+
+        let is_established = match connection.state {
```
melekes, line 421: this could be a method on ConnecState, right?

melekes (approved):

Issue #2245, opened 2022-04-25
Add unstable JSON-RPC function that exports recent networking events


tomaka:
We have a big problem in substrate-connect right now: the behavior of smoldot when it comes to the peer-to-peer networking is extremely opaque, and if smoldot doesn't manage to connect to other peers on the chain it is very difficult to understand why.

To solve this problem, I propose to add a new unstable subscription-style JSON-RPC function that notifies of all the networking events that happen or that have happened recently or that concern an item that is still alive or that concern an item that has died recently.

Showing on a UI the state of the networking in a way that is understandable is a complicated problem, and by providing to the UI an exhaustive list of networking events that have happened recently, we move to substrate-connect (or alternative libraries built on top of smoldot) the problem of visualizing the state of the networking. This will make it possible to easily experiment with various different visualizations.

In details, the notifications would look like this:

```js
{
    "type": "startConnect",
    "when": ... # A unix timestamp
    "connectionId": ... # Opaque ids allocated by smoldot
    "multiaddr": ...
}
```

```js
{
    "type": "connected",
    "when": ... # A unix timestamp
    "connectionId": ...
}
```

```js
{
    "type": "handshakeFinished",
    "when": ... # A unix timestamp
    "connectionId": ...
    "peerId": ...
}
```

```js
{
    "type": "stop",   # Can mean either "disconnected" or "stopped trying to connect"
    "when": ... # A unix timestamp
    "connectionId": ...
    "reason": ...
}
```

```js
{
    "type": "outSlotAssign",
    "when": ... # A unix timestamp
    "peerId": ...
}
```

```js
{
    "type": "outSlotUnassign",
    "when": ... # A unix timestamp
    "peerId": ...
}
```

```js
{
    "type": "inSlotAssign",
    "when": ... # A unix timestamp
    "peerId": ...
}
```

```js
{
    "type": "inSlotUnassign",
    "when": ... # A unix timestamp
    "peerId": ...
}
```

```js
{
    "type": "inSlotToOutSlot",   # Atomically unassign and in slot and assigns an out slot
    "when": ... # A unix timestamp
    "peerId": ...
}
```

```js
{
    "type": "substreamOutOpen",
    "when": ... # A unix timestamp
    "connectionId": ...
    "substreamId": ...
    "protocolName": ...
}
```

```js
{
    "type": "substreamOutAccept",
    "when": ... # A unix timestamp
    "substreamId": ...
}
```

```js
{
[...]

PR #2417, opened 2022-06-21
Add JSON-RPC definitions of networking events subscriptions

tomaka:
cc https://github.com/paritytech/smoldot/issues/2245

This PR can be reviewed, but I would maybe not merge it now.
There is still quite a lot of preparation before we can implement this JSON-RPC method.


melekes (approved):

melekes (approved):
👍

PR #2420, opened 2022-06-22
Fix missing match blocks in collection.rs

tomaka:
Fixes the two `_ => todo!()` blocks.

I initially left them as `_ => todo!()` because I thought that these were only invalid states, and so didn't really need much caring.

However it turns out I forgot to handle the `(StartRequest, ShutdownWaitingAck)` combination (which, before this PR would panic, and after this PR does nothing), so it's a good thing that they are now explicitly laid out.

While this fixes a potential panic, I've never actually seen this panic happen and it might not actually be possible for it to happen, so I don't think it's worth mentioning in the CHANGELOG.

### DIFF
--- a/src/libp2p/collection/multi_stream.rs
+++ b/src/libp2p/collection/multi_stream.rs
@@ -418,6 +418,7 @@ where
             (
                 CoordinatorToConnectionInner::AcceptInNotifications { .. }
                 | CoordinatorToConnectionInner::RejectInNotifications { .. }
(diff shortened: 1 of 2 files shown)


tomaka:
Pushed another commit. I realized that if the coordinator requests to start shutting down after the tasks has already started shutting down on its own, we should ignore the request rather than panic.

melekes (approved):
👍

Issue #1822, opened 2021-12-30
fork_tree panic


tomaka:
https://github.com/paritytech/smoldot/blob/f9dbbfd3049de54da156e25b8dbcc1e7b0f84cc4/src/chain/fork_tree.rs#L477-L478

The `unwrap()` panics.

PR #2424, opened 2022-06-22
Adding a fork tree fuzzing code

tomaka:
Close https://github.com/paritytech/smoldot/issues/1822

The fork tree has a history of having a lot of bugs that have been fixed over time.

In order to have a reasonably good confidence that there's no panic anymore, this PR adds a fuzzer.
The fuzzer randomly generates operations amongst all the possible mutable functions that the fork tree lets you perform, then applies them.

As of the opening of this PR, it panics because of https://github.com/paritytech/smoldot/pull/2423, but if I manually apply the fix in #2423 then the panics disappear.

### DIFF
--- a/bin/fuzz/fuzz_targets/fork-tree.rs
+++ b/bin/fuzz/fuzz_targets/fork-tree.rs
@@ -125,8 +125,8 @@ impl<'a> arbitrary::Arbitrary<'a> for OperationsList {
                 return Ok(ControlFlow::Continue(()));
             }
 
-            // 9/10th of the time, insert a new node in the tree.
-            if u.ratio(9, 10)? {
+            // 9/10th of the time, or if the tree is empty, insert a new node in the tree.
+            if tree_root.children.borrow().is_empty() || u.ratio(9, 10)? {
                 let index = u.choose_index(all_nodes.len())?;
 
                 operations.push(Operation::Insert {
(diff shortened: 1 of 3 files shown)


bin/fuzz/fuzz_targets/fork-tree.rs
```diff
@@ -0,0 +1,166 @@
...
+                    children: RefCell::new(Vec::new()),
+                }));
+                next_node_id += 1;
+
+                return Ok(ControlFlow::Continue(()));
+            }
+
+            // The remaining half of the time, we prune ancestors. Otherwise, we prune uncles.
+            if u.ratio(1, 2)? {
+                let index = u.choose_index(all_nodes.len() - 1)? + 1; // Avoid the tree root
```
melekes, line 149: this will fail if the tree only consists of the single root element, right? It's fine, just something to be aware of if it's the case indeed.

bin/fuzz/fuzz_targets/fork-tree.rs
```diff
@@ -0,0 +1,166 @@
...
+                operations.push(Operation::PruneAncestors(
+                    all_nodes[index].id.get().unwrap(),
+                ));
+                tree_root = all_nodes[index].clone();
+                all_nodes[index].id.set(None);
+            } else {
+                let index = u.choose_index(all_nodes.len() - 1)? + 1; // Avoid the tree root
```
melekes, line 156: same comment

melekes (approved):

bin/fuzz/fuzz_targets/fork-tree.rs
tomaka: Yeah, `choose_index(0)` returns an error, which then instructs the fuzzer to change the data.

melekes (approved):

PR #2425, opened 2022-06-22
Add support for digest item other type

fanyang1988:
The DigestItem type in substrate has a other type:

```rust
pub enum DigestItemRef<'a> {
        ///...
	/// Any 'non-system' digest item, opaque to the native code.
	Other(&'a [u8]),
	///...
}
```

This will be used for some chain based on substrate, and if use this digestitem, the smoldot will failed to sync blocks from it.

This PR fix this problem by add this type to DigestItem.

### DIFF
--- a/src/header.rs
+++ b/src/header.rs
@@ -636,7 +636,8 @@ impl<'a> DigestRef<'a> {
                 DigestItem::BabeSeal(_) => return Err(Error::SealIsntLastItem),
                 DigestItem::Beefy { .. }
                 | DigestItem::PolkadotParachain { .. }
-                | DigestItem::Frontier { .. } => {}
+                | DigestItem::Frontier { .. }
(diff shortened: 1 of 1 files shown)


src/header.rs
```diff
@@ -978,6 +980,9 @@ pub enum DigestItemRef<'a> {
         opaque: &'a [u8],
     },
 
+    /// Some other thing. Unsupported and experimental.
```
tomaka, line 983:
```suggestion
    /// Some other thing. Always ignored.
```

I don't think it's true that it's "unsupported and experimental", but I'd like to explicitly mention that it's ignored, in order to make it clear that it's not a good idea to put something consensus-critical in there, similar to what Substrate says.

tomaka (changes requested):
Thanks for the PR!

tomaka (approved):

PR #2426, opened 2022-06-23
Pass `type: "module"` to `new Worker()`

tomaka:
Deno throws an exception if we call `new Worker` without `type: "module"`.

See the comments for more information.

cc @harrysolovay


wirednkod:
Maybe the info in the [README.md](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/javascript/README.md#about-the-worker) should be updated accordingly @tomaka

wirednkod (approved):

Issue #2370, opened 2022-06-10
Consider adding a "shutting down" phase in the API of peers.rs


tomaka:
See https://github.com/paritytech/smoldot/pull/2369 for an explanation.

Instead of having a fake requests system, we should consider notifying the upper API layers that we only have one connection and that it is shutting down.

This might cause other consistency issues, so I don't know whether it's a good idea.

PR #2429, opened 2022-06-24
Report reason for shutdown in collection.rs

tomaka:
When a connection starts its shutdown phase, we now report why.
This is at the moment only reported by the API of `collection.rs`, and we need https://github.com/paritytech/smoldot/issues/2370 in order to propagate it further.


melekes (approved):
👍

PR #2432, opened 2022-06-24
Refactor platform.rs

tomaka:
As promised, this PR refactors `platform.rs`.

Before this PR, we were allocating objects on the heap then turning the heap pointer into a `u32`. When the JS would call methods, we would convert the `u32` back into a heap pointer. This is a bit risky.

This PR completely removes this and tracks the networking state in two maps, one of connections and one of streams.

It also implements multi-stream connections, and adds a function that was missing in the bindings: `connection_stream_close`.

### DIFF
--- a/bin/wasm-node/rust/src/platform.rs
+++ b/bin/wasm-node/rust/src/platform.rs
@@ -460,7 +460,8 @@ enum ConnectionInner {
         /// Peer id we're connected to.
         peer_id: smoldot_light_base::PeerId,
         /// List of substreams that the host (i.e. JavaScript side) has reported have been opened,
-        /// but that haven't been reported through [`Platform::next_substream`] yet.
+        /// but that haven't been reported through [`smoldot_light_base::Platform::next_substream`]
+        /// yet.
         opened_substreams_to_pick_up: VecDeque<(u32, PlatformSubstreamDirection)>,
         /// Number of objects (connections and streams) in the [`Platform`] API that reference
         /// this connection. If it switches from 1 to 0, the connection must be removed.
--- a/bin/wasm-node/rust/src/platform.rs
+++ b/bin/wasm-node/rust/src/platform.rs
@@ -459,8 +459,8 @@ enum ConnectionInner {
     MultiStream {
         /// Peer id we're connected to.
         peer_id: smoldot_light_base::PeerId,
-        /// List of substreams that the host (i.e. JS side) has reported have been opened, but
-        /// that haven't been reported through [`Platform::next_substream`] yet.
+        /// List of substreams that the host (i.e. JavaScript side) has reported have been opened,
+        /// but that haven't been reported through [`Platform::next_substream`] yet.
         opened_substreams_to_pick_up: VecDeque<(u32, PlatformSubstreamDirection)>,
         /// Number of objects (connections and streams) in the [`Platform`] API that reference
         /// this connection. If it switches from 1 to 0, the connection must be removed.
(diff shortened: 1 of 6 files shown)


bin/wasm-node/rust/src/platform.rs
```diff
@@ -106,183 +104,463 @@ impl smoldot_light_base::Platform for Platform {
...
+                    bytes,
+                );
+
+                read_buffer.buffer_first_offset += read_buffer_advance;
+                bytes -= read_buffer_advance;
+            }
 
-        this.messages_queue_first_offset += bytes;
+            // Avoid the whole locking process if `bytes` is 0.
+            if bytes == 0 {
```
melekes, line 306: so we only exit when we have exact number of `bytes` (i.e. `read_exact`)? what if stream is closed?

bin/wasm-node/rust/src/platform.rs
```diff
@@ -106,183 +104,463 @@ impl smoldot_light_base::Platform for Platform {
...
-    /// Prevents the [`Connection`] from being unpinned.
-    _pinned: marker::PhantomPinned,
-}
+pub(crate) struct StreamWrapper((u32, u32), ReadBuffer);
+
+impl Drop for StreamWrapper {
+    fn drop(&mut self) {
+        let mut lock = STATE.try_lock().unwrap();
+
+        let connection = lock.connections.get_mut(&self.0 .0).unwrap();
```
melekes, line 349: format the code?

```suggestion
        let connection = lock.connections.get_mut(&self.0.0).unwrap();
```

bin/wasm-node/rust/src/platform.rs
```diff
@@ -106,183 +104,463 @@ impl smoldot_light_base::Platform for Platform {
...
+                unsafe {
+                    bindings::connection_close(self.0);
+                }
             }
         }
     }
 }
 
+lazy_static::lazy_static! {
+    static ref STATE: Mutex<NetworkState> = Mutex::new(NetworkState {
```
melekes, line 431: is it a common / preferred way of writing "bindings" code? ie having global state as a const? just curious

bin/wasm-node/rust/src/platform.rs
```diff
@@ -297,34 +575,90 @@ pub(crate) fn stream_message(connection_id: u32, _stream_id: u32, ptr: u32, len:
...
+            (connection_id, stream_id),
+            Stream {
+                closed: false,
+                messages_queue: VecDeque::with_capacity(8),
+                something_happened: event_listener::Event::new(),
+            },
+        );
+
+        if _prev_value.is_some() {
+            panic!() // StreamId has been reused.
```
melekes, line 603:
```suggestion
            panic!("StreamId has been reused")
```

bin/wasm-node/rust/src/platform.rs
tomaka: This function is to discard the first N bytes from the read buffer. If you pass a N that is larger than the number of bytes in the read buffer you get a panic. There is no relation with the stream being closed or not.

bin/wasm-node/rust/src/platform.rs
tomaka: Rustfmt formats this code as `self.0 .0`.
In the past the compiler treated this as `self.(0.0)`, which was a compilation error. This has been fixed, but rustfmt has apparently not been updated.

bin/wasm-node/rust/src/platform.rs
tomaka: Well, this code is specifically to bind the light client to the "wasm platform" capabilities, so I went for a global const because the "platform" is a global thing.
A similar example is that on Linux the list of open file descriptors is a global "const" as well.

melekes (approved):

Issue #1768, opened 2021-12-10
Add known peers to wasm-node database


tomaka:
The database of the wasm node should contain the content of the k-buckets (at least partially) so that we don't rely on bootnodes every single time.

tomaka:
Thanks to #1773 we can now be sure that adding the peers to the database will not make it exceed the maximum allowed size for local storage.

PR #2439, opened 2022-06-27
Store peers in the database

tomaka:
Close https://github.com/paritytech/smoldot/issues/1768

We currently store the finalized state of the chain in the database as a JSON text.
This PR modifies the format and wraps this finalized state of the chain within another object.
The format of the database now looks like `{chain:{...}, nodes:{...}}`.

The `nodes` entry stores the list of nodes that we know are part of the network.

The next time a chain is added with a database, we immediately insert the nodes in the k-buckets.

Because the database format is intentionally not stable, we simply ignore faulty values. This PR breaks the format of the database, but database coming from previous versions will simply be ignored.


bin/light-base/Cargo.toml
```diff
@@ -21,6 +21,7 @@ itertools = "0.10.3"
 log = { version = "0.4.17", default-features = false }
 lru = { version = "0.7.7", default-features = false }  # TODO: there's no way to use a custom hasher; remove this dependency
 rand = "0.8.5"
+serde = { version = "1.0.137", default-features = false, features = ["alloc", "derive"] }
```
melekes, line 24:
```suggestion
serde = { version = "1.0", default-features = false, features = ["alloc", "derive"] }
```

does it have to be >= 137?

bin/light-base/src/lib.rs
```diff
@@ -1236,3 +1221,112 @@ async fn start_services<TPlat: Platform>(
...
+) -> String {
+    // Craft the structure containing all the data that we would like to include.
+    let mut database_draft = SerdeDatabase {
+        chain: match services.sync_service.serialize_chain_information().await {
+            Some(ci) => {
+                let encoded = finalized_serialize::encode_chain(&ci);
+                serde_json::from_str(&encoded).unwrap()
+            }
+            None => {
+                // If the database can't be obtained, we just return a dummy value that will
```
melekes, line 1237: are you saying `services.sync_service.serialize_chain_information()`  above obtains the database? maybe you thought to say "If the chain information can't be obtained"?

bin/light-base/Cargo.toml
tomaka: Well, I don't know, I might be using features that were added only in `1.0.130` or something, in which case we want to prevent users from trying to compile with a version below that.

In general I don't understand what it would bring to use `1.0` instead of `1.0.137`.

melekes: > In general I don't understand what it would bring to use 1.0 instead of 1.0.137.

broader scope of supported versions, so other libraries who're dependent on smoldot don't have to upgrade to 137?

You can check if code compiles using https://github.com/taiki-e/cargo-minimal-versions.

Anyway, I don't have strong opinion here (in smoldot).

melekes (approved):

PR #2443, opened 2022-06-28
Split the types used in sync and in networking

tomaka:
So far, the syncing code was using `network::protocol::GrandpaWarpSyncResponse` in order to store a GrandPa warp sync response.

This PR changes that to have a separate type named `WarpSyncFragment` used in the syncing. The networking code is untouched. The higher level code (`sync_service`) does the conversion between the two.

There is no real benefit to this change except better code clarify, but I think that it's worth it.


melekes (approved):
👍

PR #2454, opened 2022-07-01
Adjust when log message about block request error is printed

tomaka:
At the moment, if a blocks request fails we print a warning log message saying that the peer is likely incompatible.

We do this even when it doesn't make sense (such as `EmptyResponse`, which simply indicates that the remote doesn't know the requested block), and we don't print it in some situations where it should make sense (for example protocol not supported).

This PR tweaks the condition under which we print the log message.


melekes (approved):
👍

PR #2455, opened 2022-07-01
Refactor the demo

tomaka:
This PR refactors the demo so that we don't have to keep adding `} else if {` statements for every chain.


melekes (approved):
👍

PR #2465, opened 2022-07-05
Fix chain_subscribeAllHeads not recreating the channel if it dies

tomaka:
Subscribing to `chain_subscribeAllHeads` opens a channel of notifications with the runtime service.
If this channel gets too full (i.e. more than 32 notifications in queue), the channel is force-closed.

This typically happens if the finalized block jumps more than 32 blocks at once, or if there are 32 new blocks in a short amount of time, because in that situation we will report one notification per finalized block or per new block.

In theory, the finalized block isn't supposed to jump more than 32 blocks at once and we're not supposed to get 32 new blocks at once. But this tends to happen for example if we lose Internet connectivity for a while then gain it back.

Unfortunately, when the channel is closed, `chain_subscribeAllHeads` will not reopen a new one, and no new block will ever be reported to the user. The consequence, in practice, is that PolkadotJS will look frozen.
This is what this PR fixes. We now properly reopen a channel. Some blocks might be missed (as this is the entire reason why the channel needs to be closed), but because nothing in the API of `chain_subscribeAllHeads` exists to handle these kind of situations, this is the best that we can do.

The diff looks big, but what I did is basically move `let new_blocks = ` within the asynchronous task.

### DIFF
--- a/bin/light-base/src/json_rpc_service/state_chain.rs
+++ b/bin/light-base/src/json_rpc_service/state_chain.rs
@@ -397,82 +397,90 @@ impl<TPlat: Platform> Background<TPlat> {
             )
             .await;
 
-        let mut new_blocks = {
-            // The buffer size should be large enough so that, if the CPU is busy, it doesn't
-            // become full before the execution of the runtime service resumes.
-            // The maximum number of pinned block is ignored, as this maximum is a way to avoid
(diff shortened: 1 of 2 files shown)


tomaka:
I've also bumped the buffer size from 32 to 64, because ideally we would like to avoid the situation altogether.

PR #2471, opened 2022-07-06
Fix some TODO in ext_crypto_finish_batch_verify_version_1

tomaka:
This TODO makes it look like we don't verify signatures. However that's not true, and it's a bit more subtle. See the diff for an explanation.


melekes (approved):
👍

PR #2474, opened 2022-07-06
Two small improvements to k-buckets code

tomaka:
- When inserting a new entry, we now return the entry that has been popped.
- Return proper errors in the `Result`s, instead of `()`.


melekes (approved):
👍

PR #2476, opened 2022-07-06
Silence NoPeers discovery errors

tomaka:
People keep mentioning this warning about no peers, but there's actually nothing that can be done anyway, and thus no point in printing the warning.


melekes (approved):
👍

Issue #2478, opened 2022-07-07
Proper block numbers handling


tomaka:
Some SCALE-encoded data structures must be decoded differently depending on whether block numbers are 4 bytes or 8 bytes long.
At the moment, smoldot assumes that block numbers are always 4 bytes long. This is problematic, because chains that use 8 bytes aren't supported.

We must update the following things to add an `expected_block_number_size` parameter:

- [x] Blocks announces and requests networking protocols
- [x] GrandPa networking protocols
- [ ] Headers parsing
- [x] Justifications parsing

Additionally:

- [x] Add a way to determine this block number from chain specs.

tomaka:
Note keeping:

This specific line also needs to be fixed, I unfortunately forgot about it:

https://github.com/paritytech/smoldot/blob/09fd6dbe37228d7cd5465d0bef263bbf768383d4/src/network/service.rs#L1324

PR #2479, opened 2022-07-07
Provide a block number size parameter to the networking service

tomaka:
First bullet point of https://github.com/paritytech/smoldot/issues/2478

### DIFF
--- a/src/network/protocol/block_request.rs
+++ b/src/network/protocol/block_request.rs
@@ -65,7 +65,10 @@ pub enum BlocksRequestConfigStart {
 // for protocol definition.
 
 /// Builds the bytes corresponding to a block request.
-pub fn build_block_request(config: &BlocksRequestConfig) -> impl Iterator<Item = impl AsRef<[u8]>> {
+pub fn build_block_request(
+    block_number_bytes: usize,
+    config: &BlocksRequestConfig,
+) -> impl Iterator<Item = impl AsRef<[u8]>> {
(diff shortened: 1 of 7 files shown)


src/network/protocol/block_request.rs
```diff
@@ -82,23 +85,11 @@ pub fn build_block_request(config: &BlocksRequestConfig) -> impl Iterator<Item =
...
-            // In other words, we send the bytes containing the little endian block number
-            // followed with enough 0s to make Substrate accept the request.
-            // This is a hack, but it is not really fixable in smoldot alone and shows a
-            // bigger problem in the Substrate network protocol/architecture. A better
-            // protocol would for example use the SCALE-compact encoding, which doesn't
-            // have this issue.
-            either::Right(protobuf::bytes_tag_encode(3, n.to_le_bytes()).map(either::Right))
+            let mut data = Vec::with_capacity(block_number_bytes);
+            data.extend_from_slice(&n.to_le_bytes());
+            // TODO: unclear what to do if the block number doesn't fit within `expected_block_number_bytes`
```
melekes, line 90: should we at least add a debug statement?

melekes (approved):

src/network/protocol/block_request.rs
tomaka: No logging statements in `src`. I'd like to have a proper solution eventually instead.

Issue #2475, opened 2022-07-06
Zeitgeist parachain integration - unknown consensus engine.


yornaath:
We at zeitgeist are trying to integrate substrate-connect(smoldot) into our frontend application layer using the extension, but running into the following errors:

```
[runtime-zeitgeist] Failed to decode header from sync service: Unknown consensus engine specified in a digest log: [110, 109, 98, 115]
[runtime-zeitgeist] Failed to download :code and :heappages of blocks 0xa800…4f96: Couldn't decode header: Unknown consensus engine specified in a digest log: [110, 109, 98, 115]
```

We are building of the purestake/substrate fork and using nimbus to implement our consensus layer, if we are using the same algorithm and format exactly as moonbeam is beyond my awareness atm, but maybe @sea212 have some more info that is relevant. 

Ref
[Frontend instantiation of Scprovider](https://github.com/zeitgeistpm/tools/blob/88b3cfcdcc9add5157102eae75fbb9264a43c597/packages/sdk/src/util/polkadot.ts#L26)
[Para chain spec](https://github.com/zeitgeistpm/tools/blob/%23201-substrate-connect-integration/packages/sdk/src/types/chainspecs/zeitgeist_parachain.json)
[Runtime creation](https://github.com/zeitgeistpm/zeitgeist/blob/fc3de00d4ad85470b5522d414437034cc29e9e11/runtime/src/lib.rs#L225)

sea212:
We use [`Nimbus`](https://github.com/PureStake/nimbus/tree/moonbeam-polkadot-v0.9.19) for parachain consensus and [`parachain-staking pallet`](https://github.com/PureStake/moonbeam/tree/8ec7a31bb511292955a73ae50b966cb1a44c6818/pallets/parachain-staking) for managing candidate sets, staking and determining the active set.

Current snapshot of the crate versions used:
```rust
nimbus-primitives = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/nimbus", optional = true }
pallet-author-inherent = { branch = "moonbeam-polkadot-v0.9.19", default-features = false, git = "https://github.com/purestake/nimbus", optional = true }
pallet-author-mapping = { rev = "78db06c0203f61b35059304f7194ed5c10dcfda8", default-features = false, git = "https://github.com/purestake/moonbeam", optional = true }
[...]

PR #2481, opened 2022-07-07
Revisit story around unknown digest log items

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2475

While Substrate supports Aura and Babe, other teams are free to extend Substrate to support other consensus algorithms (such as Nimbus). This PR revisits our strategy regarding these other consensus algorithms.

Before this PR, block headers that contain items that aren't supported by smoldot would simply fail to parse, with some whitelisted exceptions such as Polkadot parachains, Beefy, and Frontier. These exceptions were added for consensus algorithms that don't have any influence on whether a block is valid or not.

After this PR, we now successfully parse block headers even if they contain unrecognized consensus engines. However, the blocks verification code will fail if it is asked to verify a block header with an unrecognized consensus engine.

In order to understand this change, one must note that the parachain code of smoldot doesn't actually verify blocks. It simply follows the blocks that the relay chain considers as the best block for that parachain. If the relay chain says a block is valid, then the parachain code will consider that block as valid as well. It never actually verifies them itself.

This means that the change in this PR now enables parachains with unrecognized consensus engines.

Relay/standalone chains with unrecognized consensus engines, however, still aren't supported. After this PR, they will simply fail to verify blocks rather than fail to parse blocks.

A slightly related change that I've sneaked in this PR is how we handle chains that don't use any consensus engine.
Before this PR, chains that don't use any consensus engine are marked as `AllAuthorized` and will have see all incoming headers accepted. This can potentially lead to a DoS vector, as anyone can just create millions of fake blocks that are all considered as valid.
After this PR, chains that don't use any consensus engine are marked as `Unknown` and will now fail all block verifications. Again, this doesn't have any influence on parachains as the parachains code doesn't verify blocks anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -413,6 +413,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "cast"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
+
+[[package]]
 name = "cc"
 version = "1.0.72"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


tomaka:
Ok I'm a bit stupid. I've just realized that this breaks everything, since Beefy/Polkadot parachains are now considered as unknown consensus engines and thus blocks validation will fail.

tomaka:
Fixed my previous comment by adding an `allow_unknown_consensus_engines` flag to the configuration of everything relevant.
It's set to `true` in the light client.

melekes (approved):

sea212:
Awesome!

PR #2485, opened 2022-07-08
Use block_number_bytes when decoding justifications

tomaka:
Fourth point of https://github.com/paritytech/smoldot/issues/2478

Properly parses justifications depending on the number of bytes in the block number.

### DIFF
--- a/bin/fuzz/fuzz_targets/grandpa-justification-parse.rs
+++ b/bin/fuzz/fuzz_targets/grandpa-justification-parse.rs
@@ -17,6 +17,9 @@
 
 #![no_main]
 
-libfuzzer_sys::fuzz_target!(|data: &[u8]| {
-    let _ = smoldot::finality::justification::decode::decode_grandpa(data);
+libfuzzer_sys::fuzz_target!(|params: (&[u8], u8)| {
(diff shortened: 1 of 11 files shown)


bin/fuzz/fuzz_targets/grandpa-justification-parse.rs
```diff
@@ -17,6 +17,9 @@
 
 #![no_main]
 
-libfuzzer_sys::fuzz_target!(|data: &[u8]| {
-    let _ = smoldot::finality::justification::decode::decode_grandpa(data);
+libfuzzer_sys::fuzz_target!(|params: (&[u8], u8)| {
+    let _ = smoldot::finality::justification::decode::decode_grandpa(
+        params.0,
```
melekes, line 22: shouldn't this be `params.0 + 1, usize::from(params.1)`?

tomaka: Hmmm, no

`params.0` is the data to decode, and `params.1` is the block number size. I pass `params + 1` to avoid passing 0, which intentionally panics.

melekes (approved):
👍

PR #2486, opened 2022-07-08
Use block_number_bytes when parsing parachain data

tomaka:
cc https://github.com/paritytech/smoldot/issues/2478
I forgot this point in https://github.com/paritytech/smoldot/issues/2478, but here is the PR.


melekes (approved):
👍

Issue #2488, opened 2022-07-09
Remove the worker from the JS bindings


tomaka:
I think it is time to admit defeat. We cannot work around issues in the broader JS ecosystem. Let's remove the worker altogether.

This means that we won't be able to detect infinite loops anymore, but the trade-off is I think necessary.

Unfortunately, this complicates the code quite a bit, as we can no longer afford the double panics that still happen: https://github.com/paritytech/smoldot/issues/739

tomaka:
This will also definitely kill https://github.com/paritytech/smoldot/pull/636

Issue #739, opened 2021-04-08
Double panics can still happen in the wasm node


tomaka:
When a panic happens in the Wasm node, we call `throw`, and intentionally let the JavaScript worker die as a result.

This is totally intended. A panic could leave memory in a corrupted state, and calling functions again is undesirable.
See [this paragraph](https://github.com/paritytech/smoldot/blob/3a4d16dc0d430e02f518c56975339a6d76b7624b/bin/wasm-node/rust/src/ffi/bindings.rs#L51-L68) for context.

It seems that, despite the worker errorring, the timeouts and sockets that are still alive in the worker continue to run and continue to call the smoldot Wasm again.

tomaka:
To reproduce: put a `panic!()` somewhere in a background task, then observe how there are two panics instead of one.

The second panic (something regarding `WouldBlock`) comes from the fact that we call `try_lock` on a mutex that isn't expected to be locked but is still locked because of the first panic.
The proper fix is that we never reach again Wasm code again after the first panic happened.

PR #2498, opened 2022-07-12
Remove usage of a WebWorker/worker thread

tomaka:
Close https://github.com/paritytech/smoldot/issues/2488
Close https://github.com/paritytech/smoldot/issues/739

This PR removes the usage of web workers altogether.
The Wasm VM inside of which smoldot runs is directly spawned by the `start` function, rather than inside of a worker.

This will also make it easier to implement WebRTC, as currently WebRTC connections can't be used from inside workers.

Removing the worker was very easy, but the difficulty was to properly handle shutdowns and panics. Right now, we just kill the worker altogether when that happens. This is no longer an option.

Can be reviewed commit by commit, although at the end I've kind of given up with making small self-contained changes.


bin/wasm-node/javascript/src/client.ts
```diff
@@ -599,58 +410,55 @@ export function start(options?: ClientOptions): Client {
...
+            throw new AlreadyDestroyedError();
+          return instance.databaseContent(chainId, maxUtf8BytesSize);
+        },
+        remove: () => {
+          if (alreadyDestroyedError)
+            throw alreadyDestroyedError;
+          if (wasDestroyed.destroyed)
+            throw new AlreadyDestroyedError();
+          wasDestroyed.destroyed = true;
+          console.assert(chainIds.has(newChain));
```
wirednkod, line 448: is this assert needed?

bin/wasm-node/javascript/src/client.ts
```diff
@@ -365,207 +363,20 @@ export function start(options?: ClientOptions): Client {
...
-      for (const promise of chain[1].databasePromises) {
-        promise.reject(workerError)
-      }
-    }
-    chains.clear();
-  });
-
-  // The first message expected by the worker contains the configuration.
-  postMessage(worker, {
+  const instance = startWorker({
```
wirednkod, line 375: maybe we should give a non-worker-related-name? maybe `startInstance`?

bin/wasm-node/javascript/src/client.ts
```diff
@@ -599,58 +410,55 @@ export function start(options?: ClientOptions): Client {
...
-      workerError = new AlreadyDestroyedError();
-
-      if (livenessTimeout !== null)
-        globalThis.clearTimeout(livenessTimeout)
-
-      return workerTerminate(worker)
+    terminate: async () => {
+      if (alreadyDestroyedError)
+        throw alreadyDestroyedError
+      alreadyDestroyedError = new AlreadyDestroyedError();
```
wirednkod, line 460: Im not sure why this is needed

bin/wasm-node/javascript/src/instance/instance.ts
```diff
@@ -0,0 +1,305 @@
...
+          // Write the database content into memory.
+          const databaseContentLen = Buffer.byteLength(databaseContent, 'utf8');
+          const databaseContentPtr = instance.exports.alloc(databaseContentLen) >>> 0;
+          Buffer.from(instance.exports.memory.buffer)
+            .write(databaseContent, databaseContentPtr);
+
+          // Write the potential relay chains into memory.
+          const potentialRelayChainsLen = potentialRelayChains.length;
+          const potentialRelayChainsPtr = instance.exports.alloc(potentialRelayChainsLen * 4) >>> 0;
+          for (let idx = 0; idx < potentialRelayChains.length; ++idx) {
```
melekes, line 178: why not `idx++`?

melekes (approved):
👍

bin/wasm-node/javascript/src/client.ts
```diff
@@ -599,58 +410,55 @@ export function start(options?: ClientOptions): Client {
...
+      // Resolve the promise that `addChain` returned to the user.
+      const newChain: Chain = {
+        sendJsonRpc: (request) => {
+          if (alreadyDestroyedError)
+            throw alreadyDestroyedError;
+          if (wasDestroyed.destroyed)
+            throw new AlreadyDestroyedError();
+          if (!options.jsonRpcCallback)
+            throw new JsonRpcDisabledError();
+          if (request.length >= 8 * 1024 * 1024)
```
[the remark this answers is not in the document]
tomaka: This code was already there before. By principle I'm not going to modify it in this PR.

bin/wasm-node/javascript/src/instance/bindings-smoldot-light.ts
```diff
@@ -37,12 +45,25 @@ export interface Config {
...
+    // Object containing a boolean indicating whether the `killAll` function has been invoked by
+    // the user.
+    const killedTracked = { killed: false };
+
+    const killAll = () => {
+        killedTracked.killed = true;
+        // TODO: kill timers as well?
+        for (const connection in connections) {
+            connections[connection]!.close()
+            delete connections[connection]
```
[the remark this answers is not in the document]
tomaka: That's not the same thing. `connections = {}` changes the object `connections` refers to, and any existing copy of `connections` would still have the connections in it.

bin/wasm-node/javascript/src/instance/instance.ts
tomaka: That code was already there before

bin/wasm-node/javascript/src/client.ts
tomaka: Asserts are never "needed". It's here to detect bugs and provide better error messages. I'm not sure what you're suggesting, but if you're suggesting that it you should be removed, then I'd say no, because it's here to detect a potential bug.

bin/wasm-node/javascript/src/client.ts
tomaka: It enforces that the API be used properly.
The API says "after having called terminate you can't call any other function anymore". This code actually enforces that.

bin/wasm-node/javascript/src/client.ts
wirednkod: Im not suggesting removal - just wondering if this is there intentionally or erroneously.

wirednkod (approved):

Issue #732, opened 2021-04-07
Consider replacing `buffer` JavaScript dependency with manually written JavaScript


tomaka:
The `buffer` dependency is huge, and we only use a small part of its functionalities.

PR #2503, opened 2022-07-13
Remove the `buffer` dependency from the JS bindings

tomaka:
Close https://github.com/paritytech/smoldot/issues/732

The `buffer` library from NPM is 2107 lines of code.
Instead of depending on it, I wrote alternatives in a new `buffer.ts` module (90 lines of code).
The relatively new `TextEncoder`/`TextDecoder` JS APIs are now used, but this shouldn't be a problem: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder#browser_compatibility

This removes a dependency and brings us towards easier Deno compatibility.


melekes (approved):
👍

Issue #2427, opened 2022-06-23
Denoland/x Publishing


harrysolovay:
I propose that––upon merging #2426––we release the Smoldot JS lib to [the Denoland/x registry](https://deno.land/x).

Assuming we want to do so, we should consider two changes:

1. **Committing assets**: Deno users point directly at JS/TS source files. When these files point to / depend on gitignored assets, their execution will fail. For instance, [the `src/worker/instance.ts` file](https://github.com/paritytech/smoldot/blob/11d9e9e2c8503c62d66fd277938160fff784930f/bin/wasm-node/javascript/src/worker/instance.ts#L25) points to `src/worker/autogen/wasm.js` (which is gitignored). We'll need to commit files such as this.
2. **The release process**: Denoland/x relies on GH tags + releases as a source of publish truth. Whenever a release is created, a webhook (specified in the repo settings) notifies Denoland/x of the new tag-specified version. A detailed description can be found [here](https://deno.land/add_module).

tomaka:
These are annoying constraints.

1. The `autogen/wasm.js` file is actually re-generated whenever you run `npm run ...` or `npm publish`. But the file generated by `npm publish` isn't the same as the one generated by `npm start` for example. The file we publish has additional optimizations that very significantly increase the build time.

2. I intentionally don't use tags in this repo because it contains 3 different projects (the library, the full node, and the light client) that each have a different version number.

tomaka:
The easiest solution, I suppose, is to put the "release versions" of the smoldot light client in a separate branch that isn't `main`.
For example, whenever there's a version bump, a bot builds the file, creates a commit, tags it, and pushes it, but the commit isn't on the `main` branch.

harrysolovay:
We could also (A) create a `deno_smoldot` repo & (B) configure this repo with...

- a GH user access token, scoped to have the ability to trigger a workflow in the `deno_smoldot` repo.
- a workflow that (upon merged release-titled PR) triggers the workflow in `deno_smoldot`

The other repo's workflow would tag in accordance with this repo's light client version and create the release, which would trigger the publish to denoland/x.

tomaka:
I've published a version manually, to test: https://deno.land/x/smoldot@light-js-deno-v0.6.20

Trying to use the published version leads to:

```
error: Remote modules are not allowed to import local modules. Consider using a dynamic import instead.
  Importing: node:crypto
[...]

Issue #1802, opened 2021-12-17
wasm-node: use native inflate implementation when available


tomaka:
After https://github.com/paritytech/smoldot/pull/1801, the wasm node decompresses the smoldot wasm bytecode before executing it.
This is done with the `pako` library.

NodeJS has https://nodejs.org/api/zlib.html, but browsers don't have any equivalent yet.
However I see that https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream is in progress.
If one day it is available then we should make use of it.

tomaka:
After https://github.com/paritytech/smoldot/pull/2519 it will be possible to use the NodeJS native API and the Deno native API. Only browsers will need to do the decompression using JS code.

tomaka:
cc https://mozilla.github.io/standards-positions/#compression-streams

PR #2506, opened 2022-07-14
Use tiny_inflate instead of pako

tomaka:
cc https://github.com/paritytech/smoldot/issues/2427 https://github.com/paritytech/smoldot/issues/1802

Removes the `pako` library and uses `tiny-inflate` instead.

The objective of this PR is to remove a dependency, again with the idea of publishing to Deno at some point.
For this reason, the code of `tiny-inflate` (MIT) is directly embedded in the source tree. It is only 376 lines of code.

The `tiny-inflate` claims to be 30% to 50% slower than `pako`.
However, some quick benchmarking on my machine has shown the opposite.
With `pako`, I get 86ms to 88ms for base64 decoding + inflating.
With `tiny-inflate`, I get 78ms to 79ms instead.
The difference might be explained by the fact that I now use "raw" deflate, without any zlib header, as it's the only thing that `tiny-inflate` supports.

### DIFF
--- a/bin/wasm-node/javascript/package-lock.json
+++ b/bin/wasm-node/javascript/package-lock.json
@@ -9,7 +9,6 @@
       "version": "0.6.23",
       "license": "GPL-3.0-or-later WITH Classpath-exception-2.0",
       "dependencies": {
-        "pako": "^2.0.4",
         "websocket": "^1.0.32"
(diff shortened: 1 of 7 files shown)


melekes (approved):
👍

PR #2512, opened 2022-07-15
Add a `blockNumberBytes` field in chain specs

tomaka:
cc https://github.com/paritytech/smoldot/issues/2478

The change is rather straight forward: we add a new optional `blockNumberBytes` field in the chain specs format and pass its value around. The value defaults to 4 if it's missing.

### DIFF
--- a/src/chain_spec.rs
+++ b/src/chain_spec.rs
@@ -263,6 +263,11 @@ impl ChainSpec {
         }
     }
 
+    /// Returns the number of bytes that the "block number" field of various data structures uses.
+    pub fn block_number_bytes(&self) -> u8 {
+        self.client_spec.block_number_bytes.unwrap_or(4)
+    }
+
     /// Returns true if the chain is of a type for which a live network is expected.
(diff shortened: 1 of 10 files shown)


src/chain_spec.rs
```diff
@@ -263,6 +263,11 @@ impl ChainSpec {
         }
     }
 
+    /// Returns the number of bytes that the "block number" field of various data structures uses.
```
melekes, line 266:
```suggestion
    /// Returns the number of bytes that the "block number" field of various data structures uses. If the field is missing, the value defaults to 4.
```

melekes (approved):

src/chain_spec.rs
tomaka: I've intentionally omitted this, because this API is more about obtaining the information that can be computed from the chain spec, rather than getting the content of the chain spec.
If it was the latter, then you'd need to be able to differentiate "4 by default" and "4 as explicitly provided" for example.

PR #2515, opened 2022-07-15
Simplify handling of monotonic_clock_ms

tomaka:
It turns out that `performance.now()` works since NodeJS v8.5 (so, ages ago): https://nodejs.org/api/perf_hooks.html#performancenow

It's unclear since when NodeJS supports `globalThis`, but there seems to be hints that it was working in NodeJS v10 already, which is far away enough.


tomaka:
Well, tests pass locally but not on CI (`globalThis.performance` is `undefined`).
CI uses NodeJS v14.
I suppose that it turns out that this was added only recently.
It's crazy that the docs don't really mention anything about `globalThis.performance` being available.

PR #2517, opened 2022-07-15
Downgrade CI from NodeJS v14 to v12

tomaka:
NodeJS v12 is still widely used, so let's check for compatibility with that version.


tomaka:
It seems that NodeJS 12 doesn't support `await` in `prepare.mjs`

melekes (approved):

wirednkod:
Node v12 is no longer officially supported. We should not adjust code based on that version.

tomaka:
Debian and Ubuntu stable still ship it by default:
https://packages.debian.org/bullseye/nodejs
https://packages.ubuntu.com/jammy/nodejs

wirednkod:
> Debian and Ubuntu stable still ship it by default:
> https://packages.debian.org/bullseye/nodejs
> https://packages.ubuntu.com/jammy/nodejs

Im not sure that this is a good enough reason to keep supporting such versions and roll our code back accordingly

tomaka:
Why not? If there are 5 lines of code to change, then it's worth it.

wirednkod:
> Why not? If there are 5 lines of code to change, then it's worth it.

I just dont find it useful to support a version that is long expired (I think end of lifetime was somewhere at the end of April)

tomaka:
Going to merge.
To me, there's no reason to upgrade NodeJS if necessary. If we could easily support NodeJS v1, we would. It's just a matter of trade-off, as we can't use features that have been introduced in later sessions. Here, we have 5 lines of code to change to support NodeJS 12, so it's worth it.

wirednkod:
Im not strictly against of keeping the older version support. Just added my 2 cents here.

PR #2519, opened 2022-07-17
Move all platform-specific functions in `index-*.ts` files

tomaka:
Close https://github.com/paritytech/smoldot/pull/2516
After trying https://github.com/paritytech/smoldot/pull/2516, I had a eureka moment about how to do this properly.

This PR removes the `compat` module. Instead, all platform-specific functions are now implemented in the main file (formerly `index.ts`) and injected through the public API of `client.ts`.

There are now two `index.ts` files: `index-browser.ts` and `index-nodejs.ts`. The `package.json` file defaults to `index-browser.ts`, but now has a `exports` -> `node` field that instead points to `index-nodejs.ts`.

In a future PR, I will add a `index-deno.ts` file that will be the Deno entry point.


melekes (approved):

wirednkod (approved):
lgtm

PR #2522, opened 2022-07-18
Add support for Deno with an index-deno.ts

tomaka:
Adds a new `index-deno.ts` file, similar to `index-nodejs.ts` and `index-browser.ts`.

This file gets compiled by TypeScript, which generates `dist/mjs/index-deno.js`.
The content of `dist/mjs` can then be published to the deno.land/x registry.

The reason for not publishing directly the content of `src` on the Deno registry is that Deno requires paths to modules to always end with a `.ts` extension, while regular TypeScript requires paths to end with either no extension or with `.js`.
It is therefore impossible to write code that works on both non-Deno TypeScript and Deno TypeScript (unless that code is squashed together within a single module, which I've considered, but this isn't worth the trade-off).

Because the Deno types aren't available on NPM (well, they actually [are](https://www.npmjs.com/package/deno-types), but not maintained), I had to copy-paste the types found in the Deno repository.
Considering that we are only using stable APIs, I expect these types to never need to be updated. Fingers crossed that I will not be proven wrong.

We now run `deno run ./dist/mjs/index-deno.js` in the CI, which should ensure that compatibility never breaks.
I've also added a `demo/demo-deno.ts` file, which isn't tested by CI but can be run from Deno in order to make sure that everything is working.


melekes (approved):

harrysolovay (approved):

tomaka:
Should be merged together with https://github.com/paritytech/smoldot/pull/2533

Issue #2521, opened 2022-07-18
Panic in Deno demo when closing JSON-RPC client


tomaka:
> Smoldot has panicked while executing task `json-rpc-messages-out`. This is a bug in smoldot. Please open an issue at https://github.com/paritytech/smoldot/issues with the following message:
panicked at 'called `Result::unwrap()` on an `Err` value: "WouldBlock"', bin/wasm-node/rust/src/init.rs:258:57

Happens in the Deno demo which I haven't uploaded yet, but it will come in a PR soon.

tomaka:
Happens because I didn't call `chain.remove` when the JSON-RPC client disconnects.

PR #2527, opened 2022-07-19
Absorb exceptions thrown in the JSON-RPC callback

tomaka:
Close https://github.com/paritytech/smoldot/issues/2521

This was an overlook. If the user uses `throw` in a callback (or, more likely, something accidentally throws), the exception will propagate through the smoldot VM and act as if smoldot had crashed.
We don't want that to happen.


tomaka:
Note that I'm absorbing exceptions at this level rather than in the callback called by the Wasm VM, because if a callback called by the Wasm VM throws an exception it indicates a bug in the smoldot JS code, in which case it isn't wrong to crash smoldot.

melekes (approved):
👍

Issue #933, opened 2021-06-14
Add support for GrandPa warp sync for Aura-based chains


tomaka:
The functions that are necessary exist in the Aura pallet: https://github.com/paritytech/substrate/blob/a0161f3b224a0607b351ae23b41621b01126a893/primitives/consensus/aura/src/lib.rs#L86-L93

It's just a matter of trying Aura in the warp sync code if the Babe functions are missing.

bernardoaraujor:
@tomaka can you give me some directions so I can try to implement this feature?

tomaka:
https://github.com/paritytech/smoldot/pull/2578 was more or less a pre-requisite and is now done

PR #2530, opened 2022-07-19
Only proceed with warp sync process if chain is using Babe

tomaka:
cc https://github.com/paritytech/smoldot/issues/933

At the moment, the warp syncing algorithm downloads the runtime, then uses it to fetch the Babe configuration. The algorithm just assumes that the chain is using Babe.
This PR tweaks the code to only proceed if the chain actually uses Babe as its consensus algorithm.
If the chain uses Aura or "Unknown", an error is returned.

I initially wanted to do https://github.com/paritytech/smoldot/issues/933 at once, but the changes are deeper than I thought, and I'll need to do some refactoring beforehand.


melekes (approved):
👍

Issue #2523, opened 2022-07-18
JS demo returns `Error while adding chain: Error: Data error`


wirednkod:
When running the demo, it is always returned:
```
Error while adding chain: Error: Data error
```
This seems to be coming from the `tiny-inflate` and specifically from [here](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/javascript/src/tiny-inflate/index.js#L330) that returns an error due to [invalid length](https://github.com/paritytech/smoldot/blob/main/bin/wasm-node/javascript/src/tiny-inflate/index.js#L298-L300)

tomaka:
This should only happen if you had cloned the repo before https://github.com/paritytech/smoldot/pull/2506 and didn't run `prepare.mjs` since then

tomaka:
Managed to reproduce. Have no idea what the problem is. Decoding the base64 and re-inflating the buffer within `prepare.mjs` using `zlib` works fine.

I'll restore pako after https://github.com/paritytech/smoldot/pull/2522

PR #2533, opened 2022-07-19
Fix base64 decoding function

tomaka:
It turns out that the cause of https://github.com/paritytech/smoldot/issues/2523 was a bug in the base64 decoding code.
This PR replaces this code with something that seems to work.


wirednkod (approved):

tomaka:
Tested in both NodeJS and browsers.

PR #2545, opened 2022-07-21
Move Kademlia protocol functions to protocol module

tomaka:
It's part of the protocol that the node uses, so let's move these functions to `protocol`.


melekes (approved):
👍

PR #2546, opened 2022-07-21
Adjust JS API to support multi-stream connections

tomaka:

Adjusts the JS code that binds the `index-*` files to the Wasm VM that runs smoldot.
It now supports multi-stream connections.


melekes (approved):
👍

Issue #1107, opened 2021-08-09
Backport "Embed version (or other metadata) into runtime wasm blob"


tomaka:
https://github.com/paritytech/substrate/pull/8688

tomaka:
After some discussion, it seems that the reason why `apis` is empty is that there is also a `runtime_apis` section containing just the APIs.

PR #2556, opened 2022-07-25
Backport runtime spec embedded in Wasm module

tomaka:
Fix https://github.com/paritytech/smoldot/issues/1107

Rather than execute an expensive runtime call, it is possible to get this information from a custom Wasm section.

The objective of this PR isn't really the speed boost brought by reading the Wasm module, but the fact that we now have a `runtime_version() -> &CoreVersion` method on `HostVmPrototype`.
In other words, there is now a low-cost guaranteed way of knowing the version of a runtime.

This makes various other changes that require knowing the runtime version easier.

### DIFF
--- a/bin/light-base/src/runtime_service.rs
+++ b/bin/light-base/src/runtime_service.rs
@@ -918,9 +918,6 @@ pub enum RuntimeError {
     /// Error while compiling the runtime.
     #[display(fmt = "{}", _0)]
     Build(executor::host::NewErr),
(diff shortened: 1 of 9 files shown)


tomaka:
I've considered adding an entry in the CHANGELOG, but considering that it's a Substrate backport, and a low level detail, and that it probably won't change anything for users, I don't think it makes sense to mention it.

melekes (approved):
👍

PR #2558, opened 2022-07-26
Also parse `runtime_apis` custom section

tomaka:
Fix https://github.com/paritytech/smoldot/issues/1107

This is an amendment to https://github.com/paritytech/smoldot/pull/2556

After asking on Element why the `apis` field was empty when obtaining the runtime version from the custom section, it has been pointed out that for some reason they're in a separate section named `runtime_apis`.

The APIs decoded from the `runtime_apis` section must then be inserted into the struct that was decoded from the `runtime_version` section.
It's very weird, and the person I talked to (@pepyakin) seems to agree that this is weird.

### DIFF
--- a/src/executor.rs
+++ b/src/executor.rs
@@ -215,7 +215,7 @@ impl<'a> CoreVersionRef<'a> {
         out.extend(self.spec_version.to_le_bytes());
         out.extend(self.impl_version.to_le_bytes());
 
-        out.extend(crate::util::encode_scale_compact_usize(self.apis.clone().len()).as_ref());
+        out.extend(crate::util::encode_scale_compact_usize(self.apis.clone().count()).as_ref());
         for api in self.apis.clone() {
             out.extend(api.name_hash);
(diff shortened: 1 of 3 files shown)


src/executor.rs
```diff
@@ -198,13 +198,69 @@ pub struct CoreVersionRef<'a> {
     pub state_version: Option<u8>,
 }
 
+impl<'a> CoreVersionRef<'a> {
+    /// Returns the SCALE encoding of this data structure.
+    pub fn scale_encoding_vec(&self) -> Vec<u8> {
+        let mut out = Vec::<u8>::with_capacity(
+            1 + self.spec_name.len() + 1 + self.impl_name.len() + 4 + 4 + 4 + 32 + 4 + 1,
```
melekes, line 205: where 4, 32, 1 are coming from? should we maybe add a spec link or document?

tomaka: It's the lengths of the various fields right below.

Issue #2564, opened 2022-07-28
Path to smoldot on win


petersopko:
While trying to build our app, with pnpm, I'm running into problem on windows machine (works just fine on MacOS / WSL instance) From what I gathered, the path to this package is unknown on win? 

Thanks for taking time to review this, got stuck with this for a while now

```
 ERROR  in ./node_modules/@substrate/smoldot-light/dist/mjs/index-browser.js                                                                                                                                                       friendly-errors 18:07:05  

Module parse failed: Unexpected token (38:43)                                                                                                                                                                                      friendly-errors 18:07:05  
File was processed with these loaders:
 * ./node_modules/unplugin/dist/webpack/loaders/transform.js
 * ./node_modules/@open-wc/webpack-import-meta-loader/webpack-import-meta-loader.js
You may need an additional loader to handle the result of these loaders.
|         },
|         connect: (config) => {
>             return connect(config, options?.forbidWs || false, options?.forbidNonLocalWs || false, options?.forbidWss || false);
|         }
|     });
 @ ./node_modules/@polkadot/rpc-provider/substrate-connect/index.js
 @ ./node_modules/@polkadot/rpc-provider/bundle.js
ang=ts&
 @ ./components/shared/format/Identity.vue?vue&type=script&lang=ts&
 @ ./components/shared/format/Identity.vue
 @ ./node_modules/unplugin/dist/webpack/loaders/transform.js??ref--44-0!./node_modules/unplugin/dist/webpack/loaders/transform.js??ref--46-0!./node_modules/babel-loader/lib??ref--0-0!./node_modules/@nuxt/webpack-edge/node_modules/vue-loader/lib??vue-loader-options!./node_modules/unplugin-vue2-script-setup/node_modules/unplugin/dist/webpack/loaders/transform.js??ref--19-0!./node_modules/unplugin/dist/webpack/loaders/transform.js??ref--45-0!./pages/westmint/u/_id.vue?vue&type=script&lang=ts&      
 @ ./pages/westmint/u/_id.vue?vue&type=script&lang=ts&
 @ ./pages/westmint/u/_id.vue
 @ ./.nuxt/router.js
 @ ./.nuxt/index.js
 @ ./.nuxt/client.js
 @ multi ./node_modules/@nuxt/bridge/dist/runtime/capi.plugin.mjs ./node_modules/eventsource-polyfill/dist/browserify-eventsoang=ts&
[...]

PR #2565, opened 2022-07-28
Generate code for the ES6 target

tomaka:
Close https://github.com/paritytech/smoldot/issues/2564


melekes (approved):

petersopko:
thanks for the fix guys, when can we expect the release on this? We've added checks for win machines and this seems to be universal problem in win builds https://github.com/kodadot/nft-gallery/pull/3603

PR #2116, opened 2022-03-03
change grandpa name and use fork_id

melekes:
Closes #1955

PR #2569, opened 2022-07-29
Remove unused fallback protocol names system

tomaka:
Ages ago, I added a dummy field called `fallback_protocol_names` that lets you specify a list of alternative protocol names that would be negotiated if the main one isn't supported by the remote.

Substrate has a similar feature, and it's very useful in order to do protocol upgrades. (e.g. https://github.com/paritytech/smoldot/pull/2116)

Unfortunately this has never been implemented.
Instead of leaving this dummy feature here, this PR just removes it. It can always be added back properly in the future.

### DIFF
--- a/src/libp2p/collection.rs
+++ b/src/libp2p/collection.rs
@@ -105,12 +105,6 @@ pub struct NotificationProtocolConfig {
     /// Name of the protocol negotiated on the wire.
     pub protocol_name: String,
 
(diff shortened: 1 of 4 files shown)


tomaka:
It turns out that this way even mis-implemented.
If a fallback name was passed, we would accept incoming substreams of that protocol then panic.

melekes (approved):

Issue #2080, opened 2022-02-20
Use a call proof in the Grandpa warp syncing instead of multiple storage proofs


tomaka:
At the moment, the warp syncing code calls the runtime, then sends one storage proof every time the runtime accesses the storage.

This means a lot of redundancy in the proofs, but more importantly a lot of network round trips.
Right now on a simple dev chain we do 12 round trips just for that. If the ping is 100ms, that means it takes 1.2 seconds instead of only 100ms.

tomaka:
For context, the reason why it was done with simple storage proofs is that I expected one or two storage items to be accessed, rather than 12.

tomaka:
At the moment, warp syncing Westend (with a moderately old checkpoint) on my local machine takes about 2.5 to 5 seconds.
In my work-in-progress branch, this is now down to 1.5 seconds.

Issue #2529, opened 2022-07-19
Properly handle warp syncing failures


tomaka:
The warp syncing process can fail for various reasons, and currently we just try again. This is a poor idea because all these reasons are "deterministic", in the sense that trying again will always lead to the same error.

What to do in that situation isn't obvious, but I think that the least we can do is print a proper error and stop trying.

tomaka:
> all these reasons are "deterministic", in the sense that trying again will always lead to the same error.

This isn't actually totally true.
For example, the runtime missing the function that we want to call is a pretty terrific/irrecoverable problem, but it could also be caused by a mistake in the runtime, and it could be that in the future this mistake gets corrected on the chain.
In other words, continuing to sync could in principle lead to a recovery.

I think that the behavior of trying again is correct, but we should probably add a delay (in number of blocks) between the syncing attempts.
Additionally, we should probably indicate through logs if this happens (cc https://github.com/paritytech/smoldot/issues/2222).

Issue #2222, opened 2022-04-05
Print more details about the warp syncing while it is in progress


tomaka:
As evidenced by https://github.com/paritytech/substrate-connect/issues/927, it happens that warp sync just takes ages. Right now we print a periodic warning saying "don't worry, we're still alive". This message should include some information or statistics about the progress of the warp syncing.

melekes:
Are there any specific data points that we want to output?

tomaka:
The warp syncing consists in multiple stages: the first one (normally the longest) consists in downloading and verifying the fragments to do the actual syncing, and the rest consists in some extra network requests to download the state at the head of the chain (without which we couldn't progress from there).

I think that we should print the stage we're at, and for the download stage print the block number of the latest fragment that we've downloaded.

tomaka:
Note that the warp sync code at the moment downloads all the fragments (this takes multiple requests) then verifies them all at once.

It would be better if fragments were verified progressively (verify the current batch of fragments while the next batch is being downloaded), both to speed things up but also to show with certainty the block number that we've synced to.

PR #2578, opened 2022-08-04
Rewrite the warp syncing code to use call proofs

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2080

cc https://github.com/paritytech/smoldot/issues/933 https://github.com/paritytech/smoldot/issues/2529 https://github.com/paritytech/smoldot/issues/2222

The initial intention behind this PR was to make it use call proofs rather than individual storage gets. See https://github.com/paritytech/smoldot/issues/2080.

However, this turned into a bigger refactoring as I've realized that a bigger refactoring is very much needed and would simplify the code a lot.

The warp syncing code now uses an API similar to `all_forks` and to `optimistic`, the two other syncing strategies.
The warp syncing code was the first syncing strategy to be written in smoldot, and at that time we didn't have the experience of the caveats of its approaches.
I don't think there's much to comment, given that the API is now extremely close to what already exists.

Apart from using call proofs, I've opted to keep the exact same behavior as today, even though it is a bit questionable. In particular, when we download fragments from a peer, we then always query the same peer for the storage items. If that peer disconnects, we try download fragments again. As mentioned, I've kept this behavior, but the refactoring will make it much easier to improve compared to before this PR.

Can be reviewed commit by commit if desired. I've taken an incremental approach, and everything still works after almost every commit.


tomaka:
The biggest change outside of `warp_sync.rs` is probably the duplication of the `RequestDetail` struct in two: `RequestDetail` and `DesiredRequest`.
`DesiredRequest` is what the syncing state machine returns, while `RequestDetail` is what the API user passes as parameter.

The only difference is in `RequestDetail::StorageGet` vs `DesiredRequest::StorageGet`. The `DesiredRequest` one contains an extra `state_root_hash` field that isn't in the `RequestDetail` one. That's because this field is redundant with the `block_hash` field that is already present.

It makes sense for the syncing state machine to provide both a `block_hash` and a `state_root_hash`, but it is weird for the API user to pass both, because then we have to deal with situations where the `block_hash` doesn't correspond to the `state_root_hash`.

melekes (approved):

src/sync/all.rs
```diff
@@ -1388,134 +1473,103 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
...
-                }
+                let user_data = sync.fail_request(request_id).user_data;
+                // TODO: notify user of the problem
+                self.inner = AllSyncInner::GrandpaWarpSync { inner: sync };
+                (user_data, ResponseOutcome::Queued)
             }
+            // Only the GrandPa warp syncing ever starts GrandPa warp sync requests.
+            (other, _, RequestMapping::Inline(_, _, user_data)) => {
+                self.inner = other;
+                (user_data, ResponseOutcome::Queued) // TODO: no
```
melekes, line 1511: what does "no" mean in this context?

tomaka: I think this appears in the diff only because I changed the indentation level, but what I mean is that `Queued` is wrong.
We simply discard the request result and not queue it in any way. `Queued` is kind of used as the default, but ideally it would be something more precise. But because this is just for informative purposes, I haven't fix it properly.

PR #2579, opened 2022-08-04
Add support for WebRTC

### LINKED ISSUE #1712: wasm-node: Add support for WebRTC

melekes:
Closes #1712


bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -121,23 +131,213 @@ function trustedBase64Decode(base64: string): Uint8Array {
...
+
+        const ipVersion = webRTCParsed[1] == 'ip4'? '4' : '6';
+        const targetIp = webRTCParsed[2];
+        const ufrag = (webRTCParsed[5] || '');
+
+        // Transform ufrag (multibase-encoded multihash) to fingerprint (upper-hex;
+        // each byte separated by ":").
+        //
+        // To add additional decoders use `or`: `base64url.decoder.or(base32.decoder)`.
+        const fingerprint = new Uint8Array(base64url.decode(ufrag)).join(':').toUpperCase();
```
tomaka, line 186: Since it's a multibase, the base64-encoded multihash-encoded certificate hash starts with a letter (I suppose 'm', 'M', 'u', or 'U').
See https://github.com/multiformats/multibase#multibase-table

Also, once base64-decoded, you still need to remove one byte for the sha-256 multihash prefix (should be 0x12) and one byte for the size (0x20).
See https://github.com/multiformats/multihash#format

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -121,23 +125,200 @@ function trustedBase64Decode(base64: string): Uint8Array {
...
+      },
+
+      send: (data: Uint8Array, streamId: number): void => {
+        const channel = dataChannels.get(streamId)!;
+        console.log("sent", data);
+        channel.send(data);
+      },
+
+      openOutSubstream: () => {
+        addChannel(pc.createDataChannel("data", { id: outId, negotiated: true }), 'outbound')
```
melekes, line 317: only the first ("noise") data channel is negotiated and has predefined ID=1. other channels don't need neither of these things.

tomaka:
Should be ready to merge in my opinion.

tomaka (approved):

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -121,23 +113,294 @@ function trustedBase64Decode(base64: string): Uint8Array {
...
+
+      dataChannels.set(dataChannelId, dataChannel);
+    }
+
+    // It is possible for the browser to use multiple different certificates.
+    // In order for our local certificate to be deterministic, we need to generate it manually and
+    // set it explicitly as part of the configuration.
+    // According to <https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-generatecertificate>,
+    // browsers are guaranteed to support `{ name: "ECDSA", namedCurve: "P-256" }`.
+    RTCPeerConnection.generateCertificate({ name: "ECDSA", namedCurve: "P-256" } as EcKeyGenParams).then((localCertificate) => {
```
melekes, line 186: I think we can specify a hash algorithm here. See https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/generateCertificate#specifying_algorithm_details

that way we won't need the for loop below

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -121,23 +113,294 @@ function trustedBase64Decode(base64: string): Uint8Array {
...
+          for (const channel of Array.from(dataChannels.values())) {
+            channel.onopen = null;
+            channel.onerror = null;
+            channel.onclose = null;
+            channel.onmessage = null;
+          }
+
+          pc!.close();  // Unclear whether this is necessary, but it doesn't hurt to do so.
+
+          for (const channel of Array.from(dataChannels.values())) {
```
melekes, line 232: this maybe redundant. https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/close

"Calling this method terminates the RTCPeerConnection's ICE agent, ending any ongoing ICE processing and **any active streams.**"

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -121,23 +113,294 @@ function trustedBase64Decode(base64: string): Uint8Array {
...
+          dataChannels.clear();
+        }
+      };
+
+      pc.onnegotiationneeded = async (_event) => {
+          // Create a new offer and set it as local description.
+          let sdpOffer = (await pc!.createOffer()).sdp!;
+          // The server excepts the ufrag and pwd to be the same. Modify the local description
+          // to ensure that.
+          // TODO: why does the server want that?
```
melekes, line 244: should we link the spec here?

"_A_ generates a random string and sets that string as the username (_ufrag_
   or _username fragment_) and password on the SDP of the local offer."
   
 the answer to this question is: otherwise there's no way to learn `pwd` for B (server). Right now B receives ICE binding request, which contains ufrag (but no pwd). Since we've agreed that ufrag == pwd, B uses this to construct SDP locally.

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -121,23 +113,294 @@ function trustedBase64Decode(base64: string): Uint8Array {
...
+          for (const channel of Array.from(dataChannels.values())) {
+            channel.onopen = null;
+            channel.onerror = null;
+            channel.onclose = null;
+            channel.onmessage = null;
+          }
+
+          pc.close();
+
+          for (const channel of Array.from(dataChannels.values())) {
```
melekes, line 361: same comment as above

melekes (commented):
LGTM

PR #2581, opened 2022-08-05
Implement Aura+Grandpa warp syncing

tomaka:
Close https://github.com/paritytech/smoldot/issues/933

It currently only works for GrandPa+Babe.
This PR changes it to support Aura.

It is done simply by querying Aura-related parameters instead of Babe-related parameters.


tomaka:
Confirmed working on substrate-node-template.

PR #2589, opened 2022-08-07
Replace `websocket` dependency with `ws`

### LINKED ISSUE #2588: Nodejs issue
```
 console.warn
    [connections] Failed to reach bootnode 12D3KooWLn22TSPR3HXMRSSmWoK4pkDtspdCVi5j86QyyUNViDeL through /dns/kusama-connect-2.parity.io/tcp/443/wss: TypeError: Cannot read properties of undefined (reading 'w3cwebsocket')
```

https://unpkg.com/browse/@substrate/smoldot-light@0.6.27/dist/cjs/index-nodejs.js Line 81 is 

```
socket: new websocket_1.default.w3cwebsocket(url)
```

It should be `new websocket_1.w3cwebsocket(url)`

So something is wrong with the build step.

tomaka:
Fix #2588 

The `websocket` package on NPM is just kind of fucked up, and it is impossible to make it work on both ModuleJS and CommonJS. This PR switches to `ws`, which works fine.
This was previously not possible because we were using the `websocket` library in the browser as well. This is no longer the case.

The biggest effort was updating the demo.

Additionally, I've made some tweaks to the `tsconfig.json` to disable some magic features. I removed them as a first attempt at understanding the issue. They turned out to be unrelated, but they're better left disabled anyway.

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -9,6 +9,10 @@
 - The GRANDPA warp sync implementation has been considerably refactored. It is possible that unintended changes in behaviour have accidentally been introduced. ([#2578](https://github.com/paritytech/smoldot/pull/2578))
 - A warning is now printed if the `badBlocks` field in a chain specification is not empty. Bad blocks are not supported by the smoldot light client. ([#2585](https://github.com/paritytech/smoldot/pull/2585))
 
+### Fix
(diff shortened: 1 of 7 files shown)


tomaka:
Since this PR makes a small tweak to `index-browser.ts`, I've tested this change in the browser, and it works.

PR #2634, opened 2022-08-15
Do another pass to fix the consistency of `addresses`

tomaka:
This PR improves the consistency of the list of addresses of peers.
We now properly take into account only addresses of outgoing connections, and handle situations where we dial an address and find a different peer than what was expected.


melekes (commented):
👍

PR #2658, opened 2022-08-18
No longer verify commit messages immediately

tomaka:
When a commit message is received from the network, we now store it for later verification, as intended.
This should also fix issues where the sync service was reporting the same block as finalized multiple times.


melekes (commented):
👍

PR #2659, opened 2022-08-18
Add per-block finality info in blocks_tree

tomaka:
This PR makes all the finality-related operations in the `BlocksTree` O(1).

Before this PR, we sometimes needed to traverse the tree of blocks in order to find an information concerning finality.
After this PR, this information is now stored in each block whenever it is inserted.
In order to not sure too much information, the list of authorities has been put in an `Arc`, meaning that it can be duplicated between all blocks that use the same list relatively cheaply.

I must admit that GrandPa is quite complicated, and I've re-read what I've done multiple times, to be sure to not mix scheduling block, triggering block, not have an off-by-one problem with scheduled changes, etc.


tomaka:
Seems to work in all situations I've tested.

melekes (commented):
👍

PR #2663, opened 2022-08-18
Implement TrieStructure::remove_prefix

tomaka:
Before this PR, whenever we want to clear all the nodes with a prefix, we simply clear the entire Merkle root calculation cache.
After this PR, we only clear from the cache the entries that have the given prefix.
The idea is to speed up blocks verification (and authorship).

Tests need to be written before merging.

Furthermore, I've discovered that the `descendants` function in this module doesn't look right at all, which would make tests pass more than they should. This should be investigated beforehand as well.

### DIFF
--- a/src/trie/calculate_root.rs
+++ b/src/trie/calculate_root.rs
@@ -182,17 +182,13 @@ impl CalculationCache {
 
     /// Notify the cache that all the storage values whose key start with the given prefix have
     /// been removed.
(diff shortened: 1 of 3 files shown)


tomaka:
Test still failing occasionally.

tomaka:
Test no longer fails after commit I just pushed.

melekes (commented):
👍

PR #2664, opened 2022-08-18
Properly iterate over the trie structure

tomaka:
cc my remark in https://github.com/paritytech/smoldot/pull/2663

The `descendants` and `all_nodes_ordered` functions were misimplemented, meaning that `structure_equal` would return `true` even when the structures were not necessarily equal.

I've also added a test, and this test indeed fails before the bugfix is applied.


melekes (commented):
👍

PR #2666, opened 2022-08-19
Test prefix removal in the trie cache

tomaka:
Tweaks the trie cache test in order to test prefix removal operations.

As I'm opening this PR, the test is failing, and it clearly indicates a bug in the trie cache.
Needs fixing.


melekes (approved):
👍

Issue #2671, opened 2022-08-22
embeddable native light-client Rust library


rphmeier:
It'd be useful to have some kind of Rust library in this repo which you can link in, pass a chain-spec and some operating parameters, and get an RPC object back. 

Is this something we want in the smoldot repo? Should it live somewhere else?

What architecture would you recommend for this? The bin/wasm-node seems to be a good starting point. 

What should its API be?

tomaka:
If you want to integrate a light client into a Rust application, the library you're looking for is `smoldot-light-base`: https://github.com/paritytech/smoldot/tree/main/bin/light-base

You just have to implement the `Platform` trait found here: https://github.com/paritytech/smoldot/blob/2f1c8e62261c47da01db834b8e6629a528e70b3b/bin/light-base/src/lib.rs#L91

I believe that this is the least opinionated API possible. Anything higher level (for example, sockets handled automatically) and you might run into blockers.

Not that I'm not sure how easy this trait is to implement on a desktop platform.

cc #2542

rphmeier:
I've had a look at the light-base but it's hardly plug-and-play. I guess I'd be looking for something like a `Platform` implementation for a desktop environment which could be combined with the light-base here.

rphmeier:
I'd be happy to contribute an example to the repo of implementing the `Platform` trait for the desktop environment - would `Connection` be a `tokio::TcpStream` in that scenario?

tomaka:
`Stream` would be a `TcpStream`, and `Connection` can be anything (`!` would be the most appropriate, but since it's not stable you can use `()`).

I think the best approach is to make the trait as easy to implement as possible, and then have everyone either implement the trait themselves or copy-paste an existing example.

rphmeier:
Is the platform trait expected to change much in the future? It'd be nice to have more developer resources but we could put something in the substrate.dev site instead.

arrudagates:
> I'd prefer to not add a demo implementation as code in this repo, unless we go for maintaining a proper desktop light client.

As a potential user of a smoldot desktop light client I do think it's desirable to have something like that, unless you want to implement it in a separate repo (like substrate-connect), but I do think we need something that can be easily plugged into subxt like substrate-connect can be plugged into polkadot.js api

rphmeier:
cc https://github.com/paritytech/subxt/issues/596

Issue #2542, opened 2022-07-20
Documentation for a light client completely in Rust


arrudagates:
I've been trying for quite a while to get light_base working without the JavaScript side, but my knowledge of JavaScript is not good enough to port that code over to the rust side of the example. If possible, I'd greatly appreciate an example like that without the JavaScript side and completely in Rust, as it's not just for web apps that light clients can be used for!

tomaka:
The documentation of the `Platform` trait are normally good enough that you shouldn't have to read the JS code.
Feel free to point out things that you don't understand in that documentation.

arrudagates:
> The documentation of the `Platform` trait are normally good enough that you shouldn't have to read the JS code. Feel free to point out things that you don't understand in that documentation.

I've tried my best to understand, but I have no idea what those functions should be doing and what those types should be set to. Can you just provide a full rust example?

gilescope:
I've had a go and am also a bit perplexed. I had a go: https://github.com/gilescope/dotsamaworld/commit/22566351f594168e51addc7042867423be063d22

I could construct something that compiled, but once I had added my chain and got a chain_id I was uncertain what to do next. Can I ask something for a block or what the latest block number is? An example would make things clearer as to how I can use smoldot once I have constructed it. If anyone else has tried calling smoldot from rust would like to see that code.

gilescope:
Also please let me know if that's definitely NOT the right way to be constructing things - I did not have much to go on and am just following the compiler...

tomaka:
> I've had a go and am also a bit perplexed. I had a go: https://github.com/gilescope/dotsamaworld/commit/22566351f594168e51addc7042867423be063d22

Your `new_task_rx` variable will then receive tasks that must be executed until they finished. You can be implemented basically like that:

```rust
tokio::spawn(async move {
    let mut all_tasks = stream::FuturesUnordered::new();
    loop {
        futures::select! {
            (_, new_task) = new_task_rx.select_next_some() => {
                all_tasks.push(new_task);
            },
            () = all_tasks.select_next_some() => {},
        }
    }
})
```

> Can I ask something for a block or what the latest block number is?

Well, all you need to do is call `Client::json_rpc_request`, passing as parameter the same format as regular JSON-RPC requests.
https://github.com/paritytech/smoldot/blob/2f8d6a270ce8c54cc66e7165a02e9cf537668892/bin/light-base/src/lib.rs#L1020

The responses will be sent to the channel you passed as part of `addChain`.

[...]

PR #2675, opened 2022-08-23
Turn `smoldot-light-base` into a properly usable library

tomaka:
Close https://github.com/paritytech/smoldot/issues/2671
Close https://github.com/paritytech/smoldot/issues/2542

This PR does the following:

- Move parameters provided to `Client::new` to a `ClientConfig` struct.
- Swap the two template parameters of `Client`.
- Make the `smoldot-light-base` properly no-std-friendly.
- Move the `Platform` trait in a separate module.
- Add a `std` feature to `smoldot-light-base` enabled by default.
- Add an implementation of the `Platform` trait based on `async-std`, gated behind the `std` feature. (see remarks below)
- Rename `smoldot-light-base` to `smoldot-light`.
- Add a lot of documentation.

The objective of this PR is to turn this library into a properly usable library that can be embedded in other Rust applications.

I'm aware that the example currently doesn't work. This is because the example doesn't support WebSocket Secure connections, and that almost all the non-WSS nodes seem to be full.
It seems, however, that there are problems other than nodes being full. Smoldot seems to try the same nodes over and over again.
I will investigate this, but the fix for that problem is out of scope of this PR. As far as I can tell, the code in this PR works as intended.

Some remarks about `async-std`: I know there's an active question of `async-std` vs `tokio`, and I want to pro-actively address his. The trait implementation provided by this PR uses `async-std` because it just works in all situations. This would not be the case with `tokio`. The objective of this trait implementation is specifically to be "plug and plug". In order to comply with this objective, only `async-std` makes sense. I will **not** add a separate trait implementation for `tokio`. Please realize that there's a trade-off between simplicity and performance, and in the case of async-std vs tokio, the balance weights very heavily in favor of keeping it simple by having only one trait implementation. In other words, I really don't care about saving 2% of CPU if it bloats the library's code. I have nothing against `tokio`, but if you want to use `tokio`, just reimplement the trait yourself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,8 @@ jobs:
     - run: RUSTFLAGS=-Dwarnings cargo check --package smoldot --locked --no-default-features --features database-sqlite
     - run: RUSTFLAGS=-Dwarnings cargo check --package smoldot --locked --no-default-features --features std
     - run: RUSTFLAGS=-Dwarnings cargo check --package smoldot --locked --no-default-features --features database-sqlite --features std
+    - run: RUSTFLAGS=-Dwarnings cargo check --package smoldot-light-base --locked --no-default-features
+    - run: RUSTFLAGS=-Dwarnings cargo check --package smoldot-light-base --locked --no-default-features --features std
 
   fuzzing-binaries-compile:
     runs-on: ubuntu-latest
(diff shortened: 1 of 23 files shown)


tomaka:
I've just uploaded `smoldot-light` v0.1.0 on crates.io, despite the example not working, in order to be sure to have the name. A new version shall be published when everything is working properly.

PR #2676, opened 2022-08-23
Publish v0.3.0 of the library

tomaka:


Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 edition = "2021"
 readme = "README.md"
 keywords = ["blockchain", "peer-to-peer"]
+exclude = [".github", "bin"]  # TODO: we might need to reorganize the repo to avoid including unwanted files
```
tomaka, line 12: `crates.io` refused my package because it was over the size limit, which is how I realized that we were accidentally uploading chain specs as well.

PR #2679, opened 2022-08-24
Pin GHA versions

sergejparity:
In order to improve our security posture with GitHub Actions usage. I've made a version pinning ether to commit hash or to specific version.

Related issues and policy:
https://github.com/paritytech/ci_cd/issues/464
https://github.com/paritytech/ci_cd/wiki/Policies-and-regulations:-GitHub-Actions-usage-policies

### DIFF
--- a/.github/workflows/ci-post-workflow.yml
+++ b/.github/workflows/ci-post-workflow.yml
@@ -20,7 +20,7 @@ jobs:
     if: ${{ github.event.workflow_run.event == 'pull_request' }}
     steps:
       # TODO: use download-artifact action
-      - uses: actions/github-script@v3.1.0
+      - uses: actions/github-script@v6.1.1
         with:
           script: |
             var artifacts = await github.actions.listWorkflowRunArtifacts({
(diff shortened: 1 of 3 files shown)


alvicsam (approved):

sergejparity:
No problem. I'll leave hanging for now.

.github/workflows/ci-post-workflow.yml
```diff
@@ -60,13 +60,13 @@ jobs:
             const fs = require('fs');
             return fs.readFileSync('./number', 'utf8').toString();
           result-encoding: string
-      - uses: peter-evans/find-comment@v2
+      - uses: peter-evans/find-comment@1769778a0c5bd330272d749d12c036d65e70d39d # v2.0.0
```
melekes, line 63: just curious: why not pin to `v2.0.0`?

```suggestion
      - uses: peter-evans/find-comment@v2.0.0
```

sergejparity: It's pretty simple. When you point to the tag, then in case if GHA repo will get compromised (which is out of our control), tag can be easily moved to another commit, branch etc. And we will not even get notified about such events.
With the hash such things will not work.
Yes, it might be not so convenient, but it is an eternal trade off - security vs convenience.
Also dependabot can track GHA updates by the hash as well.

PR #2681, opened 2022-08-24
Set the permissions of GitHub actions

tomaka:
Close https://github.com/paritytech/smoldot/pull/2679
I'm opening this as an alternative to https://github.com/paritytech/smoldot/pull/2679

Rather than pinning actions to a certain commit, we just prevent actions from doing anything problematic.

If you're ok with that approach, I'll fix the TODOs that I've left in the PR.


sergejparity:
Actually I would prefer to pin GHA's as well for sake of uniformity of how they are handled throughout an org. Your change of permissions will greatly compliments that as well.

tomaka:
What the advantage of having uniformity through the org?

Also, I don't understand the idea behind pinning commit hashes. Does that mean that I can never ever update these specific actions?

sergejparity:
There is no problem with action updates. Dependabot also is working with commit hashes matching them with new releases.

What goes to uniformity we were requested by security team to standardize the approach how GHA's defined in our repos. Based on this https://forum.parity.io/t/github-actions-gha-versioning-updates-and-security/953/11 conversation the policy was created https://github.com/paritytech/ci_cd/wiki/Policies-and-regulations:-GitHub-Actions-usage-policies

tomaka:
> There is no problem with action updates. Dependabot also is working with commit hashes matching them with new releases.

What's the point of pinning by hash if we let dependabot just update these hashes liberally? Are we supposed to review the code of every action every time a dependabot PR wants to update it?

sergejparity:
> What's the point of pinning by hash

Hash is hard to counterfeit. Tag can be easily moved to another commit and nobody even notice it.

tomaka:
I understand the advantage of hash as opposed to version, what I'm saying is that "hash + dependabot" together seems contradictory.
To me, to be coherent, it should be either "pin by hash and never update any action", or "use dependabot and don't care about hashes or versions".

sergejparity:
Reason is pretty simple to get notified about that something has changed in third party GHA. Although it was mentioned in the forum discussion, that we do not have a capacity to perform full audit of changes, but at least there is a chance to avoid [similar to this issues](https://gitlab.parity.io/parity/vulnerability-response/-/issues/50)

Issue #2685, opened 2022-08-25
RPC AccountId deserialization rejects prefixes larger than 1 byte


arrudagates:
This line is the main cause of the problem: https://github.com/paritytech/smoldot/blob/main/src/json_rpc/methods.rs#L576
But the lines below also need to be changed to accommodate for more prefix lengths.

tomaka:
Closing after #2686 

While the issue isn't "properly" fixed, I think it's "fixed enough" for the foreseeable future.

PR #2686, opened 2022-08-25
Make AccountId deserialization accept prefixes larger than 1 byte

arrudagates:
This fixes the issue described in #2685 however it doesn't properly solve it in a clean manner. A perfect solution would be to somehow retrieve the prefix length from the current chain and use that as the offset for the account id byte array.

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Fixed
 
-- Fix inability to decode addresses with prefixes longer than 1 byte in the RPC. ([#2686](https://github.com/paritytech/smoldot/pull/2686))
+- Fix inability to decode addresses with prefixes longer than 1 byte when calling `system_accountNextIndex`. ([#2686](https://github.com/paritytech/smoldot/pull/2686))
 
 ## 0.6.30 - 2022-08-12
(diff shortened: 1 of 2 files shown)


tomaka:
Thanks! Could you add an entry in `bin/wasm-node/CHANGELOG.md` as well?

arrudagates:
> Thanks! Could you add an entry in `bin/wasm-node/CHANGELOG.md` as well?

Done!

bin/wasm-node/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - In case of protocol error, or if a peer refuses a block announces substream, no new substream with the same peer will be attempted for 20 seconds. This avoids loops where the same peer is tried over and over again.
 
+### Fixed
+
+- Fix inability to decode addresses with prefixes longer than 1 byte in the RPC. ([#2686](https://github.com/paritytech/smoldot/pull/2686))
```
tomaka, line 11:
```suggestion
- Fix inability to decode addresses with prefixes longer than 1 byte when calling `system_accountNextIndex`. ([#2686](https://github.com/paritytech/smoldot/pull/2686))
```

tomaka (approved):

PR #2694, opened 2022-08-30
Provide a Host in the WebSocket handshake

tomaka:
Right now, on the client side, we pass `Host: ...` in the HTTP headers.

While libp2p itself doesn't care about the value in the `Host` header, many nodes that use WebSocket have an HTTP reverse proxy in front of them. This reverse proxy does care about this value, and it is thus important to pass it properly.


melekes (approved):
👍

PR #2696, opened 2022-08-30
Properly poll the socket in a loop

tomaka:
When reading or writing returns `Poll::Ready`, it is important that we try reading/writing again until `Pending` is returned, so that the `Context` is properly registered.

### DIFF
--- a/bin/light-base/src/platform/async_std.rs
+++ b/bin/light-base/src/platform/async_std.rs
@@ -189,21 +189,24 @@ impl Platform for AsyncStdTcpWebSocket {
             async_std::task::spawn(future::poll_fn(move |cx| {
                 let mut lock = shared.guarded.lock();
 
-                match Pin::new(&mut read_data_tx).poll_ready(cx) {
-                    Poll::Ready(Ok(())) => {
-                        if let Poll::Ready(result) =
-                            Pin::new(&mut socket).poll_read(cx, &mut read_buffer)
-                        {
-                            match result {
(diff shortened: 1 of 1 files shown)


bin/light-base/src/platform/async_std.rs
```diff
@@ -189,50 +189,64 @@ impl Platform for AsyncStdTcpWebSocket {
...
+                loop {
+                    match Pin::new(&mut read_data_tx).poll_ready(cx) {
+                        Poll::Ready(Ok(())) => {
+                            match Pin::new(&mut socket).poll_read(cx, &mut read_buffer) {
+                                Poll::Pending => break,
+                                Poll::Ready(result) => {
+                                    match result {
+                                        Ok(0) | Err(_) => return Poll::Ready(()), // End the task
+                                        Ok(bytes) => {
+                                            let _ = read_data_tx
```
melekes, line 201: shouldn't we break here too? otherwise how we'd go to line \#213?

tomaka: No we shouldn't break here, that's the entire point of the PR.

After we have read from the socket, we jump back to line 193. Then from 193 we jump either to 209 and break, or to 195 and read again. And reading again might lead to line 196 which breaks.

It is important that, after reading we try to read again, as otherwise the task will not wake up if there is data available on the socket.
This is important in this context because this task is a background task that just reads and sends the read data to a channel.

melekes (approved):

PR #2701, opened 2022-08-31
Remove hack that checks for duplicate finality

tomaka:
https://github.com/paritytech/smoldot/pull/2658 removed the code that sometimes erroneously sent a `Notification::Finalized` about a block that was already finalized. As such, we can remove the hack in `runtime_service` that checks for this.


melekes (approved):
👍

PR #2708, opened 2022-09-01
General clean-up of the single stream connections code

tomaka:
This PR is a general clean-up of the single stream connection code. It fixes many small issues, and also handles the shutdown phase more properly.
We now properly handle incoming GoAway frames, and have the possibility to send some.

I understand that this code is generally extremely complicated, but I don't really have a good solution for this. Networking code *is* extremely complicated, both in terms of logic and implementation.
Maybe a solution is tests, such as https://github.com/paritytech/smoldot/pull/2323, but at the moment the code isn't even robust enough to have tests.

### DIFF
--- a/src/libp2p/connection/established/single_stream.rs
+++ b/src/libp2p/connection/established/single_stream.rs
@@ -161,11 +161,17 @@ where
 
         // First, update all the internal substreams.
         // This doesn't read data from `read_write`, but can potential write out data.
+        for substream_id in self
+            .inner
+            .yamux
+            .user_datas()
+            .map(|(id, _)| id)
+            .collect::<Vec<_>>()
         {
-            let read_bytes_before = read_write.read_bytes;
-            let out = self.update_all(read_write);
-            debug_assert_eq!(read_bytes_before, read_write.read_bytes);
-            if let Some(event) = out {
+            let (_num_read, event) =
+                Self::process_substream(&mut self.inner, substream_id, read_write, &[]);
+            debug_assert_eq!(_num_read, 0);
+            if let Some(event) = event {
                 return Ok((self, Some(event)));
             }
         }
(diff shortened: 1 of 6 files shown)


src/libp2p/connection/established/single_stream.rs
```diff
@@ -258,12 +266,17 @@ where
                 .map_err(Error::Yamux)?;
             self.inner.yamux = yamux_decode.yamux;
 
-            // TODO: it is possible that the yamux reading is blocked on writing
+            // If bytes_read is 0 and detail is None, then Yamux can't do anything more. On the
```
melekes, line 292: I don't see how this comment adds anything. it just duplicates the code basically

src/libp2p/connection/yamux.rs
```diff
@@ -1564,6 +1604,33 @@ impl<'a, T> ExtractOut<'a, T> {
                 }
 
                 Outgoing::Idle => {
+                    // Send a "GoAway" frame if demanded.
+                    if let OutgoingGoAway::Required(code) = self.yamux.outgoing_goaway {
+                        let mut header = arrayvec::ArrayVec::new();
+                        header.push(0);
```
melekes, line 1739: what `0` is representing here?

tomaka: It's a Yamux version number. See https://github.com/hashicorp/yamux/blob/master/spec.md#version-field

There are several snippets of code that generate a header, and it's true that it wouldn't be a bad idea to move them to the `header` module for example, but I also don't think it's a big deal.

src/libp2p/connection/established/single_stream.rs
tomaka: It's true that it's repeating a piece of information of the Yamux API.
But IMO, the fact that we have to loop again if Yamux didn't process all the data or returned an event is a bit subtle (especially the "or" here, so for example if Yamux processed all the data and returned an event we have to loop again), and is worth repeating for clarity.

melekes (approved):
👍

PR #2710, opened 2022-09-02
Move erronous chain system to wasm-node code

tomaka:
Adding a chain can fail. If adding a chain fails, we need to report the reason for the failure to the user.
Unfortunately, it is rather difficult to pass a string from Rust to JavaScript. The best way to do so is for the JavaScript to call a Rust function that returns a pointer to said string.
And that's exactly what we do if chain initialization fails. If adding a chain fails, we still insert a chain but insert it in "erroneous" mode. When a chain is in erroneous mode, the error message is kept in memory and can be retrieved by the JavaScript. Once retrieved, the JavaScript calls another function that cleans up the chain.

Before this PR, this system was done in the `light-base` library.
Unfortunately, this leads to a rather weird API for the client. It is much more idiomatic if `add_chain()` simply returns a `Result`.
That's what this PR does.
The "erroneous chain" system described above is still present, but is now done entirely in the code of `wasm-node`, as this system is highly specific to the Wasm node.

The unfortunate consequence is that there is now an additional level of indirection. When the JavaScript for example wants to send a JSON-RPC request, we first look up the ID of the chain in a table to determine whether it's healthy and find the actual ID of the chain that the client manipulates. But I think the trade-off is worth it.

### DIFF
--- a/bin/light-base/examples/basic.rs
+++ b/bin/light-base/examples/basic.rs
@@ -47,46 +47,36 @@ fn main() {
     let (json_rpc_responses_tx, mut json_rpc_responses_rx) = mpsc::channel(32);
 
     // Ask the client to connect to a chain.
-    let chain_id = client.add_chain(smoldot_light::AddChainConfig {
-        // The most important field of the configuration is the chain specification. This is a
-        // JSON document containing all the information necessary for the client to connect to said
-        // chain.
-        specification: include_str!("../../polkadot.json"),
+    let chain_id = client
+        .add_chain(smoldot_light::AddChainConfig {
+            // The most important field of the configuration is the chain specification. This is a
+            // JSON document containing all the information necessary for the client to connect to said
+            // chain.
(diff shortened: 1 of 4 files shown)


melekes:
> Adding a chain can fail. If adding a chain fails, we need to report the reason for the failure to the user.
Unfortunately, it is rather difficult to pass a string from Rust to JavaScript. The best way to do so is for the JavaScript to call a Rust function that returns a pointer to said string.
And that's exactly what we do if chain initialization fails. If adding a chain fails, we still insert a chain but insert it in "erroneous" mode. When a chain is in erroneous mode, the error message is kept in memory and can be retrieved by the JavaScript. Once retrieved, the JavaScript calls another function that cleans up the chain.

Thanks for the description. It really shines the light on some design choices. You can probably write a book by now wrt JS and Rust system design 😀

melekes (approved):
👍

PR #2711, opened 2022-09-02
Handle websocket client API usage error

tomaka:
Handles the situation where the API user tries to send data on a socket after having closed its writing side.
Unfortunately, the `Read`/`Write`/`AsyncRead`/`AsyncWrite` traits are completely unspecified regarding a lot of aspects, and one of these aspects is the situation in question here.
Considering that it is an incorrect API usage, this path is never supposed to be reached anyway.


melekes (approved):
👍

PR #2714, opened 2022-09-04
Add some tests for remove_prefix

tomaka:
Adds some tests regarding the code added in https://github.com/paritytech/smoldot/pull/2663

I wanted to maybe figure out why https://github.com/paritytech/smoldot/pull/2666 was failing, but didn't manage to find a failing case.


melekes (approved):
👍

PR #2719, opened 2022-09-07
Refactor parachains service

tomaka:
This PR refactors the parachains service to make it easier to read and understand.

No change to the logic has been performed.
Can be reviewed commit by commit. I've intentionally done `rustfmt`almost at the end in order to not pollute the diff.

### DIFF
--- a/bin/light-base/src/sync_service/parachain.rs
+++ b/bin/light-base/src/sync_service/parachain.rs
@@ -198,14 +198,13 @@ impl<TPlat: Platform> ParachainBackgroundTask<TPlat> {
         match self.subscription_state {
             ParachainBackgroundState::NotSubscribed { .. } => {
 
-        self.subscription_state = ParachainBackgroundState::Subscribed({
+        
             log::debug!(target: &self.log_target, "Subscriptions <= Reset");
 
-            let relay_chain_subscribe_all = loop {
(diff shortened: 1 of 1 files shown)


bin/light-base/src/sync_service/parachain.rs
```diff
@@ -45,752 +45,1032 @@ pub(super) async fn start_parachain<TPlat: Platform>(
...
+                    }
+                }
+
+                ParachainBackgroundState::Subscribed(ref mut runtime_subscription) => {
+                    futures::select! {
+                        () = &mut runtime_subscription.next_start_parahead_fetch => {
+                            // Do nothing. This is simply to wake up and loop again.
+                        },
+
+                        relay_chain_notif = runtime_subscription.relay_chain_subscribe_all.next().fuse() => { // TODO: remove fuse()?
```
melekes, line 253: extract into `process_relay_chain_notification`?

melekes (approved):
👍

bin/light-base/src/sync_service/parachain.rs
tomaka: I don't think it's a good idea.
The block that I've left it handles situations where the channel of relay chain notifications has been closed. In other words, it's more "meta" than processing the notification itself.

PR #2723, opened 2022-09-07
Use a struct for runtime service pinned blocks

tomaka:
Minor refactoring to help with clarity of the code.

### DIFF
--- a/bin/light-base/src/runtime_service.rs
+++ b/bin/light-base/src/runtime_service.rs
@@ -400,8 +400,7 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
             ..
         } = &mut guarded_lock.tree
         {
-            let block_counts_towards_limit = match pinned_blocks
-                .remove(&(subscription_id.0, *block_hash))
+            let block_ignores_limit = match pinned_blocks.remove(&(subscription_id.0, *block_hash))
             {
                 Some(b) => b.block_ignores_limit,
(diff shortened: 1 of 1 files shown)


bin/light-base/src/runtime_service.rs
```diff
@@ -417,7 +417,7 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
 
             guarded_lock.runtimes.retain(|_, rt| rt.strong_count() > 0);
 
-            if block_counts_towards_limit {
+            if !block_counts_towards_limit {
```
melekes, line 420: don't want to bike shed, but I wouldn't call this PR "minor refactoring" as it changes the logic here. I hope it's on purpose

melekes (approved):
👍

bin/light-base/src/runtime_service.rs
```diff
@@ -403,7 +403,7 @@ impl<TPlat: Platform> RuntimeService<TPlat> {
             let block_counts_towards_limit = match pinned_blocks
```
melekes, line 403:
```suggestion
            let block_ignores_limit = match pinned_blocks
```

bin/light-base/src/runtime_service.rs
melekes, line 420:
```suggestion
            if !block_ignores_limit {
```

PR #2724, opened 2022-09-08
Add a limit to the number of inbound substreams

tomaka:
In order to avoid DoS attacks, it is important to prevent remotes from opening an unlimited number of substreams.


melekes (approved):
👍

Issue #2726, opened 2022-09-08
Debug assert failure in optimistic syncing


tomaka:
https://github.com/paritytech/smoldot/blob/072b7a936734d76cdf878b7dbf9ad0d3407d08d9/src/sync/optimistic/verification_queue.rs#L377

tomaka:
Suspecting the `insert_request` function

tomaka:
This one fails: https://github.com/paritytech/substrate-lite/blob/f6d13d0cd5c693d1dac1d08cd35337b88fd33b22/src/sync/optimistic/verification_queue.rs#L406-L407

PR #2727, opened 2022-09-08
Add debug assertions to the verification queue

tomaka:
cc https://github.com/paritytech/smoldot/issues/2726

Didn't manage to reproduce the issue so far, but at least these debug assertions should help diagnosing the next time it happens.


melekes (approved):
👍

PR #2734, opened 2022-09-12
Add debug_assert! in parachain code

tomaka:
Checks that the logic of the `async_tree` code is correct and that it finalized blocks one after the other.


melekes (approved):
👍

PR #2735, opened 2022-09-12
Move network services tasks to sub-modules

tomaka:
Extracts the background tasks dedicated to connections in sub-modules.
Improves clarity. No logic has been touched.


melekes (approved):
👍

Issue #2468, opened 2022-07-05
Panic in runtime call after upgrade


tomaka:
Trying to send a transaction after a runtime upgrade happened has led to:

> Error: panicked at 'called `Result::unwrap()` on an `Err` value: Call(StorageQuery(StorageQueryError { errors: [ProofVerification(TrieRootNotFound), ProofVerification(TrieRootNotFound), ProofVerification(TrieRootNotFound)] }))', /home/tomaka/Projets/smoldot/bin/light-base/src/json_rpc_service/state_chain.rs:1188:26
> (demo) Error during JSON-RPC request: Error: Error: panicked at 'called `Result::unwrap()` on an `Err` value: Call(StorageQuery(StorageQueryError { errors: [ProofVerification(TrieRootNotFound), ProofVerification(TrieRootNotFound), ProofVerification(TrieRootNotFound)] }))', /home/tomaka/Projets/smoldot/bin/light-base/src/json_rpc_service/state_chain.rs:1188:26

tomaka:
https://github.com/paritytech/smoldot/blob/84745accd5608c98ceb4e3d6733e573808e442ad/bin/light-base/src/json_rpc_service/state_chain.rs#L1188

wirednkod:
Omni reported the following:
![image](https://user-images.githubusercontent.com/5408605/186389802-efc7c21a-3759-40c6-b06e-5545cad62606.png)

I think this is relevant to this issue

pgolovkin:
@tomaka when the release may be expected?

tomaka:
https://github.com/paritytech/smoldot/pull/2745

PR #2736, opened 2022-09-12
Rewrite `state_get_runtime_version` to use shared code

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2468

This PR does two things:

- Split `runtime_call` in half, where the first half is now a separate function named `runtime_lock`.
- Make `state_get_runtime_version` use this new `runtime_lock` function.

This is better so that we don't repeat a lot of code, and also fixes a few potential panics.


melekes (approved):
👍

PR #2748, opened 2022-09-14
Fix #2726

### LINKED ISSUE #2726: Debug assert failure in optimistic syncing

tomaka:
Fix #2726 

The issue was simply that we didn't check `index - 1`


melekes (approved):
👍

PR #2751, opened 2022-09-14
Fix infinite loop in single-stream connection

### LINKED ISSUE #2750: Infinite loop in established::SingleStream::read_write
My node is stuck in an infinite loop.
The backtraces look like this:

image

image

tomaka:
Fix #2750


melekes (approved):
👍

Issue #2754, opened 2022-09-15
Change the multi-stream connections API


tomaka:
After trying to implement the WebRTC handshake, I've changed my mind on how to design the APIs.

Rather than providing a PeerId (the actual one that's been negotiated), the multi-stream connection opener now just says "this is a libp2p webrtc connection", and the negotiation is done entirely within smoldot.

In the same vain, single-stream connections should also get an extra parameter indicating something like "this needs multistream-select for yamux and noise". In the future, we can also imagine additionally indicating "this is websocket" or "this is websocket secure" and let smoldot do the WebSocket and TLS handshakes.

cc @melekes 
cc https://github.com/paritytech/smoldot/issues/1712

melekes:
Should we then rename 'multi-stream' to 'webrtc-multi-stream'?

tomaka:
I had instead in mind to add a parameter when creating a multistream connection indicating the type (in this case WebRTC)

PR #2755, opened 2022-09-15
Perform the multi-stream connections handshake inside smoldot

tomaka:
cc https://github.com/paritytech/smoldot/issues/2754

In terms of API, this removes the necessity to pass a `PeerId` when creating a "multi-stream connection". The inner implementation now performs the substream opening and Noise negotiation internally.

### DIFF
--- a/src/libp2p/collection.rs
+++ b/src/libp2p/collection.rs
@@ -377,8 +377,9 @@ where
 
     /// Adds a new multi-stream connection to the collection.
     ///
-    /// Note that no [`Event::HandshakeFinished`] event will be generated. The connection is
-    /// immediately considered as fully established.
+    /// Must be passed the moment (as a `TNow`) when the connection as been established, in order
+    /// to determine when the handshake timeout expires.
+    // TODO: add an is_initiator parameter? right now we're always implicitly the initiator
     pub fn insert_multi_stream<TSubId>(
         &mut self,
         now: TNow,
@@ -394,6 +395,7 @@ where
             self.randomness_seeds.gen(),
             now,
             self.max_inbound_substreams,
+            self.noise_key.clone(),
             self.notification_protocols.clone(),
             self.request_response_protocols.clone(),
             self.ping_protocol.clone(),
--- a/src/libp2p/collection/multi_stream.rs
+++ b/src/libp2p/collection/multi_stream.rs
@@ -145,14 +145,6 @@ where
                 })),
             },
         }
-
-        /*
-        TODO: old code:
-
-
-
-
-        */
     }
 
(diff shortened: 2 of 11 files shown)


tomaka:
Note that the code has a couple of `todo!`s and unimplemented things, because I don't actually expect it to work first try. This PR is merely an attempt to move #2579 forward. The code that this PR touches isn't used anywhere at the moment (apart from #2579).

src/libp2p/collection.rs
```diff
@@ -377,8 +377,9 @@ where
 
     /// Adds a new multi-stream connection to the collection.
     ///
-    /// Note that no [`Event::HandshakeFinished`] event will be generated. The connection is
-    /// immediately considered as fully established.
+    /// Must be passed the moment (as a `TNow`) when the connection as been established, in order
+    /// to determine when the handshake timeout expires.
+    // TODO: add an is_initiator parameter? right now we're always implicitly the initiator
```
melekes, line 382: if it's an incoming substream, we're not initiators

melekes (commented):
are you sure this is easier than the old approach + Rust binding called from JS, which performs Noise handshake?

src/libp2p/collection/multi_stream.rs
```diff
@@ -35,11 +38,32 @@ pub struct MultiStreamConnectionTask<TNow, TSubId> {
     connection: MultiStreamConnectionTaskInner<TNow, TSubId>,
 }
 enum MultiStreamConnectionTaskInner<TNow, TSubId> {
+    /// Connection is still in its handshake phase.
+    Handshake {
+        /// Substream that has been opened to perform the handshake, if any.
+        opened_substream: Option<TSubId>,
+
+        /// Noise handshake in progress. Always `Some`, except to be temporarily extracted.
+        handshake: Option<noise::HandshakeInProgress>,
```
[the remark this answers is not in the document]
tomaka: Part 2 of https://github.com/paritytech/smoldot/issues/2754 is to add a parameter when creating a single-stream or multi-stream connection indicating the nature of the handshake.

src/libp2p/collection.rs
tomaka: This function is for adding multi-stream connections as a whole.

If one day we add support for browser-to-browser WebRTC connections, then `is_initiator` could be `false`.

tomaka:
> are you sure this is easier than the old approach + Rust binding called from JS, which performs Noise handshake?

This is definitely easier. The old approach requires going through the JS<->Rust FFI layer, which is kind of complicated. It also requires proper handling of what to do if an exception is thrown by smoldot, which is also super complicated.

melekes (approved):

PR #2758, opened 2022-09-16
Provide the type of handshake when inserting a connection

tomaka:
Close https://github.com/paritytech/smoldot/issues/2754

At the moment, we simply say "this is a single stream connection" and smoldot negotiates Noise+Yamux on top of it, or we say "this is a multistream connection" (without saying what kind) and smoldot automatically negotiates WebRTC on top of it.

Instead of doing it this way, after this PR we now pass the type of handshake as a parameter.

In addition, when we say "this is a WebRTC connection", we must now provide the local and remote TLS certificate fingerprints. This is necessary in order to properly perform the Noise handshake.

The main difficulty of this change was going through the Rust <-> JS FFI layer with the two multihashes, as we can't know the length of a multihash in advance.


melekes (approved):
👍

Issue #2763, opened 2022-09-20
Panicked at 'called `Result::unwrap()`


tuul-wq:
Got the following error while connected to Polkadot using lates version of `@polkadot/rpc-provider": "9.4.1` / `"@substrate/smoldot-light": "0.6.33"`

> Smoldot has panicked while executing task "json-rpc-polkadot"
> `Panicked at 'called Result::unwrap() on an Err value: ()', /__w/smoldot/smoldot/bin/light-base/src/json_rpc_service.rs:1346:61`

tomaka:
https://github.com/paritytech/smoldot/blob/405c38c395327bc3b834f5c3bedec026ded9fa34/bin/light-base/src/json_rpc_service.rs#L1354

PR #2764, opened 2022-09-20
Remove unwrap() in runtime_lock

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2763

This PR gives a proper error type to the `state_trie_root_hash` function (which retrieves the state trie root hash of a block from the network), and makes `runtime_lock` properly propagates the error instead of unwrapping.


melekes (approved):
👍

PR #2774, opened 2022-09-23
Don't ignore extra substreams opened during handshake

tomaka:
Right now, if an extra substream is opened during the handshake, we simply ignore it. The substream will plain not work.

In principle, no extra substream should be opened during the handshake, provided we are the client. However, due to bugs and implementation difficulties, it is actually not impossible that this happens in reality, in which case simply ignoring the substream is the worst thing we can do.


melekes (approved):
👍 

We don't need to do anything wrt extra_open_substreams if handshake fails, right?

tomaka:
> We don't need to do anything wrt extra_open_substreams if handshake fails, right?

If the handshake fails, the connection is thrown away entirely, which will end up calling `close` in `index-browser.ts`, which closes the entire `RTCPeerConnection`

Issue #2528, opened 2022-07-19
Replace JSON-RPC callback with a Promise-based API


tomaka:
Instead of having a callback, the user should manually call a method to retrieve the next JSON-RPC response.
This would make it possible to have back-pressure on the JSON-RPC server.

PR #2778, opened 2022-09-25
Remove AddChainOptions.jsonRpcCallback in favor of Chain.nextJsonRpcResponse

tomaka:
Close https://github.com/paritytech/smoldot/issues/2528

This PR is a major change in the public API of the JS library, as such I will bump to 0.7.0 at the next release.

Instead of passing a callback when adding a chain where the JSON-RPC responses are sent, you are now supposed to call `nextJsonRpcResponse` on the chain.
This asynchronous function waits (if necessary) for a JSON-RPC response to be available then returns it.

In addition to this, `sendJsonRpc` now returns an error if the request is malformed or if the queue is full.

This change in API has several major advantages:

- It removes an ambiguity about what to do if the `jsonRpcCallback` throws an exception.
- It is now possible to back-pressure the client by not calling `nextJsonRpcResponse`. This will cause the responses to pile up in the buffer, until a point where `sendJsonRpc` will return an error.
- It is overall easier to use the API because you no longer have to create things ahead of time (before the chain is even created) in order to pass them to the callback. See the tests that this PR rewrites for example.

## Migration

It is always possible to migrate to the new API.

If your code looks like this:

```js
const chain = await client.addChain({ jsonRpcCallback, /* other options */ ... });
```

You can rewrite it like this:

```js
const chain = await client.addChain({ /* other options */ ... });

(async () => {
    try {
        while(true) {
            const response = await chain.nextJsonRpcResponse();
            /* do whatever you want here */
        }
    } catch(_error) {
        // Will happen when the chain is removed or the client terminated.
    }
})()
```

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Changed
 
+- `Chain.sendJsonRpc` now throws a `MalformedJsonRpcError` exception if the JSON-RPC request is too large or malformed, or a `QueueFullError` if the queue of JSON-RPC requests of the chain is full.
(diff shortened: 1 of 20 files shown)


melekes (approved):
👍

PR #2779, opened 2022-09-25
Fix Deno demo

tomaka:
I had to look at the source code of Deno because the doc is [non-existent](https://deno.land/api@v1.25.4?s=Deno.HttpConn#method_nextRequest_0).

When `null` is returned, it means the connection can no longer serve requests, and so if we `continue` we just get an infinite loop.


melekes (approved):
👍

PR #2780, opened 2022-09-26
Make Yamux return an event on stream reset

tomaka:
Fixes a TODO in the Yamux code.

This change has absolutely no influence, as `StreamReset` events are simply ignored by the higher level code, but in principle higher level code could use it.


melekes (approved):
👍

Issue #2467, opened 2022-07-05
Memory isn't freed after a runtime upgrade


tomaka:
When a runtime upgrade happens, and smoldot is connected, it will download the new runtime. This obviously temporarily increases the memory usage of smoldot.
However, afterwards, the previous runtime should be freed. For an unknown reason this doesn't seem to happen, and the memory usage remains double the normal amount.

tomaka:
I've checked (outside of any runtime upgrade situation) whether there was a leak in the block pins, and the answer is no.
The JSON-RPC service cache will keep the previous runtime alive for 32 blocks, however I believe that I was way more than 32 blocks past the upgrade.

PR #2781, opened 2022-09-26
Clean-up runtimes management in runtime_service

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2467

We now clean up old runtimes when it makes sense, rather than after a download where it doesn't make sense.

### DIFF
--- a/bin/light-base/src/runtime_service.rs
+++ b/bin/light-base/src/runtime_service.rs
@@ -1528,19 +1528,20 @@ impl<TPlat: Platform> Background<TPlat> {
     ) {
         let mut guarded = self.guarded.lock().await;
 
-        guarded.runtimes.retain(|_, rt| rt.strong_count() > 0);
-
-        // Try to find an existing identical runtime.
+        // Try to find an existing runtime identical to the one that has just been downloaded.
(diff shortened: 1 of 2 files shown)


bin/light-base/src/runtime_service.rs
```diff
@@ -1965,7 +1975,7 @@ impl<TPlat: Platform> Background<TPlat> {
         // Clean up unused runtimes to free up resources.
         guarded
             .runtimes
-            .retain(|_, runtime| runtime.strong_count() == 0);
```
tomaka, line 1968: This was the main bug behind https://github.com/paritytech/smoldot/issues/2467

bin/light-base/src/runtime_service.rs
```diff
@@ -1965,7 +1975,7 @@ impl<TPlat: Platform> Background<TPlat> {
...
+            .retain(|_, runtime| runtime.strong_count() > 0);
```
tomaka, line 1978: This was the main bug behind https://github.com/paritytech/smoldot/issues/2467

melekes (approved):
👍

Issue #2740, opened 2022-09-12
Panic at single_stream


wirednkod:
Got the following error with Smoldot 0.6.32 (latest extension installed 0.6.19)

> Internal error in smoldot: Error: panicked at 'called `Option::unwrap()` on a `None` value', /__w/smoldot/smoldot/src/libp2p/connection/established/single_stream.rs:845:14

tomaka:
https://github.com/paritytech/smoldot/blob/d2c79186249a6f6a5492b01550763985421e1470/src/libp2p/connection/established/single_stream.rs#L856

tomaka:
https://github.com/paritytech/smoldot/blob/d2c79186249a6f6a5492b01550763985421e1470/src/libp2p/collection/single_stream.rs#L328

tuul-wq:
@wirednkod just got an error in the same file, it might be connected to your catch
`@substrate/smoldot-light^0.6.34`

> panicked at 'called `Option::unwrap()` on a `None` value', /__w/smoldot/smoldot/src/libp2p/connection/established/single_stream.rs:859:14

PR #2785, opened 2022-09-27
Fix race conditions in the handling of notifications in events

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2740

The logic of the handling of `NotificationsInOpenCancel` events wasn't correct, as it wasn't taking into account the fact that everything is asynchronous.


melekes (approved):
👍

PR #2795, opened 2022-09-28
Several clean ups and refactors of peers.rs

tomaka:
This PR does a major clean-up pass on `peers.rs` and cleans up the remnants of the old times when it was asynchronous.

In details:

- Completely remove the `DesiredOutNotificationId` system where you first allocated a "desired outbound request" and then send the handshake with a different function.
- Add a `unfulfilled_desired_peers` field so that we simply return the content of the field in the `unfulfilled_desired_peers()` function instead of iterating through everything.
- When a substream is marked as "not desired" and it is currently opening, we no longer immediately cancel the opening and instead we generate an event about the opening having failed. This makes it possible for API users to track which substreams are currently being open.
- A couple of minor changes: remove a redundant function, add documentation, etc.

I have other changes in mind, which I'll do later as they also touch files other than `peers.rs`.

### DIFF
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -774,6 +774,30 @@ where
                     let peer = if let Some(peer_index) = self.inner[id].peer_index {
                         let peer_id = self.peers[peer_index].peer_id.clone();
 
+                        // We might have to insert the peer back in `unfulfilled_desired_peers` if
+                        // it is desired.
+                        if (self.peers[peer_index].desired
+                            || self
+                                .peers_notifications_out
+                                .range(
+                                    (peer_index, usize::min_value())
+                                        ..=(peer_index, usize::max_value()),
+                                )
+                                .any(|(_, state)| state.desired))
+                            && !self
+                                .connections_by_peer
+                                .range(
+                                    (peer_index, ConnectionId::min_value())
+                                        ..=(peer_index, ConnectionId::max_value()),
+                                )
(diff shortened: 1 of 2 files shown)


src/libp2p/peers.rs
```diff
@@ -370,21 +388,36 @@ where
...
-
                     let connection_state = self.inner.connection_state(id);
                     debug_assert!(connection_state.shutting_down);
 
                     let peer = if let Some(peer_index) = self.inner[id].peer_index {
                         let peer_id = self.peers[peer_index].peer_id.clone();
 
+                        // We might have to insert the peer back in `unfulfilled_desired_peers` if
+                        // it is desired.
+                        if (self.peers[peer_index].desired
```
melekes, line 399: do you want to extract this into a function to avoid duplication?

tomaka:
I've reverted the last commit about not immediately cancelling substream openings, as I'm going to tackle this problem separately as part of https://github.com/paritytech/smoldot/issues/2799.

The PR now normally no longer changes the behavior of `peers.rs`, and only modifies the implementation details.

src/libp2p/peers.rs
tomaka: So there are 4 blocks that are similar, but only two of the four are actually exactly the same, and for these 2 it's not just this block that is repeated but way more code. I want to tackle this separately afterwards.

PR #2759, opened 2022-09-16
Set the Noise prologue in the WebRTC handshake

tomaka:
Let's wait for libp2p-webrtc to support this as well before merging this PR.

cc https://github.com/libp2p/rust-libp2p/pull/2903

PR #2807, opened 2022-09-30
set the Noise prologue in the WebRTC handshake

melekes:
Updates https://github.com/paritytech/smoldot/pull/2759
See https://github.com/melekes/rust-libp2p/pull/10 and https://github.com/libp2p/specs/pull/412#discussion_r978398623


src/libp2p/collection/multi_stream.rs
```diff
@@ -140,16 +140,35 @@ where
         request_response_protocols: Arc<[ConfigRequestResponse]>,
         ping_protocol: Arc<str>,
     ) -> Self {
-        // We only support one kind of handshake at the moment. Make sure (at compile time) that
-        // the value provided as parameter is indeed the one expected.
-        let MultiStreamHandshakeKind::WebRtc { .. } = handshake_kind;
+        // In the WebRTC handshake, the Noise prologue must be set to `"libp2p-webrtc-noise:"`
+        // followed with the multihash-encoded fingerprints of the local and remote certificates
+        // in ascending order.
```
tomaka, line 145:
```suggestion
        // followed with the multihash-encoded fingerprints of the initiator's certificate
        // and the receiver's certificate.
        // TODO: we currently assume that the local node is always the initiator
```

tomaka (approved):

PR #2808, opened 2022-09-30
Add a lot of documentation to the warp syncing algorithm code

tomaka:


src/sync/warp_sync.rs
```diff
@@ -675,6 +772,9 @@ impl<TSrc, TRq> InProgressWarpSync<TSrc, TRq> {
                     ..
                 },
             ) if source_id == *warp_sync_source_id => {
+                // If the source has failed a request, we jump back to downloading fragments
+                // in order to try a different source.
+                self.sources[source_id.0].already_tried = true;
```
tomaka, line 777: I've made one change to the behavior: if a request fails, we mark the source as "already tried", meaning that a different source will be tried next time. Otherwise, the same source might be tried again.

melekes (approved):
👍

PR #2818, opened 2022-10-04
More backports to the JSON-RPC API

tomaka:
In complement with https://github.com/paritytech/smoldot/pull/2812, this PR updates `chainHead_unstable_storage` with the latest changes in the JSON-RPC API: https://github.com/paritytech/json-rpc-interface-spec/pull/16 and https://github.com/paritytech/json-rpc-interface-spec/pull/17

We should now be up to date.


melekes (approved):
👍

PR #2819, opened 2022-10-04
Implement encoding proof node values

tomaka:
Necessary in order to finish https://github.com/paritytech/smoldot/pull/1166

Right now we can decode items from Merkle proofs. This PR implements encoding back these items.

While implementing encoding, I've discovered several cases where the decoding was too lenient and was accepting as valid things that shouldn't really be valid. I've fixed this as part of this PR. See the changes in `src/util.rs` and at the bottom of `src/trie/proof_node_codec.rs`.

When I say "shouldn't really be valid" I mean that there were multiple different sequences of bytes that could lead to the same decoded struct. Since we're hashing things a lot here and there, it is important that there is a bijection of decoded <-> encoded. The Substrate implementation does the same. cc https://github.com/paritytech/substrate/pull/12417


melekes (approved):
👍

PR #2828, opened 2022-10-05
Fix service.rs never cleaning up in_peers

tomaka:
This is a pretty long-standing bug.
In the case where all out slots are occupied, and a remote opens an inbound service then closes it, `service.rs` doesn't clean up `in_peers`. This leads to a debug assert failure when we later try to reinsert into `in_peers` only to realize that there's already a value there.


melekes (approved):
👍

Issue #2831, opened 2022-10-06
Problem in identify protocol


tomaka:
When connecting a smoldot node to a Substrate node, this can be found in the Substrate node's logs:

```
2022-10-06 11:03:24.529 DEBUG tokio-runtime-worker libp2p_identify::protocol: Invalid message: Custom { kind: InvalidData, error: DecodingError { msg: "RSA X.509", source: Some(Asn1DerError { error: InOutError("Failed to decode object @/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/asn1_der-0.7.5/src/typed/mod.rs:42"), source: Some(ErrorSource { inner: Asn1DerError { error: InOutError("Failed to read tag @/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/asn1_der-0.7.5/src/der.rs:116"), source: Some(ErrorSource { inner: Asn1DerError { error: InOutError("Failed to read element from underlying source @/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/asn1_der-0.7.5/src/data.rs:70"), source: Some(ErrorSource { inner: Asn1DerError { error: InOutError("Cannot read beyond end of chain-iterator @/usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/asn1_der-0.7.5/src/data.rs:52"), source: None } }) } }) } }) }) } }    
2022-10-06 11:03:24.529 DEBUG tokio-runtime-worker libp2p_core::upgrade::apply: Failed to apply negotiated protocol    
2022-10-06 11:03:24.529 DEBUG tokio-runtime-worker sub-libp2p: Identification with peer PeerId("12D3KooWFFrgz9mGYnb7XWGJ842pr79W9f23RrfniE7DA5VEvdBf") failed => upgrade apply error: Key decoding error: RSA X.509    
```

tomaka:
I have no idea what the problem could be. The fact that rust-libp2p even tries to decode as RSA is a bit puzzling.

tomaka:
The problem is actually probably that the identify protocol shouldn't send back a length prefix for the response either.

tomaka:
[...]

PR #2838, opened 2022-10-06
Fix the identify protocol using a length prefix when it shouldn't

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2831

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Fix the responses to libp2p identify requests being wrongly encoded.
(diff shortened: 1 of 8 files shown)


tomaka:
Actually still failing: 

> 2022-10-06 20:04:02.862 DEBUG tokio-runtime-worker sub-libp2p: Identification with peer PeerId("12D3KooWNLaXTidiJ7sRbpcnnd2NpJzV5AbmxQwYpRREP42AT7sF") failed => upgrade apply error: Key decoding error: RSA X.509

tomaka:
Still need to do the no-prefix response parsing.

tomaka:
Ok, it turns out that this PR is conceptually wrong. Rust-libp2p (haven't checked the other implementations) actually sends the length of the message.

It's very subtle, but `prost_codec` actually adds length prefixes
https://github.com/libp2p/rust-libp2p/blob/87ab49e8122c9b8daa083bf02816b5ad92453f7a/protocols/identify/src/protocol.rs#L216

Will leave this PR open until https://github.com/paritytech/smoldot/issues/2831 is fixed, just in case.

PR #2842, opened 2022-10-07
Fix responding to block requests mixing ascending and descending

tomaka:
After this PR, Substrate can now sync from smoldot.


melekes (approved):
👍

Issue #1681, opened 2021-11-25
Use read_only_runtime_host in chain_information submodules


tomaka:

tomaka:
The `chain_information` submodules should be entirely reconsidered.
It is a bit cumbersome to create a separate API for a specific runtime function execution, while the caller could just use `read_only_runtime_host` directly.

Issue #1752, opened 2021-12-07
Full node should load consensus data from chain storage instead of storing it in database


tomaka:
Instead of storing the Aura/Babe/Grandpa authorities in the database, the full node should load them from the chain storage.

Before tackling this issue, one should check whether it is actually possible to load everything from the chain storage.

tomaka:
First we should refactor the code that builds the chain information.

cc https://github.com/paritytech/smoldot/issues/1681

tomaka:
I've realized a problem with https://github.com/paritytech/smoldot/pull/2845, which is that there were moments when we didn't provide any runtime function to get the GrandPa current set id.

tomaka:
This can't be done in we want to support syncing from scratch. We can, however, implement this if we force the full node to warp sync.

Issue #949, opened 2021-06-17
Smoldot should check the API version of the runtime specs before doing a call


tomaka:
Right now smoldot just assumes that runtime functions exist and match a certain API. Unfortunately, the runtime function APIs can change while keeping the same name. The authoritative thing here is the `apis` field of the runtime spec.

tomaka:
cc https://github.com/paritytech/smoldot/pull/1975

tomaka:
I believe a good way to do this is to build the `CoreVersion` in `HostVmPrototype::new` and then hold it in the `HostVmPrototype`.
Then the `runtime_host` and `read_only_runtime_host` should require an API name and version.

Also relates to https://github.com/paritytech/smoldot/issues/1107

tomaka:
According to https://github.com/paritytech/substrate/pull/8688, the `apis` field is actually more or less deprecated.
It seems that the design of Substrate does not consist in checking `apis`.

This issue could be closed as "wontfix", but I'm leaving it open because we need a substitute mechanism and I don't know which yet.

tomaka:
> According to https://github.com/paritytech/substrate/pull/8688, the apis field is actually more or less deprecated.
It seems that the design of Substrate does not consist in checking apis.

This turned out to be completely wrong.

PR #2845, opened 2022-10-08
Refactor chain information building process

tomaka:
Close https://github.com/paritytech/smoldot/issues/1681

Right now the code of `warp_sync` and `chain_spec` both build a `ChainInformation` struct by making runtime calls. However the code of these two modules is completely different.
Because these two modules were built very early on in smoldot's history, they are getting help from the sub-modules of the `chain_information` module, that are also very old.

This PR modernizes all of this: it removes all the sub-modules under `chain_information` and merges them into one that does everything.

This new module is now used by `warp_sync` and `chain_spec`, and `warp_sync` and `chain_spec` are now more simple.

I'm also planning to use this new module in order to refactor the SQLite database. (#1752)

This PR slightly changes the behavior of the warp syncing: we now download the runtime, then build it, and then only download the call proofs to obtain the chain information. Right now all the downloads are done in parallel.
This change is in principle more correct. At the moment we blindly call functions that we expect to exist. By downloading and building the runtime first, we can check the list of supported functions and then only call them. (cc #949) This is really not a problem at the moment because none of the functions that we use has ever been deprecated, but it could have become one in the future.


src/chain/chain_information/build.rs
```diff
@@ -0,0 +1,979 @@
...
+            } else if api.name_hash == grandpa_api_name.as_bytes() {
+                runtime_has_grandpa = true;
+            }
+        }
+
+        let inner = ChainInformationBuildInner {
+            finalized_block_header: config.finalized_block_header,
+            call_in_progress: None,
+            virtual_machine: Some(config.runtime),
+            runtime_has_aura,
```
melekes, line 208: q: can runtime have multiple options? e.g. aura & babe? if not, why it's not an enum?

melekes (approved):
👍

src/chain/chain_information/build.rs
tomaka: Yes it can, in which case a `MultipleConsensusAlgorithms` error is returned once we try to build the chain info.

We could also instantly return an error at initialization, but it makes the API more annoying to use, and this is such a niche error that I don't think it matters when we return the error.

PR #2868, opened 2022-10-12
Report duplicate notification substreams as InboundError events

tomaka:
Fixes a TODO.

If a peer opens the same notifications substream multiple times, we refuse all substreams but the first. This is correct.
However when we refuse the substream we do so completely silently. Nothing is logged, and thus it makes it hard to figure out whether this happens.

This PR reports this situation as an event so that the upper layers can output it in a log for example (that's how `InboundError` events are handled at the moment).

### DIFF
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -59,7 +59,7 @@ use rand::{Rng as _, SeedableRng as _};
 
 pub use collection::{
     ConfigRequestResponse, ConfigRequestResponseIn, ConnectionId, ConnectionToCoordinator,
-    CoordinatorToConnection, InboundError, MultiStreamConnectionTask, MultiStreamHandshakeKind,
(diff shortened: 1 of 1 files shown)


melekes:
> This PR reports this situation as an event so that the upper layers can output it in a log for example (that's how InboundError events are handled at the moment).

Are you sure this is even needed. For me this looks like a situation where silent dropping is fine imho.

melekes (approved):

tomaka:
> Are you sure this is even needed? For me this looks like a situation where silent dropping is fine imho.

The idea is to detect bugs.

If there's a bug somewhere that causes a node to violate the protocol, it means that there's either a bug locally or a bug on the other node. In that situation, it's a good idea to print something in the logs in order to alert the users that something is off so that they can report the issue.

PR #2869, opened 2022-10-12
Properly process dead substreams

tomaka:
At the moment, when a substream is closed gracefully, the code in `established::single_stream` just assumes that the substream state machine will return zero or one event then shut down.

While this is true in practice, this is a pretty big breach of abstraction layers. The state machine could in principle decide to return two events. Nothing in the API prevents it from doing so. And if it does, before this PR the second event would be ignored.

This PR splits the dead substreams handling of the yamux code in two: first we iterate over dead substreams, then only we remove them. Thanks to this, we can choose to not remove a dead substream for now, and instead continue processing it at the next iteration. In the example above, this would make it possible for the dead substream to send its second event.

### DIFF
--- a/src/libp2p/connection/established/single_stream.rs
+++ b/src/libp2p/connection/established/single_stream.rs
@@ -389,28 +389,66 @@ where
 
             // Substreams that have been closed or reset aren't immediately removed the yamux state
             // machine. They must be removed manually, which is what is done here.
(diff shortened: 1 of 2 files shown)


tomaka:
Running the `network-connection-encrypted` fuzzing target doesn't detect any panic.

melekes (approved):

PR #2870, opened 2022-10-12
Fix panic when sending block announces to disconnected peers

tomaka:
Doesn't warrant a CHANGELOG entry because this relates to the full node only.

### DIFF
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -1286,7 +1286,8 @@ where
     ///
     /// # Panics
     ///
-    /// Panics if [`SubstreamId`] is not a fully open outbound notifications substream.
+    /// Panics if there is no fully-open outbound substream with that peer-protocol combination.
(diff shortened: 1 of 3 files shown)


src/libp2p/peers.rs
```diff
@@ -1322,6 +1323,35 @@ where
...
+            None => return false,
+        };
+
+        match self
+            .peers_notifications_out
+            .get(&(peer_index, notifications_protocol_index))
+            .map(|state| &state.open)
+        {
+            Some(NotificationsOutOpenState::Open(_)) => true,
+            None
```
melekes, line 1346:
```suggestion
            _ => false
```

?

melekes (approved):

src/libp2p/peers.rs
tomaka: Yeah it's technically the same, but enumerating all the variants explicitly protects against mistakes if we add another variant to the enum.

Issue #165, opened 2020-11-08
Add a checkpoint warp sync system


tomaka:
We now have chain specs that contain the state of the chain at the finalized block.

It can't be used by full nodes at the moment, because they're missing the storage of the finalized block.
However, we can add a system that makes full nodes download the storage from other nodes, for instance through the existing storage proofs system.

tomaka:
cc #576

tomaka:
cc https://github.com/paritytech/smoldot/pull/1627

PR #2876, opened 2022-10-14
Implement downloading the chain storage

tomaka:
cc https://github.com/paritytech/smoldot/issues/165

Adds a mode to the `warp_sync` where, instead of downloading just the runtime and then the consensus-related information, we download the entire storage of the chain.

To achieve this, the `StorageGet` requests have been renamed to `StorageGetKeys`, and a new `StorageGetRange` request has been added.

Works for Westend. For Polkadot I'm getting `UnexpectedEntriesCount` errors which I can't manage to solve at the moment.
This needs to be solved, and this PR needs a bit of clean up, before being ready for review/merging. However it is 90% ready.


tomaka:
Another thing to note is that at the moment we use the "no proof" version of the network protocol.
However, as I'm implementing this PR, I've stumbled upon some tricky safety issues that would be solved if we used the "with proof" version of the protocol.
So we might change this before merging.

tomaka:
This is blocked on child tries support, because Polkadot has child tries.

tomaka:
Doesn't work due to the lack of child tries support.

tomaka:
I'm going to update this PR to use the "proof version" of the request, and afterwards I would like to merge it.
We're going to need state requests in order to properly implement the `state_getKeysPaged` RPC function without running into situations where the storage response is too large.

Issue #2799, opened 2022-09-29
Peers::set_peer_notifications_out_desired doesn't generate NotificationsOutResult/NotificationsOutClose events


tomaka:
Not fixed after https://github.com/paritytech/smoldot/pull/2795
A bit tricky.

PR #2880, opened 2022-10-16
Add fulfilled_undesired_outbound_substreams and close_out_notification

tomaka:
Close https://github.com/paritytech/smoldot/issues/2799

Before this PR, calling `set_peer_notifications_out_desired` and marking a peer+substream as "not desired" instantly closes any existing substream or substream in progress without generating any event.

After this PR, the function just flags the substream as "not desired", and the user has to additionally call `fulfilled_undesired_outbound_substreams` to determine the list of substreams that are open but not desired, then `close_out_notification` to close the substream.
This mimics the existing API with `unfulfilled_desired_outbound_substream` and `open_out_notification`.

I don't think (but I'm not sure) that this PR actually changes the behavior anywhere, because right now we mark peers as undesired only as a consequence of a connection shutting down, a connection attempt failing, or as a consequence of a substream having been closed. However, it is likely that this PR fixes some bugs in some niche situations.

### DIFF
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -1024,6 +1024,9 @@ where
     /// When a combination of network protocol and [`PeerId`] is marked as "desired", it will
     /// be returned by [`Peers::unfulfilled_desired_outbound_substream`].
     ///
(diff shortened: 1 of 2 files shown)


melekes (approved):
👍

PR #2881, opened 2022-10-16
Implement keystore storing keys on disk

tomaka:
The keystore can now save and reload keys, as intended.

### DIFF
--- a/src/identity/keystore.rs
+++ b/src/identity/keystore.rs
@@ -15,15 +15,32 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-// TODO: doc
+//! Data structure containing cryptographic key pairs.
+//!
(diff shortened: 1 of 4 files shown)


src/identity/keystore.rs
```diff
@@ -205,15 +420,39 @@ impl Keystore {
...
+                        PrivateKey::MemorySr25519(key) => Cow::Borrowed(key),
+                        PrivateKey::FileSr25519 => {
+                            match Self::load_sr25519_from_file(
+                                self.path_of_key_sr25519(key_namespace, public_key).unwrap(),
+                            )
+                            .await
+                            {
+                                Ok(key) => {
+                                    drop(guarded);
+                                    // TODO: is creating the signing context expensive?
```
melekes, line 444: seems unrelated

melekes (approved):

src/identity/keystore.rs
tomaka: Ah right, I copy-pasted some code and this comment moved at the wrong place

PR #2882, opened 2022-10-16
Update Rust version to v1.64

tomaka:
https://github.com/paritytech/smoldot/pull/2881 currently fails CI because it uses a function that wasn't stable yet in Rust 1.61.
While I could bypass the issue, it makes sense to just update the Rust version.


tomaka:
Will merge the main branch after https://github.com/paritytech/smoldot/pull/2884, to see if this size difference is still there.

PR #2883, opened 2022-10-16
Fail if wasm-opt isn't installed when passing `--release`

tomaka:
When building the Wasm file to embed in the browser, and `--release` is provided, we try to run `wasm-opt` on it.
If `--debug` is provided, we don't do it.

Because we used to do it with `--debug` as well, there exists a fallback if `wasm-opt` isn't installed that will simply not run `wasm-opt`.

This PR removes this fallback, so that the release build fails if `wasm-opt` isn't installed. `--debug` will still work even if it's not installed.


tomaka:
CI will pass after https://github.com/paritytech/smoldot/pull/2883

melekes (approved):
👍

PR #2884, opened 2022-10-16
Add wasm-opt when checking wasm node size diff

tomaka:
It turns out that it wasn't installed, completely skewing the results.

I've also removed `cmake`, which was still there from before we switched to our own protobuf codec, thereby fixing a TODO.


tomaka:
Opened https://github.com/paritytech/smoldot/issues/2886 because the size diff being 0 is weird.

PR #2888, opened 2022-10-17
Reduce the allocations when emitting logs

tomaka:
Instead of turning things immediately into a string, we instead put the values provided to the host function into the `LogEmit` struct, and do the formatting and turning into string later.

This unlocks a potential future change where the API user can format the runtime logs the way they want.


melekes (approved):
👍

Issue #2893, opened 2022-10-17
nested storage transaction not supported


xlc:
https://github.com/paritytech/smoldot/blob/b27b25bc792a74d628cd1210e6f2cb55e9ad5aef/src/executor/host.rs#L1076-L1082

PR #2894, opened 2022-10-17
support nested transactions

xlc:
Closes https://github.com/paritytech/smoldot/issues/2893

### DIFF
--- a/src/executor/host.rs
+++ b/src/executor/host.rs
@@ -562,6 +562,10 @@ pub enum HostVm {
     /// Ends a storage transaction. All changes made to the storage (e.g. through a
     /// [`HostVm::ExternalStorageSet`]) since the previous
     /// [`HostVm::StartStorageTransaction`] must be rolled back if `rollback` is true.
+    ///
+     /// Guaranteed by the code in this module to never happen if no transaction is in progress.
+     /// If the runtime attempts to end a non-existing transaction, an [`HostVm::Error`] is
+     /// generated instead.
     EndStorageTransaction {
         /// Object used to resume execution.
         resume: EndStorageTransaction,
--- a/src/executor/host.rs
+++ b/src/executor/host.rs
@@ -486,7 +486,7 @@ impl HostVmPrototype {
                 allow_unresolved_imports: self.allow_unresolved_imports,
                 memory_total_pages: self.memory_total_pages,
                 registered_functions: self.registered_functions,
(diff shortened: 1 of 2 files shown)


src/executor/runtime_host.rs
```diff
@@ -473,13 +473,13 @@ struct Inner {
     /// Pending changes to the top storage trie that this execution performs.
     top_trie_changes: storage_diff::StorageDiff,
 
-    /// `Some` if and only if we're within a storage transaction. When changes are applied to
+    /// Contains pending storage reverts if and only if we're within a storage transaction. When changes are applied to
     /// [`Inner::top_trie_changes`], the reverse operation is added here.
```
tomaka, line 477:
```suggestion
    /// Contains pending storage reverts if and only if we're within a storage transaction.
    /// When changes are applied to [`Inner::top_trie_changes`], the reverse operation is
    /// added here.
```

Nit formatting

src/executor/host.rs
```diff
@@ -1074,39 +1070,32 @@ impl ReadyToRun {
...
             HostFunction::ext_storage_start_transaction_version_1 => {
-                if self.inner.within_storage_transaction {
-                    return HostVm::Error {
-                        error: Error::NestedTransaction,
-                        prototype: self.inner.into_prototype(),
-                    };
-                }
-
-                self.inner.within_storage_transaction = true;
+                self.inner.storage_transaction_depth += 1;
```
tomaka, line 1073:
```suggestion
                // TODO: a maximum depth is important in order to prevent a malicious runtime from crashing the client, but the depth needs to be the same as in Substrate; figure out
                self.inner.storage_transaction_depth += 1;
```

I don't think it's important to enforce a maximum depth right now, but I think we eventually should. Since it's a non-trivial question, I'd go for just a todo.

tomaka (commented):
:+1:

src/executor/host.rs
xlc: There is a limit enforced by Substrate. I can add it here https://github.com/paritytech/substrate/blob/e8a4408c53ae2a1a0ef6cb2f7b603887a231c16f/frame/support/src/storage/transactional.rs#L37

xlc: But that's enforced at runtime level not client level so I guess I shouldn't enforce it

melekes (approved):
👍

tomaka (approved):

PR #2895, opened 2022-10-18
expose storage diff

xlc:
so I can inspect the content


tomaka:
There is already a `diff_iter_unordered` function, which I think is what you want?
I'd be ok with adding `diff_into_iter_unordered` if it's useful.

tomaka (approved):

PR #2896, opened 2022-10-18
Implement WebRTC messages framing

tomaka:
cc https://github.com/libp2p/specs/pull/412

This PR finishes implementing the WebRTC spec by adding the last remaining item: the messages framing.

Implementing this messages framing while minimizing the amount of data copies is rather challenging.
Instead of going for the complicated solution, I went for the more easy solution of having an intermediate read buffer where data is first copied.
Going for the simple solution decreases the chances of bugs and increases the ease of debugging, so it's preferable at the moment.

In the future, once WebRTC is fully working, we can rewrite this messages framing code in a more optimized way.

### DIFF
--- a/bin/wasm-node/javascript/src/index-browser.ts
+++ b/bin/wasm-node/javascript/src/index-browser.ts
@@ -256,17 +256,17 @@ export function start(options?: ClientOptions): Client {
               "v=0" + "\n" +
               // Identifies the creator of the SDP document. We are allowed to use dummy values
               // (`-` and `0.0.0.0`) to remain anonymous, which we do. Note that "IN" means
-              // "Internet". (RFC8866)
+              // "Internet" (and not "input"). (RFC8866)
               "o=- 0 0 IN IP" + ipVersion  + " " + targetIp + "\n" +
               // Name for the session. We are allowed to pass a dummy `-`. (RFC8866)
               "s=-" + "\n" +
               // Start and end of the validity of the session. `0 0` means that the session never
               // expires. (RFC8866)
               "t=0 0" + "\n" +
-              // A lite implementation is only appropriate for devices that will
-              // *always* be connected to the public Internet and have a public
+              // A non-lite implementation is only appropriate for devices that will
+              // always be connected to the public Internet and have a public
               // IP address at which it can receive packets from any
-              // correspondent.  ICE will not function when a lite implementation
+              // correspondent.  ICE will not function when a non-lite implementation
               // is placed behind a NAT (RFC8445).
               "a=ice-lite" + "\n" +
               // A `m=` line describes a request to establish a certain protocol.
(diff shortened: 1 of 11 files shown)


tomaka:
This PR turned into a bit of a "fix every WebRTC issue" PR, but it is much more productive to just push everything onto a branch instead of analyze each bug individually.

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -256,29 +256,29 @@ export function start(options?: ClientOptions): Client {
...
+              // "Internet" (and not "input"). (RFC8866)
               "o=- 0 0 IN IP" + ipVersion  + " " + targetIp + "\n" +
               // Name for the session. We are allowed to pass a dummy `-`. (RFC8866)
               "s=-" + "\n" +
               // Start and end of the validity of the session. `0 0` means that the session never
               // expires. (RFC8866)
               "t=0 0" + "\n" +
-              // A lite implementation is only appropriate for devices that will
-              // *always* be connected to the public Internet and have a public
+              // A non-lite implementation is only appropriate for devices that will
```
melekes, line 266: ??? This was a copy from the specification. Did something change?

bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -303,8 +304,9 @@ export function start(options?: ClientOptions): Client {
               // (UDP or TCP)
               "a=sctp-port:5000" + "\n" +
               // The maximum SCTP user message size (in bytes) (RFC8841)
-              "a=max-message-size:100000" + "\n" +
-              // A transport address for a candidate that can be used for connectivity checks (RFC8839).
+              "a=max-message-size:16384" + "\n" +  // TODO: should this be part of the spec?
```
melekes, line 307: subject to a change, but yes!

bin/wasm-node/javascript/src/index-browser.ts
tomaka: ? The specification doesn't contain this text. It's me who wrote it. It has a mistake in it, it says that lite implementations are only for publicly-accessible devices. But no, it's non-lite implementations that are for public devices. We're not a public device, so we use the lite implementation.

bin/wasm-node/javascript/src/index-browser.ts
melekes: https://github.com/libp2p/specs/pull/412#issuecomment-1283650780

bin/wasm-node/javascript/src/index-browser.ts
melekes: "Certain ICE agents will always be connected to the public Internet
   and have a public IP address at which it can receive packets from any
   correspondent.  To make it easier for these devices to support ICE,
   ICE defines a special type of implementation called "lite" (in
   contrast to the normal full implementation).  Lite agents only use
   host candidates and do not generate connectivity checks or run state
   machines, though they need to be able to respond to connectivity
   checks." https://datatracker.ietf.org/doc/rfc8445/

melekes: By specifying `ice-lite` in server's SDP (that we're pretending we've received through some channel, but instead construct manually), we're saying that server is publicly accessible.

tomaka: To be honest this text doesn't make sense to me no matter whether it says "lite" or "not lite", so I've just reverted it :shrug:

melekes (approved):

PR #2898, opened 2022-10-19
Add two caches to the `peers` state machine

tomaka:
The `Peers` state machine has two functions: `unfulfilled_desired_outbound_substreams` and `fulfilled_undesired_outbound_substreams`. They return a list of peer-substreams tuples.

These two functions are called very frequently (every time something happens on the networking), they are `O(n)` (they iterate over every single substream of every single peer), and most of the time they return nothing.

This PR adds two new fields to the `peers` state machine with the same names as these functions.
The functions now simply return the content of the field, and the field is updated pro-actively whenever something else changes.

### DIFF
--- a/src/libp2p/peers.rs
+++ b/src/libp2p/peers.rs
@@ -239,11 +239,11 @@ where
             ),
             peers_notifications_out: BTreeMap::new(),
             unfulfilled_desired_outbound_substreams: hashbrown::HashMap::with_capacity_and_hasher(
-                0, // TODO: capacity?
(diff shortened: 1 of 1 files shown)


tomaka:
Thanks to the debug_asserts!, I'm quite confident that the logic is correct. Every single time we make sure that the value is in there if it should be in there, or not in there if it shouldn't be in there. All the debug_asserts are run pretty frequently.

melekes (approved):
👍

PR #2909, opened 2022-10-20
Simplify the database content by restoring chain information from the database

### LINKED ISSUE #1752: Full node should load consensus data from chain storage instead of storing it in database

tomaka:
Close https://github.com/paritytech/smoldot/issues/1752

This PR removes a lot of fields from the database, as the values of these fields is now computed from the runtime of the finalized block when the client initialized.

While this PR is fully ready, I'm opening it as a draft. Unfortunately, the runtime functions used by the chain information builder weren't always available, and so the loading might fail. This needs to be first figured out.


tomaka:
Blocked on https://github.com/paritytech/smoldot/issues/2930

tomaka:
Since https://github.com/paritytech/smoldot/issues/2930 is not doable, this is instead blocked on https://github.com/paritytech/smoldot/issues/165.

Issue #2782, opened 2022-09-26
WebSocket is already in CLOSING or CLOSED state.


tomaka:
Running smoldot in Chrome leads to errors in the console:

> WebSocket is already in CLOSING or CLOSED state.

The backtraces show that this happens when trying to send data on said WebSocket.

PR #2910, opened 2022-10-20
Make a few WebRTC-related assumptions more explicit

tomaka:
WebRTC has a few assumptions, most notably the fact that the reading and writing side of substreams must never close (in other words: the browser can't control the individual closing of the reading and writing, so the code within smoldot expects them to remain always open).

Since this is a property of WebRTC specifically and not of multi-stream connections in general, I've opted to keeping an API that allows closing the reading/writing and simply documenting that this is not allowed.

An alternative could have been for the internals of smoldot to accept in its API that the reading/writing was closed, and if that actually happens consider this as a protocol violation and reset the substream. I've however opted to not do that because of the "external" nature of multi-stream connections. The encryption and multiplexing are already ensured by the JS code, and thus it makes sense to include as part of the responsibilities of the JS code to never report the substream being closed.

As part of this PR, I've found the cause for https://github.com/paritytech/smoldot/issues/2782
Because the fix I have in mind would conflict with this PR, I'll do it afterwards.

### DIFF
--- a/bin/light-base/src/network_service/tasks.rs
+++ b/bin/light-base/src/network_service/tasks.rs
@@ -172,7 +172,7 @@ pub(super) async fn connection_task<TPlat: Platform>(
             .await
         }
         either::Right((socket, task)) => {
-            multi_stream_connection_task::<TPlat>(
+            webrtc_multi_stream_connection_task::<TPlat>(
                 socket,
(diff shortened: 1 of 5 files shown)


melekes (approved):
👍

PR #2922, opened 2022-10-24
Verify that SDP offer contains UDP

tomaka:
The comment concerning `m=application` at the bottom says that we tweak the local offer. This isn't true anymore, but I think it's a good idea to make sure that the local offer does contain a UDP + data channel entry.

Tested on Firefox and Chrome. Firefox somehow doesn't work anymore (ugh) but this PR doesn't actually modify any behavior, so it can't be related.

### DIFF
--- a/bin/wasm-node/javascript/src/index-browser.ts
+++ b/bin/wasm-node/javascript/src/index-browser.ts
@@ -261,13 +261,18 @@ export function start(options?: ClientOptions): Client {
       pc.onnegotiationneeded = async (_event) => {
           // Create a new offer and set it as local description.
           let sdpOffer = (await pc!.createOffer()).sdp!;
+          // We check that the locally-generated SDP offer has a data channel with the UDP
+          // protocol. If that isn't the case, the connection will likely fail.
+          if (sdpOffer.match(/^m=application(\s+)(\d+)(\s+)UDP\/DTLS\/SCTP(\s+)webrtc-datachannel$/m) === null) {
+            console.error("UDP local offer doesn't contain UDP data channel. WebRTC connections will likely fail. Please report this issue.");
+          }
           // According to the libp2p WebRTC spec, the ufrag and pwd are the same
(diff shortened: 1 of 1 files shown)


bin/wasm-node/javascript/src/index-browser.ts
```diff
@@ -261,13 +261,18 @@ export function start(options?: ClientOptions): Client {
       pc.onnegotiationneeded = async (_event) => {
           // Create a new offer and set it as local description.
           let sdpOffer = (await pc!.createOffer()).sdp!;
+          // We check that the locally-generated SDP offer has a data channel with the UDP
+          // protocol. If that isn't the case, the connection will likely fail.
+          if (sdpOffer.match(/^m=application(\s+)(\d+)(\s+)UDP\/DTLS\/SCTP(\s+)webrtc-datachannel$/m) === null) {
+            console.error("UDP local offer doesn't contain UDP data channel. WebRTC connections will likely fail. Please report this issue.");
```
melekes, line 267:
```suggestion
            console.error("local offer doesn't contain UDP data channel. WebRTC connections will likely fail. Please report this issue.");
```

melekes (commented):
are u sure that webrtc fails if local offer doesn't contain `webrtc-datachannel` string?

tomaka:
> are u sure that webrtc fails if local offer doesn't contain webrtc-datachannel string?

Well, I don't know if it does, but it's not supposed to work. The spec say that the `m=application` string of the SDP response must match the one in the offer. If the offer doesn't contain `webrtc-datachannel`, then the response violates the spec.

Note that `UDP/DTLS/SCTP` and `webrtc-datachannel` are specced by RFC 8841 section 4.4.2. It's not just a string that the browser decides arbitrarily.

melekes:
I was under the impression that it's only valid for remote offer (that we're pretending we've received once we get a ICE binding request).

Anyways, I guess it doesn't hurt to add this check.

melekes (approved):

PR #2925, opened 2022-10-24
Clarify platform reading API

tomaka:
At the moment, the `read_buffer` function of the `Platform` trait (representing how the light client binds to the environment) returns either `Some` in normal situations or `None` if the connection has been either closed or reset.

This PR turns this `Option` into an enum with three variants: `Open`, `Closed`, and `Reset`.
This lets the code properly differentiate between these two latter situations.

Additionally, I've renamed "closed" to "reset" in the entire Wasm<->JS code (all the functions and variables), for clarity. While the `Platform` trait lets you report that the reading side is closed but the writing side still open, the JS code can't represent that state. Instead, it's either "connection open" or "connection reset".

This PR fixes https://github.com/paritytech/smoldot/issues/2782. The issue was that we were treating `read_buffer()` returning `None` as meaning "reading side closed", but actually the entire connection was already reset. We would then call `send` on the already-reset connection, leading to the error. This kind of confusion is exactly why I've performed the rename.

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -7,6 +7,10 @@
 - The `payment_queryInfo` JSON-RPC function now works with runtimes that have defined the type of `Balance` to be less than 16 bytes. ([#2914](https://github.com/paritytech/smoldot/pull/2914))
 - The parameter of `chainHead_unstable_finalizedDatabase` has been renamed from `max_size_bytes` to `maxSizeBytes`. ([#2923](https://github.com/paritytech/smoldot/pull/2923))
 
+### Fixed
+
+- Fix errors showing in the browser's console about WebSockets being already in the CLOSED or CLOSED state.
+
 ## 0.7.3 - 2022-10-19
 
 ### Changed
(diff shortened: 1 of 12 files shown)


bin/wasm-node/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
 - The `payment_queryInfo` JSON-RPC function now works with runtimes that have defined the type of `Balance` to be less than 16 bytes. ([#2914](https://github.com/paritytech/smoldot/pull/2914))
 - The parameter of `chainHead_unstable_finalizedDatabase` has been renamed from `max_size_bytes` to `maxSizeBytes`. ([#2923](https://github.com/paritytech/smoldot/pull/2923))
 
+### Fixed
+
+- Fix errors showing in the browser's console about WebSockets being already in the CLOSED or CLOSED state. ([#2925](https://github.com/paritytech/smoldot/pull/2925))
```
tomaka, line 12:
```suggestion
- Fix errors showing in the browser's console about WebSockets being already in the CLOSING or CLOSED state. ([#2925](https://github.com/paritytech/smoldot/pull/2925))
```

melekes (approved):

PR #2935, opened 2022-10-27
Install clang before the spellcheck

tomaka:
Our spellcheck CI job has started failing yesterday because compiling `clang-sys` can't find LLVM installed on the system.

I was curious why this suddenly happened. Unfortunately, we were caching the compiled version of `cargo-spellcheck`, and suddenly the cache expired, which means that we're now compiling it whereas we weren't compiling it before. I don't really have the motivation to find the CI run where we initially compiled it in order to compare it with the failing CI run, and even if I found it the logs are probably expired now.

A quick investigation of the cargo-spellcheck and clang-sys packages don't reveal anything abnormal.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -201,7 +201,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - run: apt-get update && apt install -y clang  # LLVM required by cargo-spellcheck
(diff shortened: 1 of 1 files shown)


tomaka:
I don't understand the problem.

They're depending on clang 6, which is super old: https://github.com/rust-lang/rust-bindgen/blob/d492ac4ae11ba7403acd533813982b052c5020a0/Cargo.toml#L53

Issue #2900, opened 2022-10-19
WebRTC opening failure panic


tomaka:
When a WebRTC connection fails to open, there seems to be a panic here:

https://github.com/paritytech/smoldot/blob/bc36d663d98d17598a179a788f1d631473d45acc/bin/wasm-node/rust/src/platform.rs#L705

tomaka:
The problem is simply caused by the fact that the browser calls the `close` or `error` callback on the data channel when it fails to open.

Because we only register the data channel after the `open` callback is called, it's not there yet.

This is a bit tricky to solve, because the definition of "open" of a WebRTC connection is a bit hacky.

tomaka:
I think that this is simply solved by making `connection_stream_open` return the stream ID, and make `connection_stream_opened` interpret already-assigned IDs as "outbound" and unknown IDs as "inbound".

tomaka:
I'll wait for https://github.com/paritytech/smoldot/pull/2925 in order to not introduce annoying merge conflicts.

tomaka:
Actually, the idea is that we should just try opening the substream again.
There is no gain is actually tracking the pending substream openings compared to using a timeout.

tomaka:
> Related comment from the spec:

That's not really related.
The problem here was that Firefox calls `onclose` or `onerror` on the handshake substream because we fail to reach the server, and our implementation didn't consider this possibility. I was assuming that `onclose`/`onerror` would only be called after the substream has been opened, and that we'd get a `onconnectionstatechange` event instead for the reach failure.

PR #2936, opened 2022-10-27
Properly handle data channel opening failures

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2900

In order to open outbound substreams on a WebRTC connection, the mechanism that the API of smoldot uses is as follows: one function to ask the implementation to open a substream, and one callback that is later called in order to notify that a substream has been fully opened.

The API of smoldot doesn't provide a way to report that opening an outbound substream has failed. This is because this isn't really supposed to happen. If you have a connection, opening a substream is purely a "local side change": you allocate an ID, then send a message. It should maybe not even be asynchronous, but that's off-topic here.

The problem here is that this mental model doesn't really fit what the browser does. In particular, the browser does nothing until a data channel is first opened, plus the browser can call `onclose` or `onerror` on the data channel before `onopen`. To accomodate for that, the current code is a bit hacky in the sense that it reports the connection as immediately open.

This PR does it a bit more differently and more properly: we immediately open a data channel at the same time as we create the `RTCPeerConnection` in order for the browser to start connecting, even though smoldot didn't request a channel yet. We store this first data channel on the side, and when smoldot asks for a data channel we provide this one.
If a data channel closes while it is still opening (the handshake one or another), we just kill the entire connection.

### DIFF
--- a/bin/light-base/src/platform.rs
+++ b/bin/light-base/src/platform.rs
@@ -80,6 +80,11 @@ pub trait Platform: Send + 'static {
     /// Queues the opening of an additional outbound substream.
     ///
     /// The substream, once opened, must be yielded by [`Platform::next_substream`].
+    ///
+    /// > **Note**: No mechanism exists in this API to handle the situation where a substream fails
+    /// >           to open, as this is not supposed to happen. If you need to handle such a
+    /// >           situation, either try again opening a substream again or reset the entire
+    /// >           connection.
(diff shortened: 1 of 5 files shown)


tomaka:
I have a headache from the complexity brought by the fact that there are dozens and dozens of asynchronous functions and that things can be called in many different orders. But I think that such is the nature JS and there's no great way to make the code more easy.

PR #2951, opened 2022-10-31
Make Grandpa warp sync protocol code more zero cost

tomaka:
We no longer copy the fragments when decoding a response sent by a peer, but return the message as a whole. This avoids copies.

Unfortunately, we have to do `to_vec()` at some point, so this doesn't actually save anything. However this is still progress towards a final zero-cost version.


melekes (approved):
nice 👍

PR #2954, opened 2022-10-31
Make it mandatory to specify `#[required]`

tomaka:
Whether a field is required or optional when decoding protobuf can be a bit tricky. I wrote a small paragraph about this and made it mandatory to specify `#[required]` instead of it being the default.


melekes (approved):
👍

PR #2968, opened 2022-11-03
Update to Rust v1.65 and use its new features

tomaka:
Updates to Rust 1.65. As usual, the CI will fail because the official Rust Docker image generally takes a day or two to be updated when a new Rust version comes out.

This PR introduces GATs in the `Platform` trait.
At the moment, if you call `Platform::next_substream` or `Platform::next_substream`, you obtain a `Future` that is completely disconnected from the object representing the connection or the stream. This means that you can for example drop the connection or stream object while the `Future` is still alive, or you can call this function multiple times with the same connection/stream.
Using GATs allows us to bind the `Future` to the object it refers to, and thus removes these annoying corner cases. This should also make the trait implementation much easier to write, but I'm not touching this complicated code right now.

I've switched some code to use `let-else`, but that's just for fun and this feature is in practice not super useful.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
   test-64bits:
     runs-on: ubuntu-latest
     container:
-      image: rust:1.64
+      image: rust:1.65
     steps:
     - uses: actions/checkout@v3
     - uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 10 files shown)


tomaka:
The bot shows a -906kB difference, however locally I see roughly the same Wasm size before and after this PR.

Issue #2970, opened 2022-11-04
WebRTC panic


tomaka:
> Smoldot has panicked while executing task `connections`. This is a bug in smoldot. Please open an issue at https://github.com/paritytech/smoldot/issues with the following message:
panicked at 'assertion failed: matches!(STATE.try_lock().unwrap().connections.get(connection_id).unwrap().inner,\n    ConnectionInner :: MultiStreamWebRtc { .. })', bin/wasm-node/rust/src/platform.rs:238:9

PR #2978, opened 2022-11-07
Don't call connection_stream_open if WebRTC connection has been reset

tomaka:
Fix https://github.com/paritytech/smoldot/issues/2970

The browser can reset a WebRTC connection at any time. When that happens, smoldot only switches the connection state to "reset", and it's only later that the connection task will query the state and see that it's reset.
However, in the meanwhile, the connection task can try to open a new substream. This PR clarifies the behavior when that happens: trying to open a new substream should simply be ignored.


tomaka:
This is a good example of a `debug_assert!` that helped detect a bug.

PR #2992, opened 2022-11-09
Extract database encoding/decoding to a separate module

tomaka:
Unclogs `lib.rs` a bit.


bin/light-base/src/database.rs
```diff
@@ -0,0 +1,182 @@
...
+        if serialized.len() <= max_size {
+            // Success!
+            return serialized;
+        }
+
+        if database_draft.nodes.is_empty() {
+            // Can't shrink the database anymore. Return the string `"<too-large>"` which will
+            // fail to decode but will indicate what is wrong.
+            let dummy_message = "<too-large>";
+            return if dummy_message.len() >= max_size {
```
melekes, line 110:
```suggestion
            return if dummy_message.len() > max_size {
```

to be consistent with the similar check above

melekes (approved):

PR #2993, opened 2022-11-09
Run npm update

tomaka:
I don't know why dependabot isn't handling that, but I've just run `npm update` and noticed a huge diff, so here it is.


tomaka:
These are mostly indirect dependencies from `typescript`, a dev-dependency.
We fortunately have very little actual dependencies.

melekes (approved):

Issue #2996, opened 2022-11-10
setTimeout has a minimum delay of 1 second (or 15mn on mobile) in inactive tabs


tomaka:
https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#timeouts_in_inactive_tabs

This was more or less known to me, but it's not something that we've actively taken into account until now.

The problem is that smoldot sometimes intentionally queues a task using `setTimeout(..., 0)` in order to intentionally yield back control to the browser so that it can run its events loop.
For example, when verifying warp sync fragments, smoldot will intentionally give control back to the browser between each fragment. There can be hundreds of fragments to verify. This wouldn't be a problem if the time between two fragments is very small, but if the minimum delay is 1 second, then it takes literally 100 seconds for an operation that would normally take 100ms or so.

Even when the tab is in the foreground, after 5 `setTimeout`s in a row, the minimum delay becomes 4ms:
https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#nested_timeouts
Again, this is problematic if there are hundreds of fragments to verify one after the other.

tomaka:
Looking at the code, the situation is quite bad. For example, when a timer is finished or a message arrives on a socket, we don't immediately process it but instead we call `setTimeout(..., 0)` and the callback does the processing.

In general, the code assumes everywhere that `setTimeout(..., 0)` works as advertised.

PR #2999, opened 2022-11-11
Refactor the execution to avoid setTimeout

tomaka:
Close https://github.com/paritytech/smoldot/issues/2996

This PR does many things:

- Remove the `crate::util::yield_twice` function in favor of a platform-specific function `TPlat::yield_after_cpu_intensive`. This way, the implementation of the yielding, which is a bit hacky due to the requirements of the Wasm node, is specific to the Wasm node.

- Refactor the `nextJsonRpcResponse` function and the `jsonRpcResponsesNonEmptyCallback` callback. Right now, when we wait for a JSON-RPC response to come, we wait for smoldot to notify that it has a response available. But because this notification is done through a callback, we can't immediately query the content of the response, so instead we use `setTimeout(..., 0)` to register a task that queries the content of the response and then notify the `Promise` that was waiting. Now we instead immediately notify a `Promise`, and this `Promise` then queries the content of the response as an aftermath.

- In the JS code, add a `registerShouldPeriodicallyYield` function whose implementation is "platform-specific" (browsers, NodeJS, Deno). This function controls a new setting called "should periodically yield". In NodeJS and Deno, this is always `false`. In the browser, this is `false` if and only if the page is in the background and `true` if it is in the foreground. Passing `false` reduces the usage of `setTimeout`, while passing `true` promotes the usage of `setTimeout` to avoid blocking the browser.

- In the JS <-> Rust bindings, add a `periodically_yield` parameter to `init` and add a new `set_periodically_yield` function. See above point.

- Refactor the way the execution is done. Currently we spawn at initialization a task (using `setTimeout(..., 0)`) that runs everything. When something needs to be executed, this task wakes up by calling `setTimeout(..., 0)` again. After this PR, there is no "background task" anymore. Instead, the `Future` that drives everything is now put in `CLIENT` variable, and every single binding function calls a new `advance_execution` function before returning. `advance_execution` polls that `Future` and, depending on the `periodically_yield` setting, either executes everything until there is nothing more to do, or executes until the first yield.

All these changes eliminate the usage of `setTimeout` to the strict minimum, which is when we actually need a timer.

### DIFF
--- a/bin/light-base/src/json_rpc_service.rs
+++ b/bin/light-base/src/json_rpc_service.rs
@@ -590,7 +590,7 @@ impl<TPlat: Platform> Background<TPlat> {
 
                         // We yield once between each request in order to politely let other tasks
                         // do some work and not monopolize the CPU.
-                        crate::util::yield_twice().await;
+                        TPlat::yield_after_cpu_intensive().await;
                     }
                 }
                 .boxed(),
--- a/bin/light-base/src/platform.rs
+++ b/bin/light-base/src/platform.rs
@@ -24,6 +24,7 @@ pub mod async_std;
 /// Access to a platform's capabilities.
 pub trait Platform: Send + 'static {
     type Delay: Future<Output = ()> + Unpin + Send + 'static;
+    type Yield: Future<Output = ()> + Unpin + Send + 'static;
     type Instant: Clone
         + ops::Add<Duration, Output = Self::Instant>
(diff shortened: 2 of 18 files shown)


tomaka:
Testing this PR is a bit complicated.
However, before this PR, when I leave smoldot running in a background tab, I get disconnects and timeouts.
After this PR, this no longer happens, so I would say that it works.

melekes (approved):

PR #3019, opened 2022-11-21
Use wrapping_sub in bandwidth calculation

tomaka:
It happened. I received more than 4 GiB while smoldot was running, and it lead to a panic in the memory printer because the total bytes transferred overflowed.
The fix is simple: just use `wrapping_sub`.

This panic only happens if debug_assertions are enabled, which is the case locally but not in the published version, so this panic isn't worth putting in the CHANGELOG.


tomaka:
I've pushed another commit that changes their type from `AtomicUsize` (which is 32bits in practice) to `AtomicU64`, because there's actually a non-negligible chance that more than 4 GiB could be transferred within 60 seconds, for example if connected through `localhost`.

Issue #1531, opened 2021-10-27
Wasm node passes wrong value for reachability in pending_outcome_err


tomaka:
The Wasm node passes the "unreachable" flag if the FFI layer returns an error on connecting to a peer, but in practice errors can happen as well if the browser for example decides that the connection has timed out.

Issue #2746, opened 2022-09-14
Brainstorm and refactor the discovery process and address book


tomaka:
There exists multiple systems that are spread between the network services and the `network/service`:

- "Out slots" are assigned by the network services following the suggestions of the `network/service`.
- Kbuckets, used to determine which PeerIds to hold (prevents memory explosion) and answer Kademlia requests.
- For each peer, a list of addresses and whether we are connecting or connected to it.
- A ban system in network services, where we prevent connections to the same peer for 20 seconds after a disconnect.

This is in general overly complicated and hard to prove correct.
Additionally, the network services don't have enough control over this process.

The whole thing should be clarified and simplified in the code.

PR #3023, opened 2022-11-22
Clarify meaning of is_unreachable

tomaka:
Close https://github.com/paritytech/smoldot/issues/1531
cc https://github.com/paritytech/smoldot/issues/2746

This PR renames `is_unreachable` to `is_bad_address`.
It is generally not possible to know whether an address is definitely unreachable or not. The only situation where we can affirm that an address is definitely unreachable is if our implementation doesn't support a certain protocol that it uses.

This matches what the browser node already does anyway.


tomaka:
For what it's worth, I don't really like the API of `pending_outcome_err` as it is a bit too vague/magic. This is covered by https://github.com/paritytech/smoldot/issues/2746

Issue #2973, opened 2022-11-07
How to decode storage proof into key-value pairs and encode them to storage proof?


xlc:
Use case: I want to mock the validation data from relaychain by decoding the trie proof, modify the value, and generate new proof and rebuild the block.

I am trying to find the right API to use in https://github.com/paritytech/smoldot/tree/main/src/trie but I can only find how to verify proof by particularly decoding it, how to create trie and get state root. But not decode proof into key-value pairs and encode the key-value pairs into storage proof.

Does smoldot have this API available? Otherwise I may need to use Substrate to do that, which comes with lot more other issues.

tomaka:
There's no way to do that at the moment. The functions are really tailored to the light client use case, and the light client never encodes a proof.

tomaka:
I would like to add some way to get the list of all elements of a proof, though. Right now, we iterate through the storage proof every time a runtime call tries to access a value, which has a bad `O` complexity.

xlc:
This is what I have figured. Let's see if smoldot implements this first of cumulus have its' issue fixed first...

tomaka:
Encoding a proof should be relatively easy by combining the `trie_structure` module and `proof_node_codec::encode`. I'm just not sure where I would put this proof encoding function.

However, `proof_node_codec::encode` only supports non-compact proofs. I'm honestly lost as to which protocols use a compact proof and which use a non-compact proof.

tomaka:
One interesting thing to note is that when making a runtime call against a compact proof, it could in theory be more efficient to just leave the proof as it is and re-decode it every time. This is because a compact proof is basically a tree.
In other words, if the proof contains for example 1000 entries but we only need 2, then it is possible to access only these 2 entries in `O(log n)` time.

However, we can only do this if the proof is correct, and verifying that the proof is correct is `O(n)` anyway, so the best solution is simply to verify the entire proof and keep a list of where in the proof each storage value is found.

tomaka:
After https://github.com/paritytech/smoldot/pull/3021, all that remains to do is:

- Add some getters to `DecodedTrieProof` to get the list of entries in the proof (I'm not doing it now to avoid merge conflicts).
- Add the proof encoding code. As I've mentioned above, it's quite easy to write if you know how proofs are encoded, as all the primitives are still there.

tomaka:
The encoding is actually more complicated than expected, because we need to provide the hashes of nodes that are in the trie but not in the proof. This makes an API kind of complicated to design.

[...]

PR #3025, opened 2022-11-23
Add iteration functions to DecodedTrieProof

tomaka:
cc https://github.com/paritytech/smoldot/issues/2973

### DIFF
--- a/src/trie/proof_decode.rs
+++ b/src/trie/proof_decode.rs
@@ -304,6 +304,92 @@ pub struct DecodedTrieProof<T> {
 }
 
 impl<T: AsRef<[u8]>> DecodedTrieProof<T> {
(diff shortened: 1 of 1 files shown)


src/trie/proof_decode.rs
```diff
@@ -304,6 +304,95 @@ pub struct DecodedTrieProof<T> {
...
+            Some((key, value))
+        })
+    }
+
+    /// Returns a list of all elements of the proof, ordered by key in lexicographic order.
+    ///
+    /// The iterator includes branch nodes.
+    pub fn iter_ordered(
+        &'_ self,
+    ) -> impl Iterator<Item = (&'_ [nibble::Nibble], StorageValue<'_>)> + '_ {
```
tomaka, line 378: After thinking about it, I think that returning `TrieNodeInfo` might be more appropriate.

PR #3037, opened 2022-11-28
Add ProofBuilder::trie_root_hash and ProofBuilder::make_coherent

tomaka:
cc https://github.com/paritytech/smoldot/issues/2973

This allows tweaking the node values inserted into the proof.

This is completely untested, and some testing indicates that something is wrong somewhere. I'll probably push fixes to this PR after having figured it out.


src/trie/proof_encode.rs
```diff
@@ -119,10 +122,6 @@ impl ProofBuilder {
                 None
             }
             (proof_node_codec::StorageValue::Hashed(ref hash), Some(ref value)) => {
-                debug_assert_eq!(
-                    blake2_rfc::blake2b::blake2b(32, b"", value).as_bytes(),
-                    &hash[..]
-                );
```
tomaka, line 125: This check has been removed in order to allow the user to insert a storage value whose hash doesn't match the one in the node value. When that is the case, `make_coherent` will put the correct hash in the node value.

PR #3045, opened 2022-11-28
Publish v0.7.9

tomaka:


tomaka:
`rustup` has just segfaulted on CI :facepalm:

PR #3078, opened 2023-01-03
Use $GITHUB_OUTPUT instead of set-output

### LINKED ISSUE #2885: ::set-output is deprecated
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

We use it in actions.

vivekvpandya:
Fixes #2885


dmitry-markin:
Already included into the changes from upstream.

PR #3090, opened 2023-01-30
Accept JSON-RPC requests without a `params` field

### LINKED ISSUE #3080: Allow requests with omitted `params` field
An issue was reported in cumulus where requests caused smoldot to crash because the `params` field is omitted in the JSON-RPC request https://github.com/paritytech/cumulus/issues/2053 .

After looking at the [JSON-RPC Spec](https://www.jsonrpc.org/specification#request_object) it seems that the `params` field should be optional and Smoldot should accept requests without it.

tomaka:
Fix https://github.com/paritytech/smoldot/issues/3080

### DIFF
--- a/bin/wasm-node/CHANGELOG.md
+++ b/bin/wasm-node/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- JSON-RPC requests without a `params` field are no longer invalid.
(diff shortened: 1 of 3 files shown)


dmitry-markin:
Closing as this PR is already included into the changes from upstream.

PR #3095, opened 2023-02-09
fix random

turuslan:
wasm smoldot node always started with same peer id  
because [`.slice()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice) (clone) and [`.subarray()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/subarray) (mutable reference) were confused  
so randomness was not written back to wasm memory


tomaka:
Nice catch!

Would it be possible for you to reopen the PR in https://github.com/smol-dot/smoldot/? This is where development will happen in the future. :pray:

Issue #3097, opened 2023-02-23
Repo migration?


xlc:
Is this repo going to be deprecated in favour of https://github.com/smol-dot/smoldot?

bkchr:
Yes. We should put some information into the Readme

PR #3098, opened 2023-02-23
Inform users about the new repo

bkchr:
Closes: https://github.com/paritytech/smoldot/issues/3097

Is that okay @tomaka?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
     - uses: baptiste0928/cargo-install@v1  # This action ensures that the twiggy compilation is cached.
       with:
         crate: twiggy
-        version: 0.6
+        version: 0.7
     - run: git checkout ${{ github.event.pull_request.base.sha }}
     - run: cd bin/wasm-node/javascript && npm ci && npm run-script build
     - run: cp ./target/wasm32-wasi/min-size-release/smoldot_light_wasm.wasm ./.ci-parent-build.wasm  # TODO: maybe get the path from the `npm build` output or something?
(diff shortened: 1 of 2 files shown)


tomaka:
Yeah, I had in mind to do a PR like that, just I was thinking of writing a bit more text, but also I don't know exactly what to write

tomaka:
The CI failure can be fixed by putting a `0.7` here: https://github.com/paritytech/smoldot/blob/76994b52cd658a4ac1dc7a08f6bd4eba702e0e87/.github/workflows/ci.yml#L79

I don't know why it suddenly broke, but I had to do the same fix in the new repo.

bkchr:
> Yeah, I had in mind to do a PR like that, just I was thinking of writing a bit more text, but also I don't know exactly what to write

Okay. I think I would move forward and if you have some better text/ideas just open a pr and we extend it! :)

tomaka:
Also the CI fails because the spellcheck doesn't know the word "repo"

tomaka (approved):
My review doesn't count, but I guess I'll put it for the principle

forgetso:
Why did the repo move to a new org?

tomaka:
I'm no longer working at Parity and I'm continuing the development of smoldot through treasury proposals. Since the project doesn't "belong to Parity" "anymore" (the code in 2022 and before belongs to Parity, but everything is under GPL 3 regardless), it makes sense for it to be in a different organization.

The reason why we didn't simply move the repo is complicated and uninteresting.

forgetso:
Thanks for the explanation @tomaka and good luck!

PR #3105, opened 2023-03-16
camelCase transaction status to align with Substrate

jsdw:
Hey! An error cropped up via Ink in trying to use Subxt to communicate with Smoldot:

```
error on call `wait_for_in_block`: Serialization(Error("unknown variant `Ready`, expected one of `future`, `ready`, `broadcast`, `inBlock`, `retracted`, `finalityTimeout`, `finalized`, `usurped`, `dropped`, `invalid`", line: 1, column: 8))', /Users/bruno/.cargo/git/checkouts/ink-1add513eda8f5a89/9c9215f/crates/e2e/src/xts.rs:270:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I believe the reason for the issue was that Smoldot's `author_submitAndWatchExtrinsic` returns transaction statuses whose variants aren't camelCased, whereas the Substrate RPC method does camelCase (see https://github.com/paritytech/substrate/blob/283106b19b85bc8deb232cfa624eed195f143541/client/transaction-pool/api/src/lib.rs#L106)

So I wonder whether you'd be open to applying the same camelCasing in Smoldot to align?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,7 +76,7 @@ jobs:
     - uses: baptiste0928/cargo-install@v1  # This action ensures that the twiggy compilation is cached.
       with:
         crate: twiggy
-        version: 0.6
+        version: 0.7
     - run: git checkout ${{ github.event.pull_request.base.sha }}
     - run: cd bin/wasm-node/javascript && npm ci && npm run-script build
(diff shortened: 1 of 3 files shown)


dmitry-markin:
The fix is already included in changes from upstream.

PR #3127, opened 2026-03-16
statement store: Implement basic statement-store support

### LINKED ISSUE #11285: statement-store: Implement a working prototype of light node for non-production use
Turn the [existing PoC](https://github.com/bkchr/smoldot/tree/bkchr-battleship) into a working prototype for non-production use. Light clients receive all statements from connected full nodes (no filtering yet).
- Define and finalize the external API: declare topic interest, submit statements.
- Ensure the implementation correctly speaks the statement-store wire protocol.
- Implement basic deduplication logic.

AndreiEres:
Summary:
- Based on https://github.com/bkchr/smoldot/tree/bkchr-battleship
- Implements the Statement Store networking protocol for peer-to-peer statement distribution
- Adds JSON-RPC methods for submitting and subscribing to statements

Protocol Implementation:
- Statement encoding/decoding
- Topic-based filtering with three modes: Any, MatchAll, and MatchAny
- Duplicate detection via LRU cache of statement hashes

Network Layer:
- New StatementProtocolConfig for enabling the protocol on a chain
- Statement notification substream handling
- StatementProtocolConnected and StatementNotification events


examples/statement-chat/README.md
```diff
@@ -0,0 +1,44 @@
+# Statement Chat
```
alexggh: I wonder if this repo is the right place for this example.

lib/src/json_rpc/service/client_main_task.rs
```diff
@@ -433,6 +433,7 @@ impl ClientMainTask {
                 | methods::MethodCall::system_properties { .. }
```
alexggh, line 433: Did you also check that the statement-store logic works ok in the full node mode ?

lib/src/network/codec/statement.rs
```diff
@@ -0,0 +1,756 @@
...
+        out.extend_from_slice(data);
+    }
+
+    Ok(())
+}
+
+fn encode_proof_into(proof: &ProofRef<'_>, out: &mut Vec<u8>) {
+    match proof {
+        ProofRef::Sr25519 { signature, signer } => {
+            out.push(0);
```
alexggh: Nit: harcoded tags .

lib/src/network/codec/statement.rs
```diff
@@ -0,0 +1,756 @@
...
+        }
+        _ => Err(nom::Err::Failure(nom::error::make_error(
+            input,
+            nom::error::ErrorKind::Verify,
+        ))),
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
alexggh, line 455: All this tests do round-trip, which means if you change serialization/deserialization accidentally in a way it doesn't match polkadot-sdk, you won't catch the bug.

Can we also hardcode, some bytes of serialized statements from polkadot-sdk, and run some tests against those.

lib/src/network/codec.rs
```diff
@@ -299,7 +313,57 @@ fn protocol_ty_to_real_protocol(
...
+        ProtocolTy::Statement => ProtocolName::Statement {
+            genesis_hash,
+            fork_id,
+        },
     }
 }
 
-// TODO: tests for the protocol names
+#[cfg(test)]
+mod tests {
```
alexggh, line 324: Same here, you also want to check the same protcol names are generated as polkadot-sdk by hardcoding some name. Doing just round-trips doesn't help if you accidentally change the serialization format.

lib/src/network/service.rs
```diff
@@ -261,6 +291,11 @@ struct Chain<TChain> {
     /// See [`ChainConfig::allow_inbound_block_requests`].
     allow_inbound_block_requests: bool,
 
+    /// See [`ChainConfig::statement_protocol_config`].
+    statement_protocol_config: Option<StatementProtocolConfig>,
+
+    seen_statements: Option<lru::LruCache<[u8; 32], (), fnv::FnvBuildHasher>>,
```
alexggh: Nit: Some doc explaining what we use this for.

lib/src/network/service.rs
```diff
@@ -2534,6 +2614,12 @@ where
                         .next()
                         .is_some()
                     {
+                        let connection_id = self
+                            .substreams
+                            .get(&substream_id)
+                            .expect("substream was just inserted above; qed")
```
alexggh:
```suggestion
let connection_id = substream_info.connection_id;
```

lib/src/network/service.rs
```diff
@@ -2745,6 +2892,55 @@ where
...
+                                        });
+                                    }
+                                };
+
+                            let chain = &mut self.chains[chain_index];
+                            let non_duplicate_statement_bytes: Vec<&[u8]> = statement_bytes
+                                .into_iter()
+                                .filter(|statement_bytes| {
+                                    let hash = codec::statement_hash(statement_bytes);
+                                    let Some(cache) = &mut chain.seen_statements else {
```
alexggh: Is this the right place for seen_statements filtering ?

light-base/examples/statement_test.rs
```diff
@@ -0,0 +1,106 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Example demonstrating how to use the statement store protocol with smoldot.
```
alexggh, line 18: What are the depdendencies for this test to actually work ?

light-base/src/json_rpc_service/background.rs
```diff
@@ -704,12 +738,58 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
                 me.block_runtimes_pending.shrink_to_fit();
             }
 
+            WakeUpReason::NetworkStatementReceived(notification_data) => {
+                // If there is no statement subscription, we can skip decoding the notification entirely.
+                if me.statement_subscriptions.is_empty() {
+                    continue;
+                }
+
+                match codec::decode_statement_notification(&notification_data) {
```
alexggh: I wonder if the seen_statements shouldn't exist here, and let the networking layer be just a networking layer that delivers notifications.

light-base/src/json_rpc_service/background.rs
```diff
@@ -704,12 +738,58 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                    Ok(statements) => {
+                        for statement in statements {
+                            // TODO: re-encoding every statement is inneficient
+                            let Ok(encoded) = codec::encode_statement(&statement) else {
+                                continue;
+                            };
+
+                            for (sub_id, topic_filter) in &me.statement_subscriptions {
```
alexggh: TODO: This will be inneficient if number of statement_subscriptions is big. I assume it shouldn't.

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
                             .await;
                     }
 
+                    methods::MethodCall::statement_submit { encoded } => {
+                        let network = me.network_service.clone();
```
alexggh: Nit: Can we refactor this logic in its own function, applies for other match arms we modified.

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            )
+                        );
+
+                        if total_peers == 0 {
+                            if me
+                                .responses_tx
+                                .send(
+                                    methods::Response::statement_submit(
+                                        methods::StatementSubmitResult::Error(
+                                            "No connected peers".to_string(),
```
alexggh: Should we make a proper enum for this errors ?

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                                        "statement_submit: failed to send to {}: {:?}",
+                                        peer, e
+                                    )
+                                ),
+                            }
+                        }
+
+                        let result = if sent_count == 0 {
+                            methods::StatementSubmitResult::Error(
+                                "Failed to send to any peers".to_string(),
```
alexggh: Nit: Could be something like this and then you don't need the extra stuff above
```
let msg = if total_peers == 0 {
                                "No connected peers"
                            } else {
                                "Failed to send to any peers"
                            };
                            methods::StatementSubmitResult::Error(msg.to_string())
```

light-base/src/sync_service/standalone.rs
```diff
@@ -809,6 +809,13 @@ pub(super) async fn start_standalone_chain<TPlat: PlatformRef>(
                 }
             }
 
+            WakeUpReason::NetworkEvent(network_service::Event::StatementNotification {
```
alexggh: Can we avoid delivering them in the first place ? Not sure what is the penalty of continously waking up here for statements.

light-base/src/network_service.rs
```diff
@@ -552,6 +559,26 @@ impl<TPlat: PlatformRef> NetworkServiceChain<TPlat> {
...
+        self.messages_tx
+            .send(ToBackgroundChain::SendStatements {
+                target: target.clone(),
+                notification,
+                result: tx,
+            })
+            .await
+            .expect("background task alive; qed");
+
+        rx.await.expect("background task responded; qed")
```
alexggh: Do this expects make sense here ?

light-base/src/json_rpc_service/background.rs
```diff
@@ -686,6 +702,24 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
                 WakeUpReason::GarbageCollection
             })
+            .or(async {
+                // Poll for network events (incoming statements)
+                let Some(rx) = &me.network_events_rx else {
+                    return future::pending().await;
+                };
+                loop {
+                    let Ok(event) = rx.recv().await else {
+                        return future::pending().await;
```
alexggh: could this hang smoldot ?

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
                             .await;
                     }
 
+                    methods::MethodCall::statement_submit { encoded } => {
+                        let network = me.network_service.clone();
+                        let peers: Vec<_> = network.peers_list().await.collect();
```
alexggh: Should we filter just for peers connected on the statement protocol, otherwise we will print errors for the ones that don't support it.

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
                             .await;
                     }
 
+                    methods::MethodCall::statement_submit { encoded } => {
```
alexggh, line 2834: Is encoded validated anywhere in smoldot ? Are we assuming we got valid statements here ?

lib/src/network/service.rs
```diff
@@ -3889,6 +4105,43 @@ where
         )
     }
 
+    /// Sends a statement notification to the given peer, wrapping the single encoded
+    /// `statement` in the `Vec<Statement>` format expected by the protocol.
+    ///
+    /// Returns [`QueueNotificationError::NoConnection`] if no
+    /// [`Event::GossipConnected`] of kind [`GossipKind::ConsensusTransactions`] has been
```
alexggh: Comment seems wrong.

lib/src/network/codec/statement.rs
```diff
@@ -0,0 +1,756 @@
...
+            nom::error::ErrorKind::NonEmpty,
+        ));
+    }
+
+    Ok(statements)
+}
+
+/// Error when decoding a statement notification.
+#[derive(Debug, derive_more::Display, derive_more::Error, Clone)]
+#[display("Failed to decode statement notification")]
```
alexggh:
```suggestion
#[display("Failed to decode statement notification {:?}")]
```

lib/src/network/codec/statement.rs
```diff
@@ -0,0 +1,756 @@
...
+                "Too many topics for MatchAny: got {}, max {}",
+                topics.len(),
+                MAX_ANY_TOPICS
+            ));
+        }
+        Ok(TopicFilter::MatchAny(topics))
+    }
+
+    /// Returns `true` if the given statement topics match this filter.
+    pub fn matches(&self, statement_topics: &[&Topic]) -> bool {
```
alexggh: Tests.

lib/src/network/codec/statement.rs
```diff
@@ -0,0 +1,756 @@
...
+    let (input, variant) = nom::number::streaming::u8(input)?;
+    match variant {
+        0 => {
+            // Sr25519
+            let (input, signature) = nom::bytes::streaming::take(64u32)(input)?;
+            let (input, signer) = nom::bytes::streaming::take(32u32)(input)?;
+            Ok((
+                input,
+                ProofRef::Sr25519 {
+                    signature: <&[u8; 64]>::try_from(signature).unwrap(),
```
alexggh: Nit: QED's for this unwraps

alexggh (commented):
Had a pass, things are looking good to me. I put some comments, but nothing major.

 One thing it is not clear to me is how do we want to test E2E, but that could be answered in followup PRs and issues.

lib/Cargo.toml
```diff
@@ -57,6 +57,7 @@ itertools = { version = "0.14.0", default-features = false }
 libm = { version = "0.2.8", default-features = false }
 libsecp256k1 = { version = "0.7.1", default-features = false, features = ["static-context", "hmac"] }
 # The log` crate is forbidden, as it is very impolite to emit logs from a library.
+lru = { version = "0.16.3", default-features = false, features = ["hashbrown"] }
```
dmitry-markin: Should not be needed in `lib/`.

lib/src/json_rpc/methods.rs
```diff
@@ -462,6 +462,13 @@ define_methods! {
     /// Returns, as an opaque string, the version of the client serving these JSON-RPC requests.
     system_version() -> Cow<'a, str>,
 
+    /// Submit a new statement to the store and broadcast to peers.
+    statement_submit(encoded: HexString) -> StatementSubmitResult,
+    /// Subscribe to statements matching the given filter. Returns subscription ID.
+    statement_subscribe(filter: crate::network::codec::TopicFilter) -> Cow<'a, str>,
+    /// Unsubscribe from statement notifications.
+    statement_unsubscribe(subscription: String) -> bool,
```
dmitry-markin: `statement_*` methods seem to be missing from https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/SUMMARY.md

They should go as `statement_unstable_*` unless we stabilize the API in `json-rpc-interface-spec` first.

lib/src/network/service.rs
dmitry-markin: Agree with another @alexggh comment, this should go into `light-base` and `lib` must be mostly stateless network message forwarding.

lib/src/network/service.rs
```diff
@@ -2549,6 +2635,66 @@ where
...
+                                },
+                            );
+                            self.notification_substreams_by_peer_id.insert((
+                                substream_protocol,
+                                peer_index,
+                                SubstreamDirection::Out,
+                                NotificationsSubstreamState::Pending,
+                                new_substream_id,
+                            ));
+                        }
```
dmitry-markin: There should be no need for this entire code block. An outbound statements protocol should be opened as soon as gossip link is established (block announcements protocol opened). Just need to add statement protocol to https://github.com/paritytech/smoldot/blob/d361c495d1935c4bcc949077cf75bb0d80eea120/lib/src/network/service.rs#L1812

The reconnects are also already handled automatically.

lib/src/network/service.rs
dmitry-markin: +1 for doing it in `light-base/src/network_service.rs`

lib/src/network/service.rs
```diff
@@ -3889,6 +4105,43 @@ where
...
+    /// Panics if the [`ChainId`] is invalid.
+    ///
+    pub fn gossip_send_statements(
+        &mut self,
+        target: &PeerId,
+        chain_id: ChainId,
+        statement: Vec<u8>,
+    ) -> Result<(), QueueNotificationError> {
+        let mut notification = Vec::with_capacity(1 + statement.len());
+        notification.extend_from_slice(util::encode_scale_compact_usize(1).as_ref());
```
dmitry-markin, line 4001: Do we prepend a payload with the number of statements on the wire? If this means "we are sending one statement", then the function name should be `gossip_send_statament`?

lib/src/network/service.rs
```diff
@@ -4377,6 +4641,27 @@ pub enum Event<TConn> {
...
+        statements: EncodedStatementNotification,
+    },
+
+    /// An inbound statement protocol substream has been accepted.
+    StatementProtocolConnected {
+        /// Identity of the peer.
+        peer_id: PeerId,
+        /// Index of the chain.
+        chain_id: ChainId,
+    },
```
dmitry-markin: nit: Not used, better remove this event from here and `network_service.rs`.

lib/src/network/service.rs
```diff
@@ -2549,6 +2635,66 @@ where
...
+                        }
+
+                        if let NotificationsProtocol::Statement { chain_index } = substream_protocol
+                        {
+                            return Some(Event::StatementProtocolConnected {
+                                peer_id: self.peers[peer_index.0].clone(),
+                                chain_id: ChainId(chain_index),
+                            });
+                        }
+
```
dmitry-markin:
```suggestion
```
nit: remove with the event.

light-base/src/network_service.rs
dmitry-markin: I would keep them as `.unwrap()`, which is a convention in smoldot. `qed` here is not a `qed`.

light-base/src/network_service.rs
```diff
@@ -1714,6 +1751,27 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                        .gossip_send_statements(&target, chain_id, notification);
+                if result.send(send_result).is_err() {
+                    log!(
+                        &task.platform,
+                        Debug,
+                        "network",
+                        "send-statements-result-dropped",
+                        peer_id = &target
+                    );
+                }
```
dmitry-markin: nit: no need to track one-shot sending result

light-base/src/network_service.rs
```diff
@@ -2611,6 +2669,21 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                debug_assert!(task.event_pending_send.is_none());
+                task.event_pending_send = Some((
+                    chain_id,
+                    Event::StatementNotification {
+                        peer_id,
+                        statements,
+                    },
+                ));
+            }
+            WakeUpReason::NetworkEvent(service::Event::StatementProtocolConnected { .. }) => {}
```
dmitry-markin:
```suggestion
```

light-base/src/json_rpc_service/background.rs
```diff
@@ -562,9 +569,17 @@ pub(super) async fn run<TPlat: PlatformRef>(
         ),
         genesis_block_hash: config.genesis_block_hash,
         printed_legacy_json_rpc_warning: false,
+        statement_subscriptions: hashbrown::HashMap::with_capacity_and_hasher(
+            2,
```
dmitry-markin: Is 2 enough? We just need two subscriptions?

light-base/src/json_rpc_service/background.rs
```diff
@@ -562,9 +569,17 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+            Default::default(),
+        ),
+        network_events_rx: None,
         platform: config.platform,
     };
 
+    // Subscribe to network events for receiving statements
+    me.network_events_rx = Some(me.network_service.subscribe().await);
```
dmitry-markin: This should be done lazily and resubscribe like it's done in other services. See for example https://github.com/paritytech/smoldot/blob/d361c495d1935c4bcc949077cf75bb0d80eea120/light-base/src/sync_service/standalone.rs#L158

light-base/src/json_rpc_service/background.rs
dmitry-markin: No, this is just one future among others being selected. Essentially means "return no result". But we must resubscribe on `Err`, see the comment above.

light-base/src/json_rpc_service/background.rs
```diff
@@ -704,12 +738,58 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
             }
 
+            WakeUpReason::NetworkStatementReceived(notification_data) => {
+                // If there is no statement subscription, we can skip decoding the notification entirely.
+                if me.statement_subscriptions.is_empty() {
+                    continue;
+                }
+
+                match codec::decode_statement_notification(&notification_data) {
+                    Ok(statements) => {
```
dmitry-markin: It should be already decoded by this point / no `Err` arm & logging needed. We can add `EncodedStatementNotification::decode` like with other types that always succeeds because `EncodedStatementNotifications` is guaranteed to contain valid data.

light-base/src/json_rpc_service/background.rs
dmitry-markin: It's better to move the logic of statement broadcasting to `network_service`. Then we won't need to await `peers_list` and loop over sending to every peer here. There should be minimal `await`ing of other services in the JSON-RPC event loop.

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            .await
+                            .is_err()
+                        {
+                            log!(
+                                &me.platform,
+                                Debug,
+                                &me.log_target,
+                                "Failed to send response for statement_subscribe: response channel closed"
+                            );
+                        }
```
dmitry-markin: Is this intended to be a dummy code?

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2834,150 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            .await
+                            .is_err()
+                        {
+                            log!(
+                                &me.platform,
+                                Debug,
+                                &me.log_target,
+                                "Failed to send response for statement_unsubscribe: response channel closed"
+                            );
+                        }
```
dmitry-markin: Also intended to be a dummy code?

light-base/src/network_service.rs
```diff
@@ -552,6 +559,26 @@ impl<TPlat: PlatformRef> NetworkServiceChain<TPlat> {
         rx.await.unwrap()
     }
 
+    /// See [`service::ChainNetwork::gossip_send_statements`].
+    pub async fn send_statements(
+        self: Arc<Self>,
+        target: &PeerId,
+        notification: Vec<u8>,
+    ) -> Result<(), QueueNotificationError> {
```
dmitry-markin:
```suggestion
    pub async fn broadcast_statements(
        self: Arc<Self>,
        notification: Vec<u8>,
    ) -> Result<(), QueueNotificationError> {
```

dmitry-markin (commented):
Left some nits, but overall the approach looks good!

lib/src/json_rpc/methods.rs
```diff
@@ -540,6 +547,9 @@ define_methods! {
     // This function is a custom addition in smoldot. As of the writing of this comment, there is
     // no plan to standardize it. See https://github.com/paritytech/smoldot/issues/2245.
     sudo_networkState_event(subscription: Cow<'a, str>, result: NetworkEvent) -> (),
+
+    // Statement notification sent when a statement matching subscribed topics is received.
+    statement_notification(subscription: Cow<'a, str>, statement: HexString) -> (),
```
alexggh: Does it make sense to send a Vec of statements here ? Since we are already receiving a Vec of statements on the P2P layer.

light-base/src/json_rpc_service/background.rs
AndreiEres: Why you think it's dummy?

examples/statement-chat/README.md
AndreiEres: It's ok for now as it is in heavy development stage, afterwards we can move it somewhere

lib/src/json_rpc/service/client_main_task.rs
AndreiEres: I didn't check and had no plans to check it, honestly

lib/src/json_rpc/methods.rs
AndreiEres: done

lib/src/network/codec/statement.rs
AndreiEres: addressed

lib/src/network/codec/statement.rs
AndreiEres: Addressed

lib/src/network/codec/statement.rs
AndreiEres: Added tests

lib/src/network/codec.rs
AndreiEres: done

lib/src/network/service.rs
AndreiEres: added doc and moved to light-base

lib/src/network/service.rs
AndreiEres: not needed

lib/src/network/service.rs
AndreiEres: moved

light-base/src/json_rpc_service/background.rs
AndreiEres: moved

light-base/src/json_rpc_service/background.rs
AndreiEres: I suppose done

light-base/src/json_rpc_service/background.rs
AndreiEres: Not needed anymore

light-base/src/network_service.rs
AndreiEres: unwrapped

light-base/src/json_rpc_service/background.rs
AndreiEres: added

light-base/src/json_rpc_service/background.rs
AndreiEres: moved to network_service

light-base/src/json_rpc_service/background.rs
AndreiEres: We assume we got a valid statement here and let other nodes to do our validation job

lib/src/network/service.rs
AndreiEres: removed

light-base/src/json_rpc_service/background.rs
AndreiEres: done

light-base/src/json_rpc_service/background.rs
AndreiEres: should be addressed

lib/src/json_rpc/methods.rs
AndreiEres: wdyt if we address it in the following PR?

light-base/src/json_rpc_service/background.rs
AndreiEres: It's just initial HashMap capacity, not a limit. Grows automatically. 2 is a reasonable default.

light-base/src/json_rpc_service/background.rs
AndreiEres: Let's do it in a following PR?

light-base/src/json_rpc_service/background.rs
AndreiEres: same: why it's dummy

light-base/src/json_rpc_service/background.rs
AndreiEres: WDYT if we open an issue about it to not forget?

light-base/src/sync_service/standalone.rs
AndreiEres: Probably we don't have it anymore

AndreiEres:
@alexggh @dmitry-markin 
Thank you for the thorough review. 

There were many comments, so to avoid getting lost in them, I resolved the fixed or outdated ones and left some with my answers. I would greatly appreciate it if you could take another look.

lib/src/json_rpc/methods.rs
alexggh: Actually, thinking more about this, this RPCs definitions need to be JSON compatible with the ones from polkadot-sdk, otherwise we need to create bespoke implementations when switching between lightnode or full node rpc.

lib/src/json_rpc/methods.rs
```diff
@@ -540,6 +547,9 @@ define_methods! {
     // This function is a custom addition in smoldot. As of the writing of this comment, there is
     // no plan to standardize it. See https://github.com/paritytech/smoldot/issues/2245.
     sudo_networkState_event(subscription: Cow<'a, str>, result: NetworkEvent) -> (),
+
+    // Statement notification sent when statements matching subscribed topics are received.
+    statement_subscribeStatement(subscription: Cow<'a, str>, result: StatementEvent) -> (),
```
alexggh: Doesn't seem right.

bkchr (commented):
There is an architectural problem with the statement store in smoldot. As smoldot is not storing the seen messages and throws them away, it can happen that we never report the statements to the application logic. The problem arises in the time frame between smoldot connecting to peers and the application logic starting to subscribe to statements. In this time frame smoldot will throw away all the statements and as the full nodes start sending all known statements to smoldot after the connection is created, statements very likely get lost.
We probably need to split the statement store protocol into a sending and receiving one. So, that both can connect individually from each other (maybe the p2p layer already supports this). Then, we only connect these sub streams when we are about to either subscribe/send statements.

For now, the best we can achieve is to keep the protocol closed until there is either a subscription or a send request. And then we should document this and tell users to first subscribe before trying to send statements.

AndreiEres:
> The problem arises in the time frame between smoldot connecting to peers and the application logic starting to subscribe to statements. 

We solved this in the next step: adding explicit affinity (read bloom filters). Peers don't send statements if there is no subscription. When a client subscribes to a new topic, light node changes its bloom filter and sends it to the peers. Peers send all missing statements to the light node.

See https://github.com/paritytech/polkadot-sdk/pull/11329, https://github.com/paritytech/smoldot/pull/3128

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2828,97 @@ pub(super) async fn run<TPlat: PlatformRef>(
                             .await;
                     }
 
+                    methods::MethodCall::statement_submit { encoded } => {
```
alexggh, line 2834: I wonder if we need somehow to give errors if we try to subscribe but we have 0 connections on the statement protocol.

examples/statement-chat/README.md
```diff
@@ -0,0 +1,44 @@
...
+3. Enter a topic (32-byte hex) or use the default
+4. Click "Subscribe"
+5. Send messages - they'll be distributed to all peers on the same topic
+
+## How It Works
+
+The app uses smoldot's statement store protocol:
+- `statement_subscribeStatement` - Subscribe to topics on a chain
+- `statement_submit` - Broadcast signed statements to the network
+- `statement_subscribeStatement` (notification) - Receive statements from other peers
```
alexggh: needs updating.

alexggh (approved):
Looks good to me.

I would caveat with codebase is new to me, so take it with a grain of salt :D.

light-base/src/json_rpc_service/background.rs
dmitry-markin: The smoldot approach is to init data structures with enough capacity to not reallocate. I wonder if there is such capacity for subscriptions?

light-base/src/json_rpc_service/background.rs
dmitry-markin: Yeah, my bad. Missed the core `me.statement_subscriptions.insert()` completely.

light-base/src/json_rpc_service/background.rs
dmitry-markin: Also not dummy of course, sorry.

examples/statement-chat/README.md
dmitry-markin: Do you plan to move it to a dedicated repo before merging this PR?

light-base/src/json_rpc_service/background.rs
dmitry-markin: +1 for issue and a TODO comment here.

examples/statement-chat/.gitignore
dmitry-markin, line 1: It's better not to put webapps in the smoldot repo — we will end up putting random unrelated things into it. @AndreiEres can you move `statement-chat` to a dedicated repo, please, may be a personal one?

lib/src/json_rpc/methods.rs
```diff
@@ -462,6 +462,13 @@ define_methods! {
     /// Returns, as an opaque string, the version of the client serving these JSON-RPC requests.
     system_version() -> Cow<'a, str>,
 
+    /// Broadcast a new statement to peers (light node has no local statement-store).
+    statement_submit(encoded: HexString) -> StatementSubmitResult,
+    /// Subscribe to statements matching the given filter. Returns subscription ID.
+    statement_subscribeStatement(filter: crate::network::codec::TopicFilter) -> Cow<'a, str>,
```
dmitry-markin: `crate::network::codec::TopicFilter` is not used by the network layer. Let's move it into `lib/src/json_rpc/methods.rs` to `StatementEvent` and co.

lib/src/network/codec/statement.rs
```diff
@@ -0,0 +1,800 @@
...
+        assert!(!filter.matches(&[]));
+    }
+
+    #[test]
+    fn decode_encoded_statement() {
+        // See "statement_encoding_matches_vec" in polkadot-sdk for the original statement fields
+        let bytes = hex::decode(
+            "1c00032a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a2a1818181818181818181818181818181818181818181818181818181818181818420000000000000001dededededededededededededededededededededededededededededededede02e70300000000000003cccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc04010101010101010101010101010101010101010101010101010101010101010105020202020202020202020202020202020202020202020202020202020202020208083763",
+        )
+        .unwrap();
```
dmitry-markin, line 595: Let's break this with `\` into lines under 100 characters.

lib/src/network/service.rs
```diff
@@ -261,6 +291,9 @@ struct Chain<TChain> {
     /// See [`ChainConfig::allow_inbound_block_requests`].
     allow_inbound_block_requests: bool,
 
+    /// See [`ChainConfig::statement_protocol_config`].
+    statement_protocol_config: Option<StatementProtocolConfig>,
```
dmitry-markin: Looks like `enable_statement_protocol: bool` should be enough here, and `StatementProtocolConfig` can go to `light-base`?

lib/src/network/service.rs
```diff
@@ -2549,6 +2630,7 @@ where
                             self.notifications_protocol_handshake(substream_protocol),
                             self.notifications_protocol_max_notification_size(substream_protocol),
                         );
+
```
dmitry-markin:
```suggestion
```

examples/statement-chat/README.md
AndreiEres: Will do

light-base/src/network_service.rs
```diff
@@ -2611,6 +2717,39 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                        cache.push(hash, ());
+                        Some(statement)
+                    })
+                    .collect();
+
+                if non_duplicates.is_empty() {
+                    continue;
+                }
+
+                debug_assert!(task.event_pending_send.is_none());
```
dmitry-markin: nit: would move this to the beginning of the match arm, so it's visible the arm doesn't execute at all if its `Some`.

light-base/src/json_rpc_service/background.rs
```diff
@@ -686,6 +700,21 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                loop {
+                    let Ok(event) = rx.recv().await else {
+                        me.network_events_rx = None;
+                        return WakeUpReason::MustSubscribeNetworkEvents;
+                    };
+                    if let network_service::Event::StatementsNotification { statements, .. } = event
+                    {
+                        return WakeUpReason::NetworkStatementsReceived(statements);
+                    }
+                }
```
dmitry-markin, line 716: We might want to use a dedicated statement notifications channel in the future instead of extending network events. The benefits:
1. Won't need to receive and discard network events in the json-rpc service.
2. If at some point statements become large or their amount becomes large, there is no point in dispatching them to all network subscribers.

At least worth estimating the overhead as a follow-up and may be decoupling them. @AndreiEres can you create an issue for this, please?

light-base/src/json_rpc_service/background.rs
```diff
@@ -704,12 +733,57 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
                 me.transactions_subscriptions.shrink_to_fit();
+                me.statement_subscriptions.shrink_to_fit();
                 me.legacy_api_stale_storage_subscriptions.shrink_to_fit();
                 me.multistage_requests_to_advance.shrink_to_fit();
                 me.block_headers_pending.shrink_to_fit();
                 me.block_runtimes_pending.shrink_to_fit();
             }
 
+            WakeUpReason::MustSubscribeNetworkEvents => {
+                me.network_events_rx = Some(me.network_service.subscribe().await);
```
dmitry-markin, line 745:
```suggestion
                debug_assert!(me.network_events_rx.is_none());
                me.network_events_rx = Some(me.network_service.subscribe().await);
```

light-base/src/json_rpc_service/background.rs
```diff
@@ -704,12 +733,57 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+            }
+
+            WakeUpReason::NetworkStatementsReceived(statements) => {
+                if me.statement_subscriptions.is_empty() {
+                    continue;
+                }
+
+                for (sub_id, topic_filter) in &me.statement_subscriptions {
```
dmitry-markin, line 755: We need to create a reverse index `topic` -> `subscription` for adequate complexity. Even ruling out topics we are not interested in at all (none of the filters have them) can speedup things dramatically, and upon match we can do a more rigorous topic/statement match inspection. There are likely even better ways out there.

Currently it's `n_statements` x `n_filters` x `n_topics_in_filter` x `n_topics_in_statement` ...

We can merge for now, but should definitely address this in a follow up. Synchronous processing here can also stall the processing of other RPC methods..

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2828,97 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            .await
+                            .is_err()
+                        {
+                            log!(
+                                &me.platform,
+                                Debug,
+                                &me.log_target,
+                                "Failed to send response for statement_submit: response channel closed"
+                            );
+                        }
```
dmitry-markin: None of other senders handles channel closed errors, this is abnormal situation. Let's not introduce more code)

```suggestion
                        let _ =  me
                            .responses_tx
                            .send(
                                methods::Response::statement_submit(result)
                                    .to_json_response(request_id_json),
                            )
                            .await;
```

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2828,97 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            .await
+                            .is_err()
+                        {
+                            log!(
+                                &me.platform,
+                                Debug,
+                                &me.log_target,
+                                "Failed to send response for statement_subscribeStatement: response channel closed"
+                            );
+                        }
```
dmitry-markin:
```suggestion
                        let _ = me
                            .responses_tx
                            .send(
                                methods::Response::statement_subscribeStatement(Cow::Owned(
                                    subscription_id,
                                ))
                                .to_json_response(request_id_json),
                            )
                            .await;
```

light-base/src/json_rpc_service/background.rs
```diff
@@ -2751,6 +2828,97 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            .await
+                            .is_err()
+                        {
+                            log!(
+                                &me.platform,
+                                Debug,
+                                &me.log_target,
+                                "Failed to send response for statement_unsubscribeStatement: response channel closed"
+                            );
+                        }
```
dmitry-markin:
```suggestion
                        let _ = me
                            .responses_tx
                            .send(
                                methods::Response::statement_unsubscribeStatement(existed)
                                    .to_json_response(request_id_json),
                            )
                            .await;
```

dmitry-markin (commented):
Everything looks good, but let's move the `statement-chat` folder to a dedicated repo before merging the PR.

lib/src/network/service.rs
AndreiEres: but StatementProtocolConfig includes options for number of max_seen_statements

dmitry-markin (approved):

light-base/src/json_rpc_service/background.rs
AndreiEres: will do

light-base/src/json_rpc_service/background.rs
AndreiEres: I'll adjust

light-base/src/json_rpc_service/background.rs
AndreiEres: Let's discuss it later

examples/statement-chat/README.md
AndreiEres: Will do

examples/statement-chat/.gitignore
AndreiEres: I'll move it to a separate PR with [DNM] now

lib/src/json_rpc/methods.rs
AndreiEres: will do

lib/src/network/codec/statement.rs
AndreiEres: Will do

lib/src/network/service.rs
AndreiEres: Got it

light-base/src/network_service.rs
AndreiEres: Will do

light-base/src/json_rpc_service/background.rs
AndreiEres: Will do

light-base/src/json_rpc_service/background.rs
AndreiEres: will do

light-base/src/json_rpc_service/background.rs
```diff
@@ -750,6 +750,7 @@ pub(super) async fn run<TPlat: PlatformRef>(
                     continue;
                 }
 
+                // TODO: not efficient on a big number of subscriptions
```
dmitry-markin: nit: a brief description of the current complexity would be nice (you can copy from the original comment).

dmitry-markin (approved):
Thanks for the solid PR! The implementation is clean and well-scoped. Ready for merging.

bkchr:
@AndreiEres let's get this merged if you are happy.

PR #3128, opened 2026-03-16
[WIP] statement-store: Add Statement Protocol V2

AndreiEres:


lib/src/network/codec/affinity.rs
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    fn match_all(seed: u128) -> Self {
+        let bits = alloc::vec![u64::MAX; 16];
+        let bloom = fastbloom::BloomFilter::from_vec(bits).seed(&seed).hashes(1);
+        AffinityFilter { bloom, seed }
+    }
+
+    fn topic_hash(&self, topic: &[u8; 32]) -> u64 {
+        let mut hasher = BloomDefaultHasher::seeded(&self.seed.to_be_bytes()).build_hasher();
```
alexggh, line 164:
```suggestion
        let mut hasher = BloomDefaultHasher::seeded(&self.seed.to_le_bytes()).build_hasher();
```

Changed this in https://github.com/paritytech/polkadot-sdk/pull/11329

light-base/src/json_rpc_service/background.rs
```diff
@@ -2873,26 +2978,68 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
                             .insert(subscription_id.clone(), filter);
 
-                        let _ = me
+                        let combined_filter =
+                            build_combined_affinity_filter(&me.statement_subscriptions);
+
+                        for peer_id in me.v2_statement_peers.iter().cloned() {
+                            let network = me.network_service.clone();
+                            let filter = combined_filter.clone();
+                            me.background_tasks.push(Box::pin(async move {
```
alexggh, line 2987: Is there a way we can make sure we don't send Topic affinity changes for every statement subscribe, so that if at startup an app starts creating 20 subscriptions we don't send 20 topic affinity changes, but just one.

Something like send explicit affinity updates ever second if something changed.

AndreiEres:
Replaced by https://github.com/paritytech/smoldot/pull/3151

PR #3143, opened 2026-03-27
[DNM] Add statement-store example

AndreiEres:


bkchr:
I don't really see a need to have this as part of the repo. Can be its own repo for example.

AndreiEres:
> I don't really see a need to have this as part of the repo. Can be its own repo for example.

You're absolutely right! (TM). No plans to merge it. I'm going to remove it soon

PR #3148, opened 2026-03-31
Bitswap protocol support & `bitswap_v1_get` RPC

dmitry-markin:
Implement Bitswap protocol & add `bitswap_v1_get` RPC method according to the [spec](https://github.com/paritytech/json-rpc-interface-spec/blob/06e3baefb9aca9361fbdf10127c3fcd2498dcb14/src/api/bitswap_v1_get.md).

### Follow-ups
1. Source Bitswap peers from the DHT instead of reusing gossip connections.
2. Only send Bitswap messages to the peers of the chain corresponding to the RPC endpoint the request has come through.


lib/src/libp2p/cid.rs
```diff
@@ -0,0 +1,476 @@
...
+    /// Return the multihash algorithm type used in this prefix.
+    pub fn multihash_type(&self) -> MultihashType {
+        // Prefix was validated at construction time, so this cannot fail.
+        decode_cid_prefix(&self.0)
+            .expect("CidPrefix is always valid; qed")
+            .mh_type
+    }
+
+    /// Build full CID from this prefix and hash digest.
+    pub fn with_digest(self, digest: &[u8; 32]) -> Cid {
```
lrubasze, line 134: Just thinking if it is ok to assume that digest is always 32 bytes.

lib/src/libp2p/collection/multi_stream.rs
```diff
@@ -318,6 +318,56 @@ where
...
                         })
                     }
+                    Some(established::Event::BitswapInOpen { id }) => {
+                        Some(ConnectionToCoordinatorInner::BitswapInOpen { id })
+                    }
+                    Some(established::Event::BitswapIn { id, message }) => {
+                        Some(ConnectionToCoordinatorInner::BitswapIn { id, message })
+                    }
+                    Some(established::Event::BitswapInClose { id, outcome }) => {
+                        // TODO: Notifications protocol acknowledges close here. Might be not
```
lexnv: dq: The suggestion here is to not propagate this further?

dmitry-markin: Not exactly, it is about adding the substream ID to the list of closed ones, so we can silently ignore subsequent accept/reject commands for this substream. This was probably needed to resolve some races with the Notifications protocol. Not sure if we need this for Bitswap. Will need to revisit this.

lib/src/network/service.rs
```diff
@@ -983,6 +1054,150 @@ where
...
+                    BitswapSubstreamState::Open,
+                    SubstreamId::MIN,
+                )
+                    ..=(
+                        peer_index,
+                        SubstreamDirection::Out,
+                        BitswapSubstreamState::Open,
+                        SubstreamId::MAX,
+                    ),
+            )
```
lrubasze, line 1089: How about adding a helper to reduce boilerplate here (and in other places):
```rust
  fn bitswap_substreams(
      &self,
      peer: PeerIndex,
      direction: SubstreamDirection,
  ) -> impl Iterator<Item = &(PeerIndex, SubstreamDirection, BitswapSubstreamState, SubstreamId)> {
      self.bitswap_substreams_by_peer_id.range(
          (peer, direction, BitswapSubstreamState::MIN, SubstreamId::MIN)
              ..=(peer, direction, BitswapSubstreamState::MAX, SubstreamId::MAX),
      )
  }
```

```suggestion
            .bitswap_substreams(peer_index, SubstreamDirection::Out)
            .filter(|s| s.2 == BitswapSubstreamState::Open)
```

lib/src/network/service.rs
```diff
@@ -983,6 +1054,150 @@ where
...
+                    BitswapSubstreamState::MIN,
+                    SubstreamId::MIN,
+                )
+                    ..=(
+                        peer_index,
+                        SubstreamDirection::Out,
+                        BitswapSubstreamState::MAX,
+                        SubstreamId::MAX,
+                    ),
+            )
```
lrubasze, line 1163: As per earlier comment:

```suggestion
            .bitswap_substreams(peer_index, SubstreamDirection::Out)
```

lib/src/network/service.rs
```diff
@@ -2874,68 +3167,324 @@ where
...
+                                BitswapSubstreamState::MIN,
+                                collection::SubstreamId::MIN,
+                            )
+                                ..=(
+                                    peer_index,
+                                    SubstreamDirection::In,
+                                    BitswapSubstreamState::MAX,
+                                    collection::SubstreamId::MAX,
+                                ),
+                        )
```
lrubasze, line 3210: As per earlier comment:

```suggestion
                        .bitswap_substreams(peer_index, SubstreamDirection::In)
```

lib/src/network/service.rs
```diff
@@ -2874,68 +3167,324 @@ where
...
```
lrubasze, line 3396: As per earlier comment:


```suggestion
                        .bitswap_substreams(peer_index, SubstreamDirection::In)
```

lib/src/network/service.rs
```diff
@@ -983,6 +1054,150 @@ where
...
+                        BitswapSubstreamState::MIN,
+                        SubstreamId::MIN,
+                    )
+                        ..=(
+                            peer_index,
+                            SubstreamDirection::Out,
+                            BitswapSubstreamState::MAX,
+                            SubstreamId::MAX,
+                        ),
+                )
```
lrubasze, line 1120: As per above comment:


```suggestion
                .bitswap_substreams(peer_index, SubstreamDirection::Out)  
```

lib/src/network/service.rs
```diff
@@ -236,6 +240,38 @@ pub struct ChainNetwork<TChain, TConn, TNow> {
...
+
+    /// All the Bitswap protocols, indexed by `PeerId`, direction, and state.
+    /// TODO: as with Notifications protocol, unclear whether PeerId should come before or after
+    /// the state, same for direction/state.
+    bitswap_substreams_by_peer_id: BTreeSet<(
+        PeerIndex,
+        SubstreamDirection,
+        BitswapSubstreamState,
+        collection::SubstreamId,
+    )>,
```
lrubasze, line 274: Was thinking that maybe we could nest these fields in a `BitswapNetwork` (or BitswapSomething) struct for more clarity?

Feel free to ignore it.

lib/src/network/service.rs
```diff
@@ -2874,68 +3167,324 @@ where
...
+                    let substream_info = self
+                        .substreams
+                        .get(&substream_id)
+                        .unwrap_or_else(|| unreachable!());
+                    let peer_index = *self.inner[substream_info.connection_id]
+                        .peer_index
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!());
+
+                    if let Err(err) = codec::decode_bitswap_message(&message) {
```
lrubasze, line 3250: The message is decoded here to validate correctness and later on it is decoded to actually interpret the contents of the message. 
But I believe it's cheap enough to do it twice. But worth adding a comment about this.

light-base/src/network_service.rs
```diff
@@ -2063,10 +2323,34 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                        what_happened,
+                        bitswap_peering_strategy::UnassignSlotAndBan::Banned { had_slot: true },
+                    ) {
+                        log!(
+                            &task.platform,
+                            Debug,
+                            "network",
+                            "bitswap-slot-unassigned",
+                            peer_id,
+                            ?ban_duration,
```
lrubasze, line 2310: Maybe worth adding a reason as in https://github.com/paritytech/smoldot/blob/94c67a2e8982f0c7fea2f04ab48aa410710f050d/light-base/src/network_service.rs#L2590-L2598

lrubasze (commented):
First round of reviews

light-base/src/bitswap_service.rs
```diff
@@ -0,0 +1,682 @@
...
+    fn allocate_request_id(&mut self) -> RequestId {
+        let request_id = RequestId(self.next_request_id_inner);
+        self.next_request_id_inner += 1;
+
+        request_id
+    }
+}
+
+fn bitswap_have_message(cid: &Cid) -> Vec<u8> {
+    build_bitswap_message(iter::once(cid), WantType::Have, true, true)
```
lexnv: Shouldn't this call use `full: false` as last parameter?

The `bitswap_have_message` accepts only a single CID, then `build_bitswap_message` would create a: 

```rust
WantType::Have {
    cid,
    full: true
}
```

IIUC, when the `full: true` from the codec/bitswap documentation it mentions it repalces the wantlist:
- T0: request for cid_0 arives and `WantType::Have { cid_0, full: true }` is broadcasted
- T1: peers have not responded
- T2: request for cid_1 arives and `WantType::Have { cid_1, full: true }` is broadcasted
- T3: when cid_1 arrives, the want list is replaced from cid_0 to cid_1

Then this could actualy cause the cid_0 to timeout? Since its been replaced by cid_1? 

Same for the `bitswap_block_message`

lib/src/network/service.rs
```diff
@@ -2874,68 +3167,324 @@ where
...
+                    // Clean up the local state and notify the API user.
+                    let Some(substream_info) = self.substreams.remove(&substream_id) else {
+                        unreachable!()
+                    };
+                    let peer_index = *self.inner[substream_info.connection_id]
+                        .peer_index
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!());
+
+                    let _was_in = self.bitswap_substreams_by_peer_id.remove(&(
```
lexnv, line 3371: We are removing the substreams based on state `BitswapSubstreamState::Open`, could we end up in a situation where te connection has been dropped just before the state was pending `BitswapSubstreamState::Pending` (ie failure during negotiation)?

Could this leak an entry in the substream and reject the peer indefinetely via further `bitswap_open` calls?

lib/src/libp2p/connection/established/substream.rs
```diff
@@ -1152,6 +1351,14 @@ where
                 NonZero::<usize>::new(queued_pings.len())
                     .map(|num_pings| Event::PingOutError { num_pings })
             }
+            SubstreamInner::BitswapIn { .. } => Some(Event::BitswapInClose {
+                outcome: Err(BitswapInClosedErr::SubstreamReset),
+            }),
+            SubstreamInner::BitswapInClosed => None,
+            SubstreamInner::BitswapOut { .. } => Some(Event::BitswapOutClose {
```
lexnv: When the `negotiation` feild inside the `SubstreamInner::BitswapOut` is some (ie calling `reset` during negotiation) shoudlnt' we emit an `BitswapOutOpenResult { result error }`? 

Trying to understand if this is symmetrical with `SubstreamInner::NotificationsOutHandshakeRecv` and since we don't have that state explicitly, the equivalent is `SubstreamInner::BitswapOut { negotiation: Some(..) }`?

lib/src/libp2p/connection/established/substream.rs
```diff
@@ -1115,6 +1165,155 @@ where
...
+            SubstreamInner::BitswapIn {
+                mut next_message_size,
+            } => {
+                let mut message = None;
+
+                if let Some(sz) = next_message_size {
+                    match read_write.incoming_bytes_take(sz) {
+                        Ok(Some(msg)) => {
+                            message = Some(msg);
+                            next_message_size = None;
```
lexnv, line 1178: For symmetry with `SubstreamInner::NotificationsIn`, shouldn't we wake the waker? 

```rust
match read_write.incoming_bytes_take(sz) {
    Ok(Some(notif)) => {
        read_write.wake_up_asap();
        notification = Some(notif);
        next_notification_size = None;
    }
```

ie, if multiple messages are buffered, the subtream won't be reinvoked until new network data comes in? Increasing the latency for propagating bitswap messages

lib/src/libp2p/connection/established/substream.rs
```diff
@@ -1115,6 +1165,155 @@ where
...
+                        }
+                        Ok(None) => {}
+                        Err(read_write::IncomingBytesTakeError::ReadClosed) => {
```
lexnv, line 1181: Same as above: for symmetry with `SubstreamInner::NotificationsIn`, shouldn't we wake the waker? 

```rust
    Err(read_write::IncomingBytesTakeError::ReadClosed) => {
        read_write.wake_up_asap();
```

lib/src/libp2p/connection/established/substream.rs
```diff
@@ -1115,6 +1165,155 @@ where
...
+                                }),
+                            );
+                        }
+                    }
+                } else {
+                    match read_write.incoming_bytes_take_leb128(MAX_BITSWAP_MESSAGE_SIZE) {
+                        Ok(Some(sz)) => next_message_size = Some(sz),
+                        Ok(None) => {}
+                        Err(err) => {
+                            return (
```
lexnv, line 1197: Nit: same here `read_write.wake_up_asap();`

light-base/src/bitswap_service.rs
```diff
@@ -0,0 +1,682 @@
...
+                    // remove these requests from internal structures.
+                    if let Some(request_ids) = task.requests_by_cid.remove(&cid) {
+                        for request_id in request_ids {
+                            let request = task.requests.remove(&request_id).unwrap();
+                            let _was_in = task
+                                .requests_by_timeout
+                                .remove(&(request.timeout, request_id));
+                            debug_assert!(_was_in);
+
+                            let _ = request.result_tx.send(Ok(data.to_owned()));
```
lrubasze, line 607: Each request matching the CID implicates allocating a new `Vec<u8>`, which could be expensive. But not sure if this is a real concern.

lib/src/libp2p/cid.rs
dmitry-markin: Yes, transaction storage / our Bitswap implementation only supports 256-bit hashes, and if this changes at some point we can update the code.

lib/src/network/service.rs
dmitry-markin: Unfortunately, this `.filter` ruins the whole idea behind a sorted set.

lib/src/network/service.rs
dmitry-markin: Let's keep it the way it's done for other protocols, introducing a function would force the reader to go to that function to understand the logic behind.

light-base/src/bitswap_service.rs
dmitry-markin: Yes, indeed this should be `false` in both cases, thank you.

lib/src/libp2p/connection/established/substream.rs
dmitry-markin: IIRC I puzzled quite a bit about it when writing this and arrived at the conclusion that `wake_up_asap` is not needed. The only way to find out will be to dive into it again. I will try to)

lib/src/network/codec/bitswap.rs
```diff
@@ -0,0 +1,453 @@
...
+    } else {
+        None
+    };
+
+    let payload = parsed
+        .payload
+        .into_iter()
+        .map(|b| Block {
+            prefix: b.prefix.unwrap_or(&[]),
+            data: b.data.unwrap_or(&[]),
```
lrubasze: Is it fine to silently default prefix and data to empty? I believe `decode_cid_prefix` from cid.rs would fail on empty prefix.
There is even a test for this:
https://github.com/paritytech/smoldot/blob/94c67a2e8982f0c7fea2f04ab48aa410710f050d/lib/src/libp2p/cid.rs#L418-L420

lib/src/libp2p/collection.rs
```diff
@@ -1005,6 +1042,166 @@ where
...
+        substream_id: SubstreamId,
+        message: impl Into<Vec<u8>>,
+    ) -> Result<(), QueueBitswapMessageError> {
+        let (connection_id, state) = match self.outgoing_bitswap_substreams.get(&substream_id) {
+            Some(s) => s,
+            None => panic!(),
+        };
+        assert!(matches!(state, SubstreamState::Open));
+
+        // TODO: add some back-pressure system and return a `QueueBitswapMessageError` if full
```
lexnv, line 1196: Lets add an issue for this to tackle it later

light-base/src/network_service.rs
```diff
@@ -2275,6 +2559,78 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+            WakeUpReason::NetworkEvent(service::Event::BitswapMessage { peer_id, message }) => {
+                log!(
+                    &task.platform,
+                    Debug,
+                    "network",
+                    "bitswap-message-received",
+                    peer_id
+                );
+                debug_assert!(task.bitswap_event_pending_send.is_none());
+                task.bitswap_event_pending_send =
```
lexnv, line 2575: If two `BitswapMessage` events arrive before the `task.bitswap_event_pending_send` is consumed for the release builds the first message is overwitten and lost. I would add a log warn to notify about it and maybe an issue to tackle this later?

light-base/src/network_service.rs
```diff
@@ -1073,7 +1205,9 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
             };
             let service_event = async {
                 if let Some(event) = (task.event_pending_send.is_none()
-                    && task.pending_new_subscriptions.is_empty())
+                    && task.bitswap_event_pending_send.is_none()
```
lexnv, line 1195: This would block the full network events if the bitswap subscriber is slow to drain the channel. Should we exclude it from the check?

light-base/src/network_service.rs
```diff
@@ -3023,6 +3392,17 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
                     peer_id,
                 );
             }
+            WakeUpReason::CanOpenBitswap(peer_id) => {
+                task.network.bitswap_open(&peer_id).unwrap();
```
lexnv, line 3345: This panics if the peer disconnects?

light-base/src/network_service.rs
```diff
@@ -2810,6 +3166,19 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
                     service::GossipKind::ConsensusTransactions,
                 );
             }
+            WakeUpReason::CanAssignBitswapSlot(peer_id) => {
+                task.bitswap_peering_strategy.assign_slot(&peer_id).unwrap();
```
lexnv, line 3119: Similarly this panics if the peer disconnects?

light-base/src/network_service.rs
```diff
@@ -2275,6 +2559,78 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                    ?error
+                );
+                let ban_duration = if error.is_protocol_not_available() {
+                    Duration::from_secs(600)
+                } else {
+                    Duration::from_secs(15)
+                };
+                if matches!(
+                    task.bitswap_peering_strategy
+                        .unassign_slot_and_ban(&peer_id, task.platform.now() + ban_duration,),
```
lexnv, line 2551: dq: What happens when the function returns `bitswap_peering_strategy::UnassignSlotAndBan::Banned { had_slot: true }`? We leak the peer inside the bitswap desiredlist? Should we remove the peer regardles of has_slot?

light-base/src/json_rpc_service/background.rs
```diff
@@ -1043,6 +1054,30 @@ pub(super) async fn run<TPlat: PlatformRef>(
                         // renewing itself the next time it generates a notification.
                     }
 
+                    methods::MethodCall::bitswap_v1_get { cid } => {
+                        log!(
+                            &me.platform,
+                            Debug,
+                            &me.log_target,
+                            // TODO: only log `cid` if it validates, do not log raw RPC input.
```
lexnv, line 1098: We could open an issue for this

light-base/src/network_service.rs
```diff
@@ -866,6 +961,15 @@ enum ToBackgroundChain {
         is_best: bool,
         result: oneshot::Sender<Result<(), QueueNotificationError>>,
     },
+    SendBitswapMessage {
+        target: PeerId,
+        message: Vec<u8>,
+        result: oneshot::Sender<Result<(), SendBitswapMessageError>>,
+    },
+    BroadcastBitswapMessage {
+        message: Vec<u8>,
```
lexnv, line 959: I believe this message gets cloned for every peer connected.

Considering this message could be up to a few MiBs, then we'd allocate on every message num_peers * chunk_size.

Could we place this behind an Arc<> for making it cheeply clonable?


```rust
let results = peers
    .iter()
    .map(|peer| {
        (
            peer,
            task.network.bitswap_send_message(peer, message.clone()),
        )
    })
    .collect::<Vec<_>>(); // we must collect first to send all messages
```

light-base/src/bitswap_service.rs
```diff
@@ -0,0 +1,682 @@
...
+use rand::RngCore;
+use rand_chacha::rand_core::SeedableRng as _;
+use smoldot::{
+    json_rpc::parse,
+    libp2p::cid::{self, Cid, CidPrefix},
+    network::codec::{Block, BlockPresence, BlockPresenceType, WantType, build_bitswap_message},
+};
+
+// TODO: how many parallel requests to expect?
+const PARALLEL_REQUESTS: usize = 50; // 100 MiB of 2 MiB chunks.
```
lrubasze, line 78: ATM this is just a capacity hint, requests are not capped. Is this intended?

lib/src/network/codec/bitswap.rs
```diff
@@ -0,0 +1,453 @@
...
+    // Field 1: block (CID)
+    for slice in protobuf::bytes_tag_encode(1, cid) {
+        entry.extend_from_slice(slice.as_ref());
+    }
+
+    // Field 2: priority (int32)
+    for slice in protobuf::uint32_tag_encode(2, priority) {
+        entry.extend_from_slice(slice.as_ref());
+    }
+
```
lexnv, line 197: dq: Skipping `Field 3: cancel` would cause the wantlist to accumulate remote peers always?

light-base/src/bitswap_service.rs
lexnv, line 78: nit: This is 200Mib total by 50 * 4 MiB chunks?

lib/src/network/service.rs
lexnv: For a 4MiB packet with many entries this might not be cheap? Maybe we can pass directly the decode message?

lrubasze (approved):
Implementation is solid and well-thought. I like that it follows patterns from the codebase (though it causes massive code repetitions, eg. searching by `range()`).
I think we could have a better test coverage.
Overall LGTM

lib/src/network/service.rs
lrubasze: Right. So maybe something like this would suffice:

```rust
fn bitswap_substreams(
      &self,
      peer: PeerIndex,
      direction: SubstreamDirection,
      state_min: BitswapSubstreamState,
      state_max: BitswapSubstreamState,
 ) -> impl Iterator<Item = &(PeerIndex, SubstreamDirection, BitswapSubstreamState, SubstreamId)> {
      self.bitswap_substreams_by_peer_id.range(
          (peer, direction, state_min, collection::SubstreamId::MIN)
              ..=(peer, direction, state_max, collection::SubstreamId::MAX),
      )
}
```

But considering the fact that the whole codebase is using such pattern probably it'd be better to do such cleanup separately and cover other repetitive pieces of code as well.

lib/src/network/service.rs
lrubasze: I think it's just a matter of assigning proper name and writing good comment. 
But as per [comment](https://github.com/paritytech/smoldot/pull/3148#discussion_r3085830623) probably better to address it separately.

lrubasze:
Added some tests https://github.com/paritytech/smoldot/pull/3192

light-base/src/network_service.rs
dmitry-markin: Yes, this is exactly the guard against having more than one pending bitswap event. Given we are limited in memory allocation backpressuring the network is the best we can do.

light-base/src/network_service.rs
dmitry-markin: This is not possible due to network backpressuring — see a comment regarding `task.bitswap_event_pending_send.is_none()` check.

light-base/src/network_service.rs
dmitry-markin: `bitswap_open` returning `Err` here would be a logic error.

light-base/src/network_service.rs
dmitry-markin: Similarly, panic here can only happen due to logic error.

light-base/src/network_service.rs
dmitry-markin: You mean `had_slot: false`? The desired list in `network` is synchronized with slots in `bitswap_peering_strategy`, so if it returns `false` it was not in the list.

light-base/src/network_service.rs
dmitry-markin: We only broadcast request messages (i.e. wantlist), and they are small.

lib/src/network/codec/bitswap.rs
dmitry-markin: No, this is needed to cancel previously sent wantlists, so if the remote keeps track of the wantlists of all peers it won't send the presence to the peers not interested in it anymore once it receives the data with the given CID (and now can answer "have"). This is not implemented in substrate — it doesn't keep track of the remote wantlists and only answers given the chunks it has at the moment.

light-base/src/bitswap_service.rs
dmitry-markin: This is 100 MiB = 50 * 2 MiB chunks. Even though the limit is set to 4 MiB, in practice max 2 MiB Bitswap messages are sent.

The constant here is used for the initial capacity of the structs and builds on the assumption that web apps can request max 100 MiB of data in parallel.

light-base/src/bitswap_service.rs
dmitry-markin: This is an edge case of the web app requesting the same data multiple times in a short succession. Not something we can easily optimize either.

light-base/src/network_service.rs
lexnv: It all makes sense now, thanks 🙏

light-base/src/bitswap_service.rs
dmitry-markin: Yes, this is the initial capacity that we ideally want to be not exceeded to not reallocate.

lexnv (approved):
Nicely done here! High quality and well written! 🙏 

Before merging would be good to have another look over:
- Waking the [read write waker](https://github.com/paritytech/smoldot/pull/3148#discussion_r3080847728): if unsure here or investigation touches too many components i would lean towards adding the `read_write.wake_up_asap()` since `SubstreamInner::NotificationsIn` introduces the call as well. The tradeoff of potentially waking up the waker one more time and wasting a few CPU cycles sounds better than potentially stalling / delaying bitswap messages until another network event occurs 
- Could the [canAssignBitswapSlot](https://github.com/paritytech/smoldot/pull/3148#discussion_r3085667631) race with peer disconnections and cause panics?
- Is it possible to poison the [state](https://github.com/paritytech/smoldot/pull/3148#discussion_r3080785699) of a peer by not removing entries based on `BitswapSubstreamState::Pending`? 
- If not too entangled with other components, the [bitswap message](https://github.com/paritytech/smoldot/pull/3148#discussion_r3085728518) could be put behind an Arc to avoid cloning the message. Similarly, [decoding](https://github.com/paritytech/smoldot/pull/3148#discussion_r3085758955) larger payloads could be performed once 
- Opening a few issues for the current todos and for API questions

light-base/src/json_rpc_service/background.rs
dmitry-markin: IMO its minor and not worth the issue on the repo level — the log is behind a `Debug` guard and the RPC requests are coming from the local client. More like "may be sometime if we happen to touch this code" kind of TODO.

lib/src/network/service.rs
dmitry-markin: With the limit of 1024 blocks per response message, parsing here is extremely fast and is practically limited by memory access (we only produce references to the data in the payload). The alternative of allocating 2 MiB and copying a parsed structure there is worse.

lib/src/libp2p/collection/multi_stream.rs
dmitry-markin: > adding the substream ID to the list of closed ones, so we can silently ignore subsequent accept/reject commands for this substream.

Resolved in https://github.com/paritytech/smoldot/pull/3148/commits/5605900634334e6b61995fa076ac1ffe3c581a54.

lib/src/libp2p/connection/established/substream.rs
dmitry-markin: This is good catch, fixed.

lib/src/network/service.rs
dmitry-markin: With the events delivered from collection in order, the first one will be `BitswapOutOpenResult`, which will transition the peer to `Open` on success. So it will be in a correct state by the time `BitswapOutClose` arrives.

lib/src/libp2p/connection/established/substream.rs
dmitry-markin: Resolved in https://github.com/paritytech/smoldot/pull/3148/commits/be2f6f116cbead931a7b541d3777e7cc780947aa.

lib/src/network/codec/bitswap.rs
dmitry-markin: Resolved in https://github.com/paritytech/smoldot/pull/3148/commits/6e60f3f09665631fbb6186f83e4a8c3588d99d04.

dmitry-markin:
Thanks a lot @lrubasze & @lexnv for thorough reviews!

Regarding the read-write waker, it seems the only edge case where it is important is when zero bytes are read (due to an error or zero-length Bitswap message), in other cases the connection task will be woken up automatically. I have added `read_write.wake_up_asap()` though for completeness and symmetry with Notifications.

All other comments should be also addressed/responded by now.

PR #3151, opened 2026-03-31
statement-store: Implement explicit affinity

### LINKED ISSUE #11287: statement-store: Implement Bloom Filters for light nodes
Reduce light node bandwidth via topic-based filtering.
- Light node constructs a bloom filter from its topic set and sends it during the handshake and after any topic updates.
- Full nodes filter outbound statements for peers that have real bloom filters. 
- Light node discards false positives locally.

AndreiEres:
Fixes https://github.com/paritytech/polkadot-sdk/issues/11287

Implements explicit topic affinity for the statement protocol V2 using bloom filters.

  - Bloom filter codec with portable cross-platform hashing (wasm32/native)
  - V2 wire format: statements (0x00) and affinity (0x01) message types
  - Light client sends affinity filter on peer connect and on subscription changes (debounced 1s)
  - False positives discarded locally via TopicFilter::matches
  - falsePositiveRate and bloomSeed configurable per chain

### DIFF
--- a/full-node/src/network_service.rs
+++ b/full-node/src/network_service.rs
@@ -2272,6 +2272,22 @@ async fn background_task(mut inner: Inner) {
                     ),
                 );
             }
+
+            WakeUpReason::NetworkEvent(service::Event::StatementProtocolConnected {
+                peer_id,
+                chain_id,
+                version,
+            }) => {
+                inner.log_callback.log(
+                    LogLevel::Debug,
+                    format!(
+                        "statement-protocol-connected; peer_id={}; chain={}; version={:?}",
+                        peer_id, inner.network[chain_id].log_name, version
+                    ),
+                );
+            }
+
+            WakeUpReason::NetworkEvent(service::Event::StatementTopicAffinityReceived { .. }) => {}
         }
     }
 }
--- a/full-node/src/network_service.rs
+++ b/full-node/src/network_service.rs
@@ -2287,7 +2287,9 @@ async fn background_task(mut inner: Inner) {
                 );
             }
 
-            WakeUpReason::NetworkEvent(service::Event::StatementTopicAffinityReceived { .. }) => {}
+            WakeUpReason::NetworkEvent(service::Event::StatementTopicAffinityReceived {
+                ..
+            }) => {}
         }
     }
 }
--- a/full-node/src/network_service.rs
+++ b/full-node/src/network_service.rs
@@ -2291,6 +2291,12 @@ async fn background_task(mut inner: Inner) {
             WakeUpReason::NetworkEvent(service::Event::StatementTopicAffinityReceived {
                 ..
             }) => {}
+
+            // Bitswap events are not handled by the full node.
+            WakeUpReason::NetworkEvent(service::Event::BitswapConnected { .. })
+            | WakeUpReason::NetworkEvent(service::Event::BitswapOpenFailed { .. })
+            | WakeUpReason::NetworkEvent(service::Event::BitswapMessage { .. })
+            | WakeUpReason::NetworkEvent(service::Event::BitswapDisconnected { .. }) => {}
         }
     }
 }
--- a/full-node/src/network_service.rs
+++ b/full-node/src/network_service.rs
@@ -2287,6 +2287,7 @@ async fn background_task(mut inner: Inner) {
                 );
             }
 
+            // TODO: we don't filter outbound statements yet
             WakeUpReason::NetworkEvent(service::Event::StatementTopicAffinityReceived {
                 ..
             }) => {}
--- a/lib/src/network/codec.rs
+++ b/lib/src/network/codec.rs
@@ -32,6 +32,7 @@ pub enum StatementProtocolVersion {
 // re-exported here.
 
 mod affinity;
+mod bitswap;
 mod block_announces;
 mod block_request;
 mod grandpa;
@@ -43,6 +44,7 @@ mod statement;
 mod storage_call_proof;
 
 pub use self::affinity::{AffinityFilter, DecodeAffinityFilterError};
+pub use self::bitswap::*;
 pub use self::block_announces::*;
 pub use self::block_request::*;
 pub use self::grandpa::*;
@@ -58,6 +60,7 @@ pub use self::storage_call_proof::*;
 pub enum ProtocolName<'a> {
     Identify,
     Ping,
+    Bitswap,
     BlockAnnounces {
         genesis_hash: [u8; 32],
         fork_id: Option<&'a str>,
@@ -118,6 +121,9 @@ pub fn encode_protocol_name(protocol: ProtocolName) -> impl Iterator<Item = impl
     let (genesis_hash, fork_id, base_protocol_name) = match protocol {
         ProtocolName::Identify => return either::Left(iter::once(Cow::Borrowed("/ipfs/id/1.0.0"))),
         ProtocolName::Ping => return either::Left(iter::once(Cow::Borrowed("/ipfs/ping/1.0.0"))),
+        ProtocolName::Bitswap => {
+            return either::Left(iter::once(Cow::Borrowed("/ipfs/bitswap/1.2.0")));
+        }
         ProtocolName::BlockAnnounces {
             genesis_hash,
             fork_id,
@@ -208,6 +214,9 @@ pub fn decode_protocol_name(name: &'_ str) -> Result<ProtocolName<'_>, ()> {
             }),
             nom::combinator::map(nom::bytes::complete::tag("/ipfs/ping/1.0.0"), |_| {
                 ProtocolName::Ping
+            }),
+            nom::combinator::map(nom::bytes::complete::tag("/ipfs/bitswap/1.2.0"), |_| {
+                ProtocolName::Bitswap
             }),
             nom::combinator::map(
                 (
--- a/lib/src/network/codec.rs
+++ b/lib/src/network/codec.rs
@@ -22,9 +22,16 @@
 use alloc::{borrow::Cow, string::String};
 use core::{fmt, iter};
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum StatementProtocolVersion {
+    V1,
+    V2,
+}
+
 // Implementation note: each protocol goes into a different sub-module whose content is
 // re-exported here.
 
+mod affinity;
 mod block_announces;
 mod block_request;
 mod grandpa;
@@ -35,6 +42,7 @@ mod state_request;
 mod statement;
 mod storage_call_proof;
 
+pub use self::affinity::{AffinityFilter, DecodeAffinityFilterError};
 pub use self::block_announces::*;
 pub use self::block_request::*;
 pub use self::grandpa::*;
@@ -85,6 +93,7 @@ pub enum ProtocolName<'a> {
     Statement {
         genesis_hash: [u8; 32],
         fork_id: Option<&'a str>,
+        version: StatementProtocolVersion,
     },
 }
 
@@ -144,7 +153,13 @@ pub fn encode_protocol_name(protocol: ProtocolName) -> impl Iterator<Item = impl
         ProtocolName::Statement {
             genesis_hash,
             fork_id,
+            version: StatementProtocolVersion::V1,
         } => (genesis_hash, fork_id, "statement/1"),
+        ProtocolName::Statement {
+            genesis_hash,
+            fork_id,
+            version: StatementProtocolVersion::V2,
+        } => (genesis_hash, fork_id, "statement/2"),
     };
 
     let genesis_hash = hex::encode(genesis_hash);
@@ -245,7 +260,8 @@ enum ProtocolTy {
     Kad,
     SyncWarp,
     State,
-    Statement,
+    StatementV1,
+    StatementV2,
 }
 
 fn protocol_ty(name: &str) -> nom::IResult<&str, ProtocolTy> {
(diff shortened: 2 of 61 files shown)


lib/src/network/codec/affinity.rs
```diff
@@ -0,0 +1,417 @@
...
+        AffinityFilter { bloom, seed }
+    }
+
+    pub fn from_topics<'a>(
+        topics: impl Iterator<Item = &'a [u8; 32]>,
+        seed: u128,
+        false_positive_rate: f64,
+    ) -> Self {
+        let topics: Vec<&[u8; 32]> = topics.collect();
+        let count = topics.len().max(1);
```
alexggh, line 157: I think we need a minimum count to ensure this works when assuming million of statements present in the statement store. Could be solved in a follow up PR.

wasm-node/javascript/src/internals/local-instance.ts
```diff
@@ -90,7 +90,7 @@ export interface Instance {
      * indicate whether the initialization was actually successful. If `success` is `false`, then
      * the `chainId` becomes unallocated.
      */
-    addChain: (chainSpec: string, databaseContent: string, potentialRelayChains: number[], disableJsonRpc: boolean, jsonRpcMaxPendingRequests: number, jsonRpcMaxSubscriptions: number, statementStoreMaxSeenStatements: number) => void,
+    addChain: (chainSpec: string, databaseContent: string, potentialRelayChains: number[], disableJsonRpc: boolean, jsonRpcMaxPendingRequests: number, jsonRpcMaxSubscriptions: number, statementStoreMaxSeenStatements: number, statementStoreFalsePositiveRate: number) => void,
```
alexggh: Feels weird to have `statementStoreFalsePositiveRate` as  a field, can it actually be a StatementStoreConfig struct ? So we can extended easily later on ?

light-base/src/network_service.rs
alexggh: We talked about removing this to make it possible to handle the case where smoldot receives a subscription that would matche statements it already received, do you plan to do it as a follow up ?

wasm-node/javascript/src/internals/local-instance.ts
AndreiEres: It is in config, this is a binding for FFI call

light-base/src/network_service.rs
AndreiEres: Yes, I want to do it in separate PR, way too much in the current one

AndreiEres: Here: https://github.com/paritytech/smoldot/pull/3163

full-node/src/network_service.rs
```diff
@@ -2272,6 +2272,25 @@ async fn background_task(mut inner: Inner) {
                     ),
                 );
             }
+
+            WakeUpReason::NetworkEvent(service::Event::StatementProtocolConnected {
```
alexggh, line 2276: Do we need a disconnected as well ?

lib/src/network/codec/statement.rs
```diff
@@ -45,6 +45,14 @@ const PROOF_ED25519: u8 = 1;
 const PROOF_SECP256K1_ECDSA: u8 = 2;
 const PROOF_ON_CHAIN: u8 = 3;
 
+pub use super::affinity::AffinityFilter;
+
+#[derive(Debug, Clone)]
+pub enum StatementMessage {
+    Statements(Vec<([u8; 32], Statement)>),
+    ExplicitTopicAffinity(AffinityFilter),
```
alexggh, line 54: I wonder if SSS light node needs to ever support his.

lib/src/network/codec.rs
```diff
@@ -22,9 +22,16 @@
 use alloc::{borrow::Cow, string::String};
 use core::{fmt, iter};
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum StatementProtocolVersion {
+    V1,
```
alexggh, line 27: Have you checked somehow with this changes V1 still works ?

light-base/src/json_rpc_service/background.rs
```diff
@@ -48,6 +49,8 @@ use smoldot::{
     trie::{minimize_proof, proof_decode},
 };
 
+const STATEMENT_AFFINITY_UPDATE_INTERVAL: Duration = Duration::from_secs(1);
```
alexggh: Nit: Part of the config ? Maybe 1 is not the right value ?

alexggh (approved):

lib/src/network/codec/affinity.rs
```diff
@@ -0,0 +1,417 @@
...
+    /// No topics always match (from broadcast statements).
+    pub fn matches_statement(&self, topics: &[&Topic]) -> bool {
+        if topics.is_empty() {
+            return true;
+        }
+        topics.iter().any(|t| self.bloom.contains(t))
+    }
+
+    pub fn encode_to_vec(&self) -> Vec<u8> {
+        debug_assert!((1..=MAX_NUM_HASHES).contains(&self.bloom.num_hashes()));
```
dmitry-markin, line 207: nit: Shouldn't this be moved to the construction of the filter, possibly returning `Err` from `new`?

Feel free to ignore if this limit is not practically reachable.

lib/src/network/codec/statement.rs
```diff
@@ -197,6 +205,65 @@ pub enum EncodeStatementError {
...
+pub fn encode_statements_message(statements: &[&[u8]]) -> Vec<u8> {
+    let tag_len = 1;
+    let max_compact_len = 5;
+    let total_len: usize = statements.iter().map(|s| s.len()).sum();
+    let mut out = Vec::with_capacity(tag_len + max_compact_len + total_len);
+    out.push(V2_TAG_STATEMENTS);
+    out.extend_from_slice(crate::util::encode_scale_compact_usize(statements.len()).as_ref());
+    for stmt in statements {
+        out.extend_from_slice(stmt);
+    }
```
dmitry-markin, line 243: May be I am missing something, but how the individual statement boundaries are respected here? I.e. should we also encode the individual statement len?

wasm-node/javascript/src/internals/client.ts
```diff
@@ -511,6 +511,7 @@ export function start(options: ClientOptions, wasmModule: SmoldotBytecode | Prom
 
             // Sanitize `statementStore`.
             let statementStoreMaxSeenStatements = 0;
+            let statementStoreFalsePositiveRate = 0.0;
```
dmitry-markin, line 514: May be introduce a non-zero lower boundary here (like 0.0001) to set some practical limit on the filter size? Even though we support up to 1 MiB filters on the wire, I don't think it makes any practical sense to have filters more than a few kB in size.

light-base/src/json_rpc_service/background.rs
```diff
@@ -2873,6 +2984,13 @@ pub(super) async fn run<TPlat: PlatformRef>(
                         me.statement_subscriptions
                             .insert(subscription_id.clone(), filter);
 
+                        if !me.statement_affinity_stale {
+                            me.statement_affinity_stale = true;
+                            me.next_statement_affinity_update = Some(Box::pin(
+                                me.platform.sleep(STATEMENT_AFFINITY_UPDATE_INTERVAL),
+                            ));
```
dmitry-markin: We can execute the affinity update immediately if the last one was more than 1 second in the past to not introduce a delay in most cases.

light-base/src/json_rpc_service/background.rs
```diff
@@ -2886,6 +3004,14 @@ pub(super) async fn run<TPlat: PlatformRef>(
 
                     methods::MethodCall::statement_unsubscribeStatement { subscription } => {
                         let existed = me.statement_subscriptions.remove(&subscription).is_some();
+
+                        if existed && !me.statement_affinity_stale {
+                            me.statement_affinity_stale = true;
+                            me.next_statement_affinity_update = Some(Box::pin(
```
dmitry-markin: Also can directly update the affinity if the last update was more than 1 second in the past.

light-base/src/json_rpc_service/background.rs
```diff
@@ -740,11 +782,80 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                    );
+                    let mut peers_to_remove = Vec::new();
+                    for peer_id in me.v2_statement_peers.clone() {
+                        if let Err(
+                            SendTopicAffinityError::NoConnection
+                            | SendTopicAffinityError::ProtocolV1,
+                        ) = me
+                            .network_service
+                            .send_topic_affinity(&peer_id, combined_filter.clone())
+                            .await
```
dmitry-markin: The JSON-RPC layer should not be interested in the individual peers / know about the peers. The topic affinity is akin to the protocol handshake that can be kept in `network_service.rs` (which has all the info about the peers) and sent when a new peer connects, without the need of extra async round trips to `json_rpc_service`.

`network_service.send_topic_affinity(peer, filter)` here should be `network_service.update_topic_affinity(filter)` instead.

light-base/src/network_service.rs
```diff
@@ -707,6 +741,10 @@ pub enum Event {
         peer_id: PeerId,
         statements: Vec<codec::Statement>,
     },
+    StatementProtocolConnected {
+        peer_id: PeerId,
+        version: StatementProtocolVersion,
+    },
```
dmitry-markin: No other components apart from `network_service.rs` should be concerned with this event, see the comment in `json_rpc_service`.

lib/src/network/codec/affinity.rs
dmitry-markin: I was also confused by this, but reading through https://en.wikipedia.org/wiki/Bloom_filter it looks like nothing depends on the total number of statements/topics — only the topics inserted into the filter matter?

wasm-node/javascript/src/internals/client.ts
dmitry-markin: Update: I don't think we need to expose the FPR to JS code at all, see other comments.

lib/src/network/codec/affinity.rs
```diff
@@ -0,0 +1,417 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+use super::statement::Topic;
+use alloc::vec::Vec;
+use core::hash::{BuildHasher, Hasher};
+use fastbloom::DefaultHasher as BloomDefaultHasher;
+
+/// Maximum number of bits allowed in a bloom filter received from the network.
+/// 1 MiB (the notification size budget) = 8_388_608 bits.
+const MAX_BLOOM_BITS: usize = 1024 * 1024 * 8;
```
dmitry-markin, line 25: A discussion here may be relevant: https://github.com/paritytech/polkadot-sdk/pull/11329/changes#r3084897017

dmitry-markin (approved):
Left some nits, but the most important things are:

1. The RPC layer must not be concerned with the individual network peers. It should update the affinity "handshake" in `network_service.rs` and all the message sending should happen there.
2. I don't think we can realistically expect/demand the developers of web apps to understand the internals of the statement store protocol. I.e., we shouldn't expose the false positive error rate to the JS clients, because people won't know what to pass there anyway. Seeing FPR as a traffic reduction factor (i.e., 1% FPR means 1% of the full bandwidth sneaks through the filter now), ideally we decide what it should be and set to a sane value. I don't know, may be 100x decrease in traffic from statements is not enough and we should set a lower FPR.

alexggh:
> I don't think we can realistically expect/demand the developers of web apps to understand the internals of the statement store protocol. I.e., we shouldn't expose the false positive error rate to the JS clients, because people won't know what to pass there anyway. Seeing FPR as a traffic reduction factor (i.e., 1% FPR means 1% of the full bandwidth sneaks through the filter now), ideally we decide what it should be and set to a sane value. I don't know, may be 100x decrease in traffic from statements is not enough and we should set a lower FPR.

Not really sure about this, in the end smoldot gets started by the Triangle User agents(not the web apps), so I think at best we should give us the flexibility to be able to override it from there, rather than smoldot directly deciding this value, having a sane default values and making the FPR value optional makes sense.

Picking the  FPR  value is a tradeoff between bandwidth and how much privacy you want to get by not declaring full interests to the nodes.

light-base/src/network_service.rs
AndreiEres: Implemented and merged

lib/src/network/codec/affinity.rs
AndreiEres: Opened https://github.com/paritytech/polkadot-sdk/issues/11776

lib/src/network/codec/affinity.rs
AndreiEres: Opened https://github.com/paritytech/smoldot/issues/3198 to move the discussion there

lib/src/network/codec/statement.rs
AndreiEres: I think it shouldn't. This is here just for consistency. Want me to remove it?

alexggh: Let's put a comment, explaining that it wont receive such filters.

light-base/src/json_rpc_service/background.rs
AndreiEres: Moved to config

lib/src/network/codec/statement.rs
AndreiEres: Done

wasm-node/javascript/src/internals/client.ts
AndreiEres: set a default

lib/src/network/codec/affinity.rs
AndreiEres: Decided not to do it

full-node/src/network_service.rs
AndreiEres: I don't think so

light-base/src/network_service.rs
AndreiEres: Addressed

light-base/src/json_rpc_service/background.rs
AndreiEres: Addressed

dmitry-markin (approved):
Nicely done! The implementation is solid and following the smoldot conventions!

lib/src/network/codec.rs
AndreiEres: Yes, it works

PR #3154, opened 2026-04-01
Log when statement protocol substream opens

AndreiEres:
### Problem
When a JSON-RPC client has an active statement subscription, there is no observable signal telling us whether smoldot is actually connected to any `/statement/1` peer. If none are connected, the subscription silently receives no data.

### Solution
Add a `Trace` log on `StatementProtocolConnected` in the network service. This gives operators a way to confirm statement substreams are actually opening, without adding new API surface or blocking the JSON-RPC thread.

Replaces the previous approach (`statement_peers_list` / `statement_connected_peers` + warning) after review feedback that those changes were overkill and the broadcast-filtering justification was invalid (peers without the statement store never accept a statement substream in the first place).

### DIFF
--- a/light-base/src/network_service.rs
+++ b/light-base/src/network_service.rs
@@ -3161,6 +3161,16 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
                 chain_id,
                 version,
             }) => {
+                log!(
+                    &task.platform,
+                    Trace,
+                    "network",
+                    "statement-protocol-open-success",
+                    chain = &task.network[chain_id].log_name,
+                    peer_id,
+                    ?version,
+                );
+
                 if matches!(version, codec::StatementProtocolVersion::V2) {
                     task.v2_statement_peers
                         .entry(chain_id)


light-base/src/json_rpc_service/background.rs
```diff
@@ -2870,6 +2870,15 @@ pub(super) async fn run<TPlat: PlatformRef>(
                             hex::encode(id)
                         };
 
+                        if me.network_service.peers_list().await.next().is_none() {
```
alexggh: Is this the list of peers connects on statement protocol or all peers ?

AndreiEres: Oh, you even don't know how deep is your question

light-base/src/json_rpc_service/background.rs
```diff
@@ -2870,6 +2870,15 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
                         };
 
+                        if me.network_service.statement_peers_list().await.next().is_none() {
+                            log!(
+                                &me.platform,
+                                Warn,
+                                &me.log_target,
+                                "No peers connected; statement subscription won't receive data",
+                            );
+                        }
```
bkchr: All these changes in this pr just to print some warning? This is total overkill, especially as you also block the entire json thread for this.

bkchr (changes requested):
Not sure we want this, especially in its current form. Blocking the entire json thread for a warning isn't that smart.

light-base/src/json_rpc_service/background.rs
AndreiEres: if we decide not to print this, I would keep the other changes anyway. Otherwise we will gossip our statements to all nodes even if they don't have statement-protocol enabled

alexggh: I'm kind of the culprit for this change :D, since I asked Andrei about a way we can tell there is something wrong with this subscriptions, I agree the changes need here are overkill and probably do more harm than good.

What I am looking for is a way we can tell something is off in smoldot state when we have subscriptions, but we aren't connected to any statement/1 peer. Something, so that we aren't flying completely blind.

bkchr: Print a trace log when we are opening a statement store substream successfully?

bkchr: > Otherwise we will gossip our statements to all nodes even if they don't have statement-protocol enabled

This makes no sense. We can not gossip statements to nodes that do not have the statement store substream open with us. If they don't have the statement store, they will not accept a statement store substream.

bkchr (approved):

alexggh (approved):

PR #3158, opened 2026-04-04
fix(sync): ignore late warp sync responses after transition

### LINKED ISSUE #3157: Late warp-sync responses can panic after #3155 transition to all-forks
## Summary

`9b4dccc1` (`Speed-up loading from the database (#3155)`) appears to introduce a regression in
`lib/src/sync/all.rs`: late warp-sync responses can panic after the state machine transitions
from warp sync into all-forks.

This can be reproduced deterministically with unit tests and does not require `host-sdk` or any
network traffic.

## Regression window

- good: `9b4dccc1` fully reverted locally in a separate probe worktree
- bad: `9b4dccc1` on `main`

## Deterministic reproduction

Added three unit tests in `lib/src/sync/all.rs`:

- `late_warp_sync_grandpa_response_is_ignored_after_transition`
- `late_warp_sync_storage_response_is_ignored_after_transition`
- `late_warp_sync_call_proof_response_is_ignored_after_transition`

The tests construct an `AllSync` with:

- `warp_sync: None`
- an outstanding request whose `RequestInfo.warp_sync` is still `Some(...)`

Then they directly inject the late response into:

- `grandpa_warp_sync_response`
- `storage_get_response`
- `call_proof_response`

### Before fix

`cargo test -p parity-smoldot late_warp_sync -- --nocapture`

Observed failures:

- panic at `lib/src/sync/all.rs:1351`
- panic at `lib/src/sync/all.rs:1385`
- panic at `lib/src/sync/all.rs:1419`

### After fix

Same command passes:

- `3 passed; 0 failed`

Full validation also passes:

- `cargo test -p parity-smoldot`

Result:

- `300 passed; 0 failed; 2 ignored`
- doc tests also pass

## Root cause

During the transition out of warp sync, the state machine removes the warp-sync object:

- `lib/src/sync/all.rs:997`

Later response handlers still assume warp sync is present and unconditionally unwrap it:

- `lib/src/sync/all.rs:1349`
- `lib/src/sync/all.rs:1381`
- `lib/src/sync/all.rs:1415`

That means late responses for already-issued warp-sync requests can hit `unwrap()` after the
transition has completed.

## Candidate fix

Treat late warp-sync responses as stale after transition and ignore them instead of unwrapping.

In practice, replacing:

- `self.warp_sync.as_mut().unwrap()...`

with:

- `if let Some(warp_sync) = self.warp_sync.as_mut() { ... }`

in the three response handlers is enough to make the deterministic repro pass and does not break
the existing package test suite.

## Live probe note

I also tried a raw JS/network probe against Paseo relay + Asset Hub Paseo. That path is useful as
supporting evidence, but it is too network-dependent to be the primary regression test:

- direct probe can succeed
- child-process AVA probe can still time out at 90s

So the unit tests above are the repeatable regression coverage that should protect against future
reintroductions.

replghost:
Fixes #3157.

This adds deterministic regression coverage for late warp-sync responses that arrive after the state machine transitions from warp sync to all-forks, and ignores those stale responses instead of panicking.

Validation:
- cargo test -p parity-smoldot late_warp_sync -- --nocapture
- cargo test -p parity-smoldot

### DIFF
--- a/lib/src/sync/all.rs
+++ b/lib/src/sync/all.rs
@@ -1346,11 +1346,15 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
         self.shared.sources[request.source_id.0].num_requests -= 1;
 
         if let Some(warp_sync_request_id) = request.warp_sync {
-            self.warp_sync.as_mut().unwrap().warp_sync_request_response(
-                warp_sync_request_id,
-                fragments,
-                is_finished,
-            );
+            // Responses can legitimately arrive after warp sync has transitioned into the
+            // all-forks state machine. In that case the request is stale and can be ignored.
+            if let Some(warp_sync) = self.warp_sync.as_mut() {
+                warp_sync.warp_sync_request_response(
+                    warp_sync_request_id,
+                    fragments,
+                    is_finished,
+                );
+            }
         }
 
         // TODO: type of request not always verified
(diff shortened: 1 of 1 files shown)


bkchr:
Duplicate of https://github.com/paritytech/smoldot/pull/3160

PR #3160, opened 2026-04-05
fix(sync): clear stale warp sync tracking after transition

### LINKED ISSUE #3159: Raw Paseo chain_getFinalizedHead can panic in all.rs:851
## Summary

On current `main` with the late warp-sync response fix from #3158 applied, a raw JS/Node probe against Paseo can still panic inside `lib/src/sync/all.rs` when calling legacy `chain_getFinalizedHead`.

This appears to be a separate issue from the late-response panic in #3157 and also blocks building a clean standalone repro for the warm-restart work in #3153.

## Reproduction

Using current smoldot JS bindings with a fresh Paseo spec from `dotli main`:

```bash
cd wasm-node/javascript
npm ci
node prepare.mjs --debug && rm -rf ./dist && npm run buildModules
node --input-type=module <<'EOF2'
import * as fs from 'node:fs';
import { start } from './dist/mjs/index-nodejs.js';

const spec = fs.readFileSync('./test/dotli-paseo.json', 'utf8');
const client = start({ maxLogLevel: 3 });
const chain = await client.addChain({ chainSpec: spec });
chain.sendJsonRpc(JSON.stringify({
  jsonrpc: '2.0',
  id: 1,
  method: 'chain_getFinalizedHead',
  params: []
}));
console.log(await chain.nextJsonRpcResponse());
client.terminate();
EOF2
```

## Observed result

The process panics in smoldot:

```text
[runtime-paseo] Finalized block runtime ready. Spec version: 2001000. Size of `:code`: 1.7 MiB.
Smoldot has panicked while executing task `sync-service-paseo`. This is a bug in smoldot. Please open an issue at https://github.com/smol-dot/smoldot/issues with the following message:
panicked at lib/src/sync/all.rs:851:17:
assertion failed: false
```

The JS side then observes the chain as destroyed and throws `AlreadyDestroyedError`.

## Code location

The panic site is here:

- `lib/src/sync/all.rs:851`

Specifically this branch:

```rust
(Some(_), None, _) | (None, Some(_), _) => {
    debug_assert!(false);
    None
}
```

That suggests request/source bookkeeping can still reach an inconsistent `(warp_sync, all_forks)` state on current Paseo startup.

## Why this matters

I was trying to make a standalone repro for #3153 (slow warm restart from restored finalized database). This raw panic prevents a clean direct-smoldot JS repro path, even before getting to the warm-restart comparison.

replghost:
Fixes #3159.

This adds deterministic regression coverage for mixed warp-sync/all-forks bookkeeping after the warp-sync transition and clears stale warp-sync IDs from sources and requests once the transition completes.

It also ignores mixed stale bookkeeping in add_request() instead of panicking.

Validation:
- cargo test -p parity-smoldot sync::all::tests -- --nocapture
- cargo test -p parity-smoldot

### DIFF
--- a/lib/src/sync/all.rs
+++ b/lib/src/sync/all.rs
@@ -1346,11 +1346,15 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
         self.shared.sources[request.source_id.0].num_requests -= 1;
 
         if let Some(warp_sync_request_id) = request.warp_sync {
-            self.warp_sync.as_mut().unwrap().warp_sync_request_response(
-                warp_sync_request_id,
-                fragments,
-                is_finished,
-            );
+            // Responses can legitimately arrive after warp sync has transitioned into the
+            // all-forks state machine. In that case the request is stale and can be ignored.
+            if let Some(warp_sync) = self.warp_sync.as_mut() {
+                warp_sync.warp_sync_request_response(
+                    warp_sync_request_id,
+                    fragments,
+                    is_finished,
+                );
+            }
         }
 
         // TODO: type of request not always verified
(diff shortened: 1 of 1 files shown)


bkchr (changes requested):
This should also not happen. Warp sync requests are not send in parallel and we process them directly and only move to all sync when this is done. So, if these crashes are happening we need to see why we are doing these extra requests and should fix it there.

replghost:
Superseded by #3184. The remaining stale warp-sync tracking fix and transition soak coverage have been folded into the consolidated restored warm-start PR.

PR #3155, opened 2026-04-01
Speed-up loading from the database

bkchr:
This fixes the loading times of smoldot when using the database. The problem basically was that it tried at startup to load the runtime,
 failed because of missing peers. The moment it finished the warp sync, it was still in the 10s cooldown after a failed runtime download. This is now fixed by this pull request.

This pull request also ensures to stop the warp sync after it is done. Also the `warp_sync_minimum_gap` is removed, because it should actually be faster to warp sync `X` blocks than downloading them one by one.

Should help with: https://github.com/paritytech/smoldot/issues/3153

PR #3161, opened 2026-04-05
fix(light-base): warm-start parachains from database state

### LINKED ISSUE #3153: Reduce browser boot time: ~7-8s warm restart to <2s for dotNS resolution
## Use case

We're building [dot.li](https://github.com/paritytech/dotli) - a `.dot` domain resolver that runs smoldot (`^2.0.40`) in a browser SharedWorker to resolve dotNS names via Asset Hub Paseo storage reads. We need three chains:

- **Paseo relay chain**
- **Asset Hub Paseo** (parachain - primary chain for dotNS resolution)
- **Bulletin Paseo** (parachain - for preimage/statement store operations)

The user flow is: type `something.dot` in the address bar, the browser extension intercepts it, boots smoldot, connects relay + Asset Hub, reads a storage slot from the dotNS contract, decodes an IPFS CID, fetches content. The entire flow is latency-sensitive - the user is waiting for a "page load".

## Problem

A full smoldot boot in the browser consistently takes **~7-8 seconds** on warm restarts (chain DB already in IDB), measured across Chrome on macOS. The breakdown:

| Phase | Time |
|---|---|
| WASM decode (base64 -> deflate -> `WebAssembly.compile`) | ~1-2s |
| `start()` / smoldot initialization | ~0.5s |
| `addChain()` relay (with `databaseContent`) | ~1-2s |
| `addChain()` parachain (with `databaseContent`) | ~1-2s |
| P2P connect + first finalized block | ~2-4s |
| **Total to first storage read** | **~7-8s** |

This is for a **warm restart** where we save/restore chain databases via IDB using `chainHead_unstable_finalizedDatabase`. Cold start (no chain DB) is similar or sometimes faster - which matches the findings in smol-dot/smoldot#2134, where `databaseContent` actually makes things slower rather than faster because smoldot's catch-up-from-near-head strategy is slower than its warp sync strategy.

## What we've tried

1. **Chain DB persistence via IDB** - We save chain databases to IndexedDB every 30-60 seconds and flush on pagehide. We pass them back via `databaseContent` on restart. Per smol-dot/smoldot#2134, this doesn't help and may hurt.

2. **`startWithBytecode` with pre-compiled `WebAssembly.Module`** - We considered caching the compiled module in IDB, but `WebAssembly.Module` IDB serialization support varies across browsers (works in Firefox, not reliably elsewhere).

3. **Running smoldot in a SharedWorker** - Keeps smoldot alive across navigations within the same origin, but the SharedWorker dies when all tabs close (per spec), so every "first tab" still pays the full boot cost.

## What would help

Our target is **<2 seconds** from `start()` to first storage read on warm restarts. Possible directions:

### 1. Fix the databaseContent catch-up strategy (smol-dot/smoldot#2134)

The maintainer noted in smol-dot/smoldot#2134 that when smoldot starts with `databaseContent` near the head, "there are periods of time where smoldot just waits for new blocks or finality messages instead of proactively querying things." If smoldot could proactively catch up from a near-head database (30-60 seconds stale) as fast as it does a warp sync, this alone could cut 3-5 seconds.

### 2. Expose a snapshot/restore API for the full smoldot state

Instead of just the chain database string, allow serializing the entire `Client` state (connection table, peer cache, verified blocks, etc.) to an `ArrayBuffer` and restoring it. Something like:

```typescript
const snapshot: ArrayBuffer = await client.snapshot();
// Later...
const client = startFromSnapshot(bytecode, snapshot);
```

### 3. Optimistic reads before full sync

Allow storage reads against the last known finalized block hash before smoldot has confirmed connectivity via P2P. If the chain DB has a finalized block from 30 seconds ago, that state is almost certainly still valid for read-only use cases.

### 4. WASM module reuse across restarts

The current `compileBytecode()` does base64 decode -> deflate -> `WebAssembly.compile()` every time. If smoldot could accept a pre-instantiated module or provide a way to cache the compiled module cross-browser, it would save 1-2s per boot.

## Environment

- smoldot `^2.0.40` via `polkadot-api ^1.23.3`
- Browser: Chrome/Firefox (SharedWorker context)
- Chains: Paseo relay + Asset Hub Paseo + Bulletin Paseo
- Chain specs include `lightSyncState` (updated periodically)
- Chain DB persistence: manual IDB save/restore of `chainHead_unstable_finalizedDatabase` output

## Related

- smol-dot/smoldot#2134 - Starting smoldot with databaseContent takes longer than without it

replghost:
Addresses a major contributor to #3153.

This preserves database-backed parachain chain information during startup and uses it to warm-start parachain sync instead of always restarting from a fresh relay-chain finalized-head fetch.

What changes:
- keep restored parachain chain information when `database_content` came from `chainHead_unstable_finalizedDatabase`
- start the parachain network service from that restored finalized head
- skip relay finalized-head fetch on warm start when restored parachain state exists
- skip parachain consensus bootstrap when the restored state already has known consensus information
- add a repeatable JS probe for cold vs warm finalized-database readiness in `wasm-node/javascript/test/parachain-finalized-database-probe.mjs`

Why this does not close #3153 by itself:
- `#3153` is defined by the end-to-end browser warm-restart outcome
- later validation showed other major contributors also matter, including `#3155` and `#3162`
- the full issue should remain open until the dot.li/browser benchmark is rerun against the `<2s` target on the fixed build

Validation:
- `cargo test -p parity-smoldot-light`
- `node --check wasm-node/javascript/test/parachain-finalized-database-probe.mjs`
- `git diff --check`

Live validation notes:
- I ran direct cold/warm probes against Kusama Asset Hub, Paseo Asset Hub, and Westend Asset Hub from this branch.
- The direct in-process live runs still did not complete a clean cold->warm finalized-database cycle in a reasonable window, which is why the new probe is included in this PR for longer soak runs.
- The warm-start code path itself is covered deterministically in `parity-smoldot-light` tests.


replghost:
Follow-up after the initial warm-start patch:

I found another real bug in the add-chain path: parachain database-backed chain information was only passed into `start_parachain()` when the database chain info "won" over the chain-spec checkpoint. If the checkpoint stayed authoritative, the restored parachain state was silently dropped for warm-start purposes.

That is fixed on this PR now in commit `e462af40` (`fix(light-base): preserve parachain db state alongside checkpoints`).

What changed:
- still use the checkpoint as the selected sync starting point when it is newer
- but keep database-backed parachain chain info separately as warm-start input
- add regression coverage for the "checkpoint wins, but DB state still exists" case

Validation:
- `cargo test -p parity-smoldot-light`

Live confirmation on Paseo + Asset Hub Paseo with:

```bash
cd /tmp/smoldot-3153-warm-restart/wasm-node/javascript
SMOLDOT_RELAY_SPEC_PATH=../../../demo-chain-specs/paseo.json \
SMOLDOT_PARA_SPEC_PATH=../../../demo-chain-specs/paseo_asset_hub.json \
SMOLDOT_REPRO_LOG_LEVEL=4 \
npm run probe:parachain-restart 2>&1 | rg 'Warm-starting parachain|Using restored parachain consensus state|Fetched parachain finalized head from relay chain|Failed to bootstrap parachain consensus|Waiting for relay chain to finalize a block|COLD_SUMMARY|WARM_SUMMARY'
```

Output:

```text
[sync-service-asset-hub-paseo] Waiting for relay chain to finalize a block...
[sync-service-asset-hub-paseo] Fetched parachain finalized head from relay chain at block #7229290
COLD_SUMMARY {"totalMs":18840,"relayDbReadyMs":29,"paraDbReadyMs":18815,"relayDbBytes":45565,"paraDbBytes":1990}
[sync-service-asset-hub-paseo] Warm-starting parachain from restored database state at block #7229290
[sync-service-asset-hub-paseo] Using restored parachain consensus state at block #7229290
WARM_SUMMARY {"totalMs":14,"relayDbReadyMs":4,"paraDbReadyMs":4,"relayDbBytes":45565,"paraDbBytes":1990,"relayReadyRatio":0.138,"paraReadyRatio":0}
```

So on Paseo this now does the thing we wanted:
- cold parachain finalized DB: `18.815s`
- warm parachain finalized DB: `4ms`

Kusama Asset Hub is still noisier and seems to have a separate bootstrap reliability problem, but this PR now proves the warm-start path itself works when restored parachain DB state is actually threaded through correctly.

bkchr (changes requested):
This doesn't work. You are ignoring the case that the db state is maybe older than the pruning window and you also ignore the case that the runtime could have changed in between.

replghost:
Superseded by #3184, which consolidates the restored parachain warm-start fixes and keeps the parachain restart probe.

PR #3163, opened 2026-04-07
Move statement dedup from network layer to per-subscription JSON-RPC layer

AndreiEres:
Global seen_statements LRU in the network layer prevented new subscriptions from receiving statements already seen by previous subscriptions. Moved dedup to per-subscription LRU caches in the JSON-RPC layer so each subscription tracks delivered statements independently. Removed StatementSubmitResult::Known since dedup is no longer tracked at the broadcast level.

### DIFF
--- a/light-base/src/json_rpc_service/background.rs
+++ b/light-base/src/json_rpc_service/background.rs
@@ -607,7 +607,7 @@ pub(super) async fn run<TPlat: PlatformRef>(
             StartStorageSubscriptionsUpdates,
             NotifyFinalizedHeads,
             NotifyNewHeadsRuntimeSubscriptions(Option<[u8; 32]>),
-            NetworkStatementsReceived(Vec<codec::Statement>),
+            NetworkStatementsReceived(Vec<([u8; 32], codec::Statement)>),
             MustSubscribeNetworkEvents,
         }
 
@@ -755,8 +755,8 @@ pub(super) async fn run<TPlat: PlatformRef>(
                 for (sub_id, topic_filter) in &me.statement_subscriptions {
                     let matching: Vec<methods::HexString> = statements
                         .iter()
-                        .filter(|s| topic_filter.matches(&s.topics))
-                        .map(|s| {
+                        .filter(|(_hash, s)| topic_filter.matches(&s.topics))
+                        .map(|(_hash, s)| {
                             methods::HexString(
                                 codec::encode_statement(s)
                                     .expect("re-encoding a decoded statement always succeeds; qed"),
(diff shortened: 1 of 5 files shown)


alexggh (approved):

light-base/src/json_rpc_service/background.rs
```diff
@@ -481,6 +488,33 @@ enum TransactionWatchTy {
...
+impl StatementSubscription {
+    fn new(topic_filter: methods::TopicFilter, max_seen: Option<NonZero<usize>>) -> Self {
+        Self {
+            topic_filter,
+            seen: max_seen
+                .map(|cap| lru::LruCache::with_hasher(cap, fnv::FnvBuildHasher::default())),
+        }
+    }
+
+    fn should_deliver(&mut self, hash: &[u8; 32], statement: &codec::Statement) -> bool {
```
alexggh:
```suggestion
    fn try_accept(&mut self, hash: &[u8; 32], statement: &codec::Statement) -> bool {
```

Since this is mutating.

light-base/src/network_service.rs
```diff
@@ -96,7 +97,7 @@ impl StatementProtocolConfig {
 impl Default for StatementProtocolConfig {
     fn default() -> Self {
         StatementProtocolConfig {
-            max_seen_statements: NonZeroUsize::new(65536).expect("65536 is not zero; qed"),
+            max_seen_statements: NonZeroUsize::new(512).expect("512 is not zero; qed"),
```
alexggh: It seems excessively small maybe 4096 :D

light-base/src/network_service.rs
```diff
@@ -78,6 +78,7 @@ pub use service::{ChainId, EncodedMerkleProof, PeerId, QueueNotificationError};
 /// Configuration for the Statement Store protocol.
 #[derive(Debug, Clone)]
 pub struct StatementProtocolConfig {
```
dmitry-markin, line 80: Now `StatementProtocolConfig` (and `statement_protocol_config`) serves no purpose in `network_service.rs` and can be replaced with a boolean flag.

What about keeping just `enable_statement_protocol: bool` in this file and moving the configuration to the place where it's used?

dmitry-markin (approved):

light-base/src/network_service.rs
AndreiEres: Decided to postpone it, https://github.com/paritytech/smoldot/issues/3193

light-base/src/network_service.rs
AndreiEres: done

light-base/src/json_rpc_service/background.rs
AndreiEres: decided to use `accept`

PR #3165, opened 2026-04-09
Get ready for npm publish automation

lrubasze:
This is to prepare for publishing `smoldot` npm with `npm_publish_automation`

https://github.com/paritytech/npm_publish_automation/blob/main/README.md

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -180,21 +180,24 @@ jobs:
       - run: rustup target add wasm32-unknown-unknown
       - uses: actions/setup-node@v6.1.0
         with:
-          # Set to the oldest version still maintained, in order to ensure compatibility. See <https://nodejs.dev/en/about/releases/>
           node-version: 18
       - uses: Swatinem/rust-cache@v2
       - run: npm install
         working-directory: ./wasm-node/javascript
-      - run: npm publish --unsafe-perm --dry-run
+      - run: npm pack
         working-directory: ./wasm-node/javascript
-      - uses: JS-DevTools/npm-publish@v3
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-        # Warning: this GitHub action doesn't seem to run prepublish scripts, hence
-        # the `npm publish --dry-run` done right above is important to ensure this.
+      - uses: actions/upload-artifact@v4
         with:
-          token: ${{ secrets.NPM_TOKEN }}
-          package: ./wasm-node/javascript/package.json
-          access: public
+          name: package
+          path: ./wasm-node/javascript/*.tgz
+      - uses: octokit/request-action@v2.x
+        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -185,16 +185,20 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: npm install
         working-directory: ./wasm-node/javascript
-      - run: npm publish --unsafe-perm --dry-run
+      - run: npm pack
         working-directory: ./wasm-node/javascript
-      - uses: JS-DevTools/npm-publish@v3
+      - uses: actions/upload-artifact@v4
+        with:
+          name: package
```
skunert: Maybe a proper name should be used here

.github/workflows/deploy.yml
```diff
@@ -185,16 +185,20 @@ jobs:
...
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
-        # Warning: this GitHub action doesn't seem to run prepublish scripts, hence
-        # the `npm publish --dry-run` done right above is important to ensure this.
         with:
-          token: ${{ secrets.NPM_TOKEN }}
-          package: ./wasm-node/javascript/package.json
-          access: public
+          route: POST /repos/paritytech/npm_publish_automation/actions/workflows/publish.yml/dispatches
+          ref: main
+          inputs: '${{ format(''{{ "repo": "{0}", "run_id": "{1}" }}'', github.repository, github.run_id) }}'
```
skunert, line 206: AI suggests that we shall double check that publish only happens on version bump. Or does the npm automation handle that?

skunert (approved):

AndWeHaveAPlan (approved):

.github/workflows/deploy.yml
AndWeHaveAPlan: as [npmjs docs](https://docs.npmjs.com/policies/unpublish) says
> So if you've ever published a package called "bob" at version 1.1.0, no other package can ever be published with that name at that version. This is true even if that package is unpublished.

dmitry-markin:
@lrubasze can you double-check that we are publishing only from `main` and not from pull requests?

.github/workflows/deploy.yml
lrubasze: We are safe here
https://github.com/paritytech/npm_publish_automation/blob/main/publishPackages.ts#L111-L133

dmitry-markin:
> @lrubasze can you double-check that we are publishing only from `main` and not from pull requests?

False alarm, the check is there:
https://github.com/paritytech/smoldot/blob/165b771d69a569899f638ff5a6191d1afcb368d4/.github/workflows/deploy.yml#L195

.github/workflows/deploy.yml
lrubasze: Nice catch. Let's do the same as `polkadot-bulletin-chain` does
https://github.com/paritytech/polkadot-bulletin-chain/blob/main/.github/workflows/npm-publish.yml#L62-L66

dmitry-markin (approved):

PR #3184, opened 2026-04-10
fix(light-base): warm-start restored parachains

replghost:
## Summary

Consolidates the restored/warm-start correctness work for parachains:

- ignore late warp-sync responses after the sync transition
- warm-start parachains from saved database chain information
- preserve parachain database state even when a chain-spec checkpoint wins for initial sync
- retain finalized runtime state for the parachain runtime service
- merge the warm-start and restored RPC paths
- add a node-side parachain restart probe for cold-to-warm database restore behavior

This is intended to replace the earlier piecemeal warm-start PRs.

## Validation

Validated locally on the rebased stack:

```text
cargo fmt --check
cargo check -p parity-smoldot-light
cargo test -p parity-smoldot-light initial_parachain_chain_information -- --nocapture
npm install
npm run build
npm run probe:parachain-restart
```

Representative probe result before the follow-up runtime-snapshot optimization:

```text
COLD_SUMMARY totalMs=64648 paraDbReadyMs=64606
WARM_SUMMARY totalMs=2835 paraDbReadyMs=2817 paraReadyRatio=0.044
```

## Notes

A follow-up performance PR stacks on this and restores a block-bound finalized parachain runtime snapshot so warm recovery can skip re-downloading the parachain runtime.


replghost:
Superseded by #3185, which includes this warm-start stack plus the restored-runtime snapshot and proof-verification changes. Keeping the branch intact; closing this PR to focus review on the combined host-facing warm recovery path.

Issue #3164, opened 2026-04-08
Full node panics with unreachable code in sync/all.rs during warp sync on Paseo


KarimJedda:
The smoldot full node panics at `lib/src/sync/all.rs:1279` with `internal error: entered unreachable code` while syncing the Paseo     
testnet. The node successfully initializes, connects to peers, and begins syncing, but crashes shortly after reaching the chain head.
                                                                                                                                       
## Environment  

- **smoldot version:** main branch (commit 9b4dccc1)                                                                                   
- **Target:** aarch64-unknown-linux-musl (cross-compiled)                                                                     
- **Chain:** Paseo Testnet (standalone relay chain, raw chain spec) from here https://github.com/paseo-network/paseo-chain-specs/blob/main/paseo.raw.json 
                                                                                                                                       
## Steps to reproduce
                                                                                                                                       
1. Cross-compile the full node for `aarch64-unknown-linux-musl`                                                                        
2. Run on the target device:
   ./full-node run                                                                                                                     
     --path-to-chain-spec paseo.raw.json                                                                                               
     --database-cache-size 64M          
     --relay-chain-database-cache-size 64M                                                                                             
     --json-rpc-address 0.0.0.0:9944      
     --tmp                                                                                                                             
                
## Observed behavior                                                                                                                   
                
The node starts from genesis, connects to 1 peer, and syncs up to around block #11,136,613. Several `JustificationEngineMismatch` finality warnings appear during sync. Then it panics:
                                                                                                                                       
Paseo Testnet   #11136613 [=======>] #11136616 (🔗 1) (🌐 1)                                                                           

thread 'main' (8503) panicked at lib/src/sync/all.rs:1279:21:  internal error: entered unreachable code
                                                                                                                                       
## Analysis
                                                                                                                                       
The panic is at line 1279 in `AllSync::block_request_response()`. The code expects `self.warp_sync` to be `Some` when                  
`request.warp_sync` is `Some`, but `self.warp_sync` has already been set to `None` (likely because warp sync completed) while there was still an in-flight request with a warp sync component.                                                                                
                
```rust
if let (true, Some(warp_sync_request_id)) = (is_first_block, request.warp_sync) {
    let Some(warp_sync) = self.warp_sync.as_mut() else {                                                                               
        unreachable!()  // <-- panics here
    };                                                                                                                                 
```
       
The same pattern appears on line 1267-1268 for the empty-blocks case, which has the same potential issue.                              
                
[...]

PR #3185, opened 2026-04-10
fix(sync): handle stale warp sync bookkeeping after transition to all-forks

### LINKED ISSUE #3157: Late warp-sync responses can panic after #3155 transition to all-forks

replghost:
## Summary

Fixes #3157, #3159, #3164.

After warp sync transitions to all-forks, stale warp-sync source/request IDs remain in the `shared.sources` and `shared.requests` slabs. Response handlers that assume `self.warp_sync` is still present panic on `unreachable!()` or `unwrap()`.

## Root cause

The transition at `all.rs:998-1002` calls `self.warp_sync.take()` but does not clear per-source `source.warp_sync` or per-request `request.warp_sync` mappings. Parallel block-body downloads (`RequestDetail::BlocksRequest { request_bodies: true }`) are tagged as warp-sync-scoped and can be outstanding when the transition occurs.

Independent corroboration: @KarimJedda hit the same panic on the full-node warp-syncing Paseo in #3164 (`all.rs:1279`).

## Fix

`clear_transition_stale_state()` is called at the transition point and nulls `source.warp_sync` on all sources and `request.warp_sync` / `request.all_forks` on all outstanding requests. Defense-in-depth `if let Some(warp_sync)` guards are added to all five response handlers so a future regression in the transition cleanup cannot re-expose the panic.

## Tests

11 unit tests covering the transition cleanup and all response-handler paths:

- `late_warp_sync_grandpa_response_is_ignored_after_transition`
- `late_warp_sync_storage_response_is_ignored_after_transition`
- `late_warp_sync_call_proof_response_is_ignored_after_transition`
- `late_warp_sync_empty_blocks_response_is_ignored_after_transition`
- `late_warp_sync_block_body_response_is_ignored_after_transition`
- `stale_warp_sync_source_mapping_does_not_panic_when_adding_request`
- `transition_cleanup_clears_stale_warp_sync_tracking`
- `repeated_mixed_state_add_request_and_cleanup_stays_consistent`
- `repeated_transition_cleanup_clears_existing_mixed_tracking`

All tests panic at the original `unreachable!()` sites when the guards are reverted — confirmed locally.

## Validation

```text
cargo fmt --check
cargo test -p smoldot --lib sync::all::tests    # 9 passed
cargo test -p smoldot --lib                      # all passed
cargo check -p smoldot-light
```

### DIFF
--- a/lib/src/sync/all.rs
+++ b/lib/src/sync/all.rs
@@ -1263,24 +1263,27 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
                 None => {
                     if let (true, Some(warp_sync_request_id)) = (is_first_block, request.warp_sync)
                     {
-                        let Some(warp_sync) = self.warp_sync.as_mut() else {
-                            unreachable!()
-                        };
-                        // TODO: report source misbehaviour
-                        warp_sync.remove_request(warp_sync_request_id);
+                        // Responses can legitimately arrive after warp sync has transitioned
+                        // into the all-forks state machine. In that case the request is stale
+                        // and can be ignored.
+                        if let Some(warp_sync) = self.warp_sync.as_mut() {
+                            // TODO: report source misbehaviour
+                            warp_sync.remove_request(warp_sync_request_id);
+                        }
                     }
                     break ResponseOutcome::Queued;
                 }
             };
 
             if let (true, Some(warp_sync_request_id)) = (is_first_block, request.warp_sync) {
-                let Some(warp_sync) = self.warp_sync.as_mut() else {
-                    unreachable!()
-                };
-                warp_sync.body_download_response(
-                    warp_sync_request_id,
-                    block.scale_encoded_extrinsics.clone(), // TODO: clone?
-                );
+                // Responses can legitimately arrive after warp sync has transitioned into the
+                // all-forks state machine. In that case the request is stale and can be ignored.
+                if let Some(warp_sync) = self.warp_sync.as_mut() {
+                    warp_sync.body_download_response(
+                        warp_sync_request_id,
+                        block.scale_encoded_extrinsics.clone(), // TODO: clone?
+                    );
+                }
             }
 
             if let Some(blocks_append) = all_forks_blocks_append {
@@ -2639,6 +2642,39 @@ mod tests {
         let (mut sync, request_id) = sync_with_late_warp_sync_request();
 
         let (_user_data, outcome) = sync.call_proof_response(request_id, Vec::new());
+
+        assert!(matches!(outcome, ResponseOutcome::Queued));
+        assert!(sync.shared.requests.is_empty());
+        assert_eq!(sync.shared.sources[0].num_requests, 0);
+    }
+
+    #[test]
+    fn late_warp_sync_empty_blocks_response_is_ignored_after_transition() {
+        let (mut sync, request_id) = sync_with_late_warp_sync_request();
+
+        let (_user_data, outcome) = sync.blocks_request_response(
+            request_id,
+            core::iter::empty::<BlockRequestSuccessBlock<()>>(),
+        );
+
+        assert!(matches!(outcome, ResponseOutcome::Queued));
+        assert!(sync.shared.requests.is_empty());
+        assert_eq!(sync.shared.sources[0].num_requests, 0);
+    }
+
+    #[test]
+    fn late_warp_sync_block_body_response_is_ignored_after_transition() {
+        let (mut sync, request_id) = sync_with_late_warp_sync_request();
+
+        let (_user_data, outcome) = sync.blocks_request_response(
+            request_id,
+            core::iter::once(BlockRequestSuccessBlock {
+                scale_encoded_header: Vec::new(),
+                scale_encoded_justifications: Vec::new(),
+                scale_encoded_extrinsics: Vec::new(),
+                user_data: (),
+            }),
+        );
 
         assert!(matches!(outcome, ResponseOutcome::Queued));
         assert!(sync.shared.requests.is_empty());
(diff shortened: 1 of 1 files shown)


light-base/src/sync_service/parachain.rs
```diff
@@ -1228,47 +1315,140 @@ async fn fetch_parachain_head_from_relay<TPlat: PlatformRef>(
...
+        "Verifying block-bound restored parachain runtime hint"
+    );
+
+    let proof = network_service
+        .clone()
+        .storage_proof_request(
+            peer_id.clone(),
+            codec::StorageProofRequestConfig {
+                block_hash,
+                keys: [&b":code"[..], &b":heappages"[..]].into_iter(),
```
skunert: Since you request `:code` here, the fetched storage proof will include the complete blob, effectively redownloading the full runtime, which you wanted to avoid. 

Instead you need to pass here the closest merkle ancestor that we already have above. Then it will only fetch the parent node, we can compare it to the hint we have, and then use the runtime blob we had stored.

light-base/src/database.rs
```diff
@@ -207,31 +254,86 @@ pub fn decode_database(encoded: &str, block_number_bytes: usize) -> Result<Datab
...
         (Some(mv), Some(sv), Some(an)) => Some(DatabaseContentRuntimeCodeHint {
-            code: base64::Engine::decode(&base64::engine::general_purpose::STANDARD_NO_PAD, sv)
-                .map_err(|_| ())?,
-            code_merkle_value: hex::decode(mv).map_err(|_| ())?,
-            closest_ancestor_excluding: an
-                .as_bytes()
-                .iter()
-                .map(|char| Nibble::from_ascii_hex_digit(*char).ok_or(()))
-                .collect::<Result<Vec<Nibble>, ()>>()?,
+            code: sv.clone(),
```
skunert: Here we clone the complete multi-mb wasm blob and keep it in memory. I think we could get away with less memory usage by just signaling whether the runtime is the finalized one instead of having the blob multiple times in the `DatabaseContent` struct.

replghost:
@skunert Thanks for the review — both points addressed:

**Comment 1 (re-downloading :code in the proof):** Fixed in 9f0374de. The verification now requests a sibling key (same trie ancestor path, different child nibble) instead of `:code`. The proof includes only the ancestor branch node (~500 bytes) with the merkle value of `:code`'s subtree, not the full ~2 MiB blob. We compare the proof's merkle value to the stored hint and compile the stored bytes directly if it matches.

**Comment 2 (cloning the wasm blob):** Acknowledged — will address in a follow-up. The clone is a one-time cost at database encode time, so it's lower priority than the proof size fix.

replghost:
@skunert — heads up: this PR has been split into smaller single-issue PRs per review feedback:

- **This PR (#3185)** is now just the warp-sync transition fix (`all.rs` — `clear_transition_stale_state` + defense-in-depth guards + 11 unit tests). It no longer touches `parachain.rs` or `database.rs`.

- **PR B** (coming shortly) will fix parachain cold start (#3202 — `known_finalized_runtime: Some(finalized_runtime)`).

- **PR C** (coming after B) will have the warm-start feature. Your two review comments apply there:
  1. Verification via merkle ancestor instead of full `:code` re-download — already implemented in the warm-start branch (`9f0374de`), will carry forward to PR C.
  2. Avoiding the multi-MB `:code` clone in `DatabaseContent` — will address in PR C.

replghost:
Closing — the warp sync transition fixes landed on main via dedicated commits (3b15a51, 2064c09, f8bd801) after this PR was split. skunert's review feedback was addressed in the split PRs.

Issue #3142, opened 2026-03-27
statement-store: close light node test coverage gaps


AndreiEres:
### Summary

Tracking issue for the remaining scenarios in statement-store light node tests. See details in the [testing strategy](https://docs.google.com/document/d/1dMEkH82ZxFdUN_vX-Qhl2kdiHVVxD67AYhYCTmTIZl4/edit?tab=t.1d4x55fgv2og).

### Tasks

**P0**
- [x] Wire protocol: unit tests cover edge cases around truncated payloads and compact encoding boundaries, polkadot-sdk cross-compatibility, can be extended with fuzzing https://github.com/paritytech/smoldot/pull/3187
- [x] Submission: unit tests cover result mapping, Zombienet tests verify statements sent by the light node reach connected full nodes
  - [x] Unit https://github.com/paritytech/smoldot/pull/3188
  - [x] Zombienet https://github.com/paritytech/smoldot/pull/3191
- [x] Reception: unit tests cover bloom filter construction, false-positive discard, deduplication, subscription notifications, and unsubscribe; Zombienet tests verify only subscribed statements are delivered without duplicates
  - [x] Unit https://github.com/paritytech/smoldot/pull/3188
  - [x] Zombienet https://github.com/paritytech/smoldot/pull/3191
- [x] Peer connection: Zombienet tests verify connectivity is maintained or restored when full nodes drop https://github.com/paritytech/smoldot/pull/3191

**P1**
- [ ] Long-session stability: Versi tests verify a light node remains stable while connected to real full nodes for hours
- [ ] Browser sanity check: Zombienet tests verify the basic flow works in a headless or real browser environment

AndreiEres:
Long-session stability tests failed.

The goal was to run 500 smoldot-based clients at load ≈540 000 st/hr and measure latency. But that isn't possible with smoldot today: the WASM client starts panicking once more than ~100 clients are spawned (across separate pods). At this scale it is not stable enough for long sessions.

The testing work that this issue was tracking is done, so I'm closing it. The follow-up — switching from “confirm long-session stability” to “find the circumstances under which smoldot loses stability” — will be tracked in a separate issue.

cc @P1sar @alexggh

PR #3191, opened 2026-04-15
statement-store: Add zombienet test

AndreiEres:
Part of https://github.com/paritytech/smoldot/issues/3142


.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,99 @@
...
+  push:
+    branches: [main]
+  workflow_dispatch:
+
+concurrency:
+  group: zombienet-e2e-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  POLKADOT_IMAGE: docker.io/parity/polkadot:stable2603
```
lrubasze: Just thinking that maybe we could use latest image instead of pinning to the specific version?

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,99 @@
...
+  CUMULUS_IMAGE: docker.io/parity/polkadot-parachain:stable2603
+
+jobs:
+  light-node-submission:
+    runs-on: ubuntu-latest
```
lrubasze, line 17: Maybe it'd be better to use similar approach as it is used in `polkadot-sdk` zombienet CI tests.
Which is:
- using a Parity runner - see [this](https://github.com/paritytech/polkadot-sdk/blob/e3f0d25d43da51712ebc669c354d555132adcf43/.github/workflows/zombienet_cumulus.yml#L65)
- running in the container - see [this](https://github.com/paritytech/polkadot-sdk/blob/e3f0d25d43da51712ebc669c354d555132adcf43/.github/workflows/zombienet_cumulus.yml#L70-L72)

Runners and images are specified [here](https://github.com/paritytech/polkadot-sdk/blob/e3f0d25d43da51712ebc669c354d555132adcf43/.github/zombienet-env#L3-L5).

lrubasze (approved):
I think this a really good starting point for setting up zombienet tests.
In general would be good keep similar approach for setting up CI as in `polkadot-sdk` repo. 
I don't mean copy-pasting everything because the setup there is super complicated.
I mean using the same runners, images, etc, some basic stuff.

We should also think of making it scalable, adding more tests in parallel. But we can cover it the next PRs.

Also feedback from @alvicsam and @pepoviola is welcome

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,99 @@
...
+        run: npm ci
+        working-directory: e2e-tests/js
+
+      - name: Extract polkadot + polkadot-parachain binaries
+        # The docker provider's multiaddr is broken for host-side clients (it advertises the
+        # container bridge IP with the host-mapped port, which is unroutable). Run native
+        # instead: use Docker only as a delivery mechanism for versioned binaries.
+        run: |
+          set -euo pipefail
+          mkdir -p "$GITHUB_WORKSPACE/bin"
```
pepoviola, line 118: I think we can just download the bins from the gh release page, we do something similar here
https://github.com/paritytech/zombienet-sdk/blob/main/.github/workflows/ci_integration.yml#L171C1-L181C21

e2e-tests/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    key.extend_from_slice(account_id.as_ref());
+    key
+}
+
+/// Creates a parachain chain spec with a statement allowance for each given public key.
+pub fn create_para_chain_spec_with_allowances(
+    pubkeys: &[[u8; 32]],
+    base_dir: &Path,
+) -> Result<PathBuf, anyhow::Error> {
+    let template = include_str!("../chain-specs/people-westend-local-spec.json");
```
pepoviola: can we generate the chain-spec from the runtime or this one is customized?

e2e-tests/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+            p.with_id(1004)
+                .with_chain_spec_path(para_spec_path.to_str().expect("Valid UTF-8 path"))
+                .with_default_command("polkadot-parachain")
+                .with_default_image(images.cumulus.as_str())
+                .with_default_args(vec![
+                    "--force-authoring".into(),
+                    "--enable-statement-store".into(),
+                    "-linfo,statement-store=info,statement-gossip=info".into(),
+                ])
+                .with_collator(|n| n.with_name("collator-0"))
```
pepoviola: if you set `.bootnode(true)` zombienet will add them to the chain-spec. So you can remove the patching logic.

.github/workflows/zombienet.yml
```diff
@@ -0,0 +1,99 @@
...
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+
+      - uses: actions/setup-node@v6.1.0
+        with:
+          node-version: '18'
```
pepoviola: This version is `EOL`, I think we should bump to  at least `22`.
```suggestion
          node-version: '22'
```

https://nodejs.org/en/about/previous-releases#looking-for-the-latest-release-of-a-version-branch

alvicsam:
Since I don't know specifics of the project I can only evaluate the ci workflow. I fully agree with @lrubasze that it's better to have a similar setup as in polkadot-sdk and particularly with [this comment](https://github.com/paritytech/smoldot/pull/3191/changes#r3128748913)

e2e-tests/src/statement.rs
```diff
@@ -0,0 +1,337 @@
...
+pub fn spawned_chain_spec_paths(
+    network: &Network<LocalFileSystem>,
+) -> Result<(PathBuf, PathBuf), anyhow::Error> {
+    let base_dir = PathBuf::from(
+        network
+            .base_dir()
+            .ok_or_else(|| anyhow!("network has no base_dir"))?,
+    );
+
+    let relay_chain = network.relaychain().chain();
```
pepoviola, line 174: I will make `pub` the method to get this path in zombienet directly.

pepoviola (approved):

alvicsam (approved):

e2e-tests/Cargo.toml
alexggh, line 1: I think this project warrants a readme, describing:
1. Approach/architecture how everything is stiched together.
2. How to run the tests locally and what pre-requisites I need if any.

e2e-tests/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+    let status = std::process::Command::new("npm")
+        .arg("run")
+        .arg("build")
+        .current_dir(&js_dir)
+        .status()
+        .expect("failed to run npm build");
+    assert!(status.success(), "smoldot npm build failed");
+}
+
+fn is_dist_stale(dist_dir: &Path) -> bool {
```
alexggh: Nit: This seems brittle can't we leverage the npm build system ?

e2e-tests/js/statement_store_submission.js
alexggh, line 1: I don't get it why do we need this per-test files, the way I imagined is that we have some wrappers that start node with smoldot in them and then you can use subxt RPC from the rust code to connect to it and perform the test.

alexggh: Anyways, this is a good way forward for the testing story, so this can be addressed latter.

alexggh (approved):
Left you some comments on the approach, test scenarios are a good start, can't comment too much on the CI configuration.

PR #3192, opened 2026-04-15
Add some tests for Bitswap protocol implementation

lrubasze:
Tests added:
- `BitswapPeeringStrategy` - peer state transitions, slot assignment, ban logic, error paths
- Bitswap codec - decoding error paths (malformed protobuf, missing fields, invalid enums)
- `BitswapService` - JSON-RPC error code mapping for all error variants
- Bitswap substreams - negotiation, message send/receive, close handling (using existing `TwoEstablished` harness)


lib/src/libp2p/connection/established/tests.rs
```diff
@@ -830,3 +831,291 @@ fn outbound_substream_close_demanded() {
...
+    let (_, event) = connections.run_until_event();
+    match event {
+        either::Left(Event::BitswapOutOpenResult {
+            id,
+            result: Err((BitswapOutOpenErr::ProtocolNotAvailable, _)),
+        }) => {
+            assert_eq!(id, substream_id);
+        }
+        // Bob may also emit an InboundError before Alice sees the rejection.
+        either::Right(Event::InboundError(InboundError::NegotiationError(_))) => {}
```
lexnv: Shouldn't we always get `Event::BitswapOutOpenResult` when rejecting the inbound? It sounds a bit undeterministic that we get inbound-error sometimes and other times open-result?

lexnv (approved):
Nice! Thanks for double-checking everything @lrubasze 🙏

skunert (approved):
Nice!

dmitry-markin:
GitHub went crazy and wants to apply changes that are already in main. It looks like we need to manually rebase everything after 94c67a2e8982f0c7fea2f04ab48aa410710f050d on top of main and force push.

PR #3201, opened 2026-04-17
reproduction + fix

KarimJedda:
Disclaimer: heavily Claude assisted, if it's missing the point, please dismiss.

Fixes parachain finality stall — legacy JSON-RPC subscriptions (`chain_subscribeNewHeads`, `chain_subscribeFinalizedHeads`) never emit notifications for parachains, making smoldot unusable for any parachain app relying on finalized state.

**Bug 1: Stale warp_sync source/request IDs after warp sync transition** (`lib/src/sync/all.rs`)

When warp sync completes and transitions to all-forks, `self.warp_sync` is consumed but per-source and per-request `warp_sync` IDs are never cleared. Debug builds panic at `debug_assert!(false)`; release builds silently skip request tracking.

**Bug 2: Runtime service deadlock blocks all legacy JSON-RPC for parachains** (`light-base/src/runtime_service.rs`)

Parachain sync subscribes with `runtime_interest: false` → runtime service enters `FinalizedBlockRuntimeUnknown` → a guard (`if finalized_block_known`) prevents runtime downloads → tree never advances → state never transitions to `Known` → legacy subscriptions queued as pending are never processed. `chainHead_v1_follow` is unaffected because it creates direct subscriptions bypassing this path.

## Reproducer

```bash
# Before fix: panics in debug, legacy API dead in release
cargo run -p smoldot-light --example parachain_finality_repro

# After fix: exits 0 with VERDICT: HEALTHY
cargo run -p smoldot-light --example parachain_finality_repro

The reproducer connects to Paseo relay + Paseo Asset Hub, monitors finality via both legacy and chainHead_v1_follow APIs for 90 seconds, and reports a structured verdict with exit code.

Test plan

- Reproducer reports VERDICT: HEALTHY with both APIs showing finality events
- cargo test -p smoldot --lib — 374 tests pass, 0 failures


skunert:
Fixes look correct to me. One downside is of course that this reverts part of https://github.com/paritytech/smoldot/pull/3155, so we could run into that retry again. I reduced the time from 10s to 4s though, that should be more than enough. IIUC this timeout should trigger when no peers are connected, 10s seems excessive for that use case.

demo-chain-specs/asset-hub-paseo.json
skunert, line 1: IMO its good to have these chain-specs here for convenience 👍

skunert (approved):

BigTava (approved):
There should be at least 1 regression test for these state transitions.

PR #3210, opened 2026-04-21
fix: parachain warm-restart hang

replghost:
## Summary

Fix parachain warm-restart hang on `fetch_parachain_head_from_relay`, without regressing cold start, by moving the warp-sync wait into `runtime_service` and emitting a synthetic Finalized event on each new subscription.

## Problem

`fetch_parachain_head_from_relay()` ignored the finalized header already returned by `subscribe_all()` and awaited a **new** `Finalized` notification via `subscription.new_blocks.next().await`. On warm restart the relay is
already synced, so nothing arrives until the next GrandPa round (~6s on Polkadot, longer with slow peers), and parachain sync hangs at "Waiting for relay chain to finalize a block..."

## Fix

In `runtime_service`:

1. Gate `subscribe_all` on warp-sync completion. Every subscription it returns has a post-warp finalized header.
2. On each new subscription, push a synthetic `Notification::Finalized` for the current finalized block as the first stream event.

Consumers waiting on `Notification::Finalized` (e.g. the parachain task) now receive one immediately. 

## API change

New public method on `SyncService`:

```rust
pub async fn wait_warp_sync_finished(&self);
```

##  Smoke test
`e2e-tests/tests/smoke.rs` (added in #3234) didn't exercise the warp-sync path. On a freshly-spawned zombienet, peers report finalized below smoldot's warp_sync_minimum_gap = 32, all-forks catches up via normal GRANDPA finality before warp sync can engage, and `wait_warp_sync_finished()` never resolves.

Updated the test to wait for the relay to finalize 50 blocks past baseline before launching `smoldot`, so warp sync has a real target and WarpSyncFinished fires.

## Credits
- @replghost — original investigation, warm-restart fix, refactor (`try_fetch_parachain_head` /`wait_for_finalized_hash`)
- @lrubasze  — cold-start staleness diagnosis, `wait_warp_sync_finished()` gating


light-base/src/sync_service/parachain.rs
lexnv, line 1135: This should be moved on the else branch where the subscription.next is polled

light-base/src/sync_service/parachain.rs
```diff
@@ -1125,20 +1130,27 @@ async fn fetch_parachain_head_from_relay<TPlat: PlatformRef>(
...
-                    // Subscription died. Re-subscribe.
-                    subscription = relay_chain_sync
-                        .subscribe_all(32, NonZero::<usize>::new(usize::MAX).unwrap())
-                        .await;
-                    break header::hash_from_scale_encoded_header(
-                        &subscription.finalized_block_scale_encoded_header,
-                    );
+        let finalized_hash = if try_initial_finalized {
+            try_initial_finalized = false;
+            header::hash_from_scale_encoded_header(
```
lexnv: Could we provide a stale hash here? Then smoldot would go into a warp sync mode and, effectively soon after terminate the subscription?

light-base/src/sync_service/parachain.rs
replghost: Done — the log now only fires in `wait_for_finalized_hash()`, which is only called when the initial attempt (using the already-available finalized block) fails. The function was also restructured: `try_fetch_parachain_head()` attempts the fetch and returns `Option`, the main loop tries the initial finalized block first, and only falls through to `wait_for_finalized_hash()` on failure.

light-base/src/sync_service/parachain.rs
replghost: Interesting idea, but feeding intentionally incorrect data to trigger a side effect is fragile — it depends on internal warp sync behavior that could change. The simpler fix: just use the already-available `subscription.finalized_block_scale_encoded_header` as the first attempt, before waiting for new notifications. The resubscribe path (on subscription death) was already doing exactly this — the initial subscribe just wasn't.

The broader warm-start optimization (restoring chain information from DB, skipping bootstrap) has been split out to a follow-up PR.

replghost:

light-base/src/sync_service/parachain.rs
lrubasze: @lexnv, you were right to flag this — the staleness is real, though the failure mode turned out to be slow bootstrap rather than subscription termination. On cold start the initial-subscribe path can return the chain-spec checkpoint pre-warp, which on stale chains causes the parachain to bootstrap from a months-old head and chase a large backlog of finality, significantly extending startup time or timing out.

Fixed in the follow-up commit by gating the shortcut on a new wait_warp_sync_finished() signal — the initial finalized header is now always post-warp, so warm restart still gets the zero-wait path while cold start waits for warp sync to promote the head first.

lrubasze:
## Benchmark results

Paseo + Asset Hub Paseo, time-to-finalized, n=10 iterations.

| Mode | Baseline v3.1.1 (median / p95 / max) | PR #3210 (median / p95 / max) | Speedup (median / p95) |
|------|--------------------------------------|-------------------------------|------------------------|
| Cold | 16.5s / 50.0s / 70.0s                | 5.5s / 11.2s / 12.5s          | **3.0× / 4.5×**        |
| Warm | 17.3s / 43.5s / 50.3s                | 4.2s / 4.7s / 4.8s            | **4.1× / 9.3×**        |

Highlights:
- Warm-start variance collapses: stddev 13.4s → 0.7s. The hang-until-next-GrandPa-round behavior is gone — every run is now fast.
- Cold start is 3× faster median, 4.5× at p95, no runs near the 120s timeout (worst 12.5s vs baseline 70s).

benchmarked with: https://github.com/paritytech/smoldot/pull/3233

skunert (approved):

lib/src/sync/all.rs
```diff
@@ -280,6 +287,29 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
...
+    /// still report the chain-spec starting block as finalized. Harmless on the short
+    /// chains where the all-forks branch fires (handful of blocks of staleness at most);
+    /// the cold-start-with-stale-checkpoint case goes through `WarpSyncFinished` and
+    /// stays in lockstep.
+    pub fn is_warp_sync_finished(&self) -> bool {
+        if self.warp_sync.is_none() {
+            true
+        } else {
+            self.finalized_block_number() > self.shared.starting_block_number
+        }
```
bkchr, line 310: This is clearly a bug. Even if we did not use `warp_sync` for whatever reason, it should be set to `None`. We should not need this weird check here.

light-base/src/sync_service/parachain.rs
```diff
@@ -1113,32 +1118,59 @@ async fn fetch_parachain_head_from_relay<TPlat: PlatformRef>(
...
+    log!(
+        platform,
+        Debug,
+        log_target,
+        "Relay chain warp sync finished."
+    );
+
+    let mut subscription = relay_chain_sync
+        .subscribe_all(32, NonZero::<usize>::new(usize::MAX).unwrap())
+        .await;
```
bkchr: If we just ensure to send a finalized event after finishing warp sync, we would not need any of the changes here?

light-base/src/sync_service/paraheads.rs
```diff
@@ -930,6 +930,18 @@ impl<TPlat: PlatformRef> ParachainBackgroundTask<TPlat> {
...
 
+                (
+                    WakeUpReason::ForegroundMessage(ToBackground::WaitWarpSyncFinished {
+                        send_back,
+                    }),
+                    _,
+                ) => {
+                    // Paraheads doesn't run a warp-sync phase of its own; delegate to the relay's
+                    // sync service.
+                    self.relay_chain_sync.wait_warp_sync_finished().await;
```
bkchr: You are blocking here the entire sync service.

lib/src/sync/all.rs
skunert: This one we should not do for now, last time this caused a lot of problems .I propose to not touch warp_sync for now.

light-base/src/sync_service/parachain.rs
skunert: Yeah this sounds like a good idea.

lib/src/sync/all.rs
lrubasze: This is to address edge-cases for warm start or short local zombienet, where warp sync proofs don't arrive and warp sync machine never reaches `WarpSyncFinished`.
After recent struggles (eg. panics) after touching `warp_sync` I was not very keen to touch it again.

light-base/src/sync_service/paraheads.rs
gab8i: Why so? The function `sync_service::parachain::start_parachain` perform multiple steps, the first is to call `fetch_parachain_head_from_relay`, where the 'real' `relay_chain_sync.wait_warp_sync_finished` will happen, and only after it all the paraheads will be spawned, thus no real lock should happen because as long as the relay_chain task keep working it should immediately respond because the warp sync already finished.

The relay task doesn't seems to depend on any parachain/parahead logic, thus no deadlock seems to be possible.

This seems to be more of an adaptation of the new `ToBacground` variant which is only used by the relay chain task.

light-base/src/sync_service/parachain.rs
lrubasze: Yes, that would basically mean that `subscribe_all` itself ensures warp sync is finished before returning.
Let me work on this.

light-base/src/sync_service/paraheads.rs
lrubasze: Yes, this is a dead code, just to support new `ToBackground` variant. Paraheads is started after warp sync is already finished. I should have left a comment.
But I agree this is a code smell, so assuming above maybe we could omit this await call entirely:
```suggestion
                    // Paraheads is spawned only after the relay's warp sync has finished,
                    // so this is always already done.
```

light-base/src/sync_service/parachain.rs
lrubasze: runtime_service refactored, so `subscribe_all()` returns only after warp sync is finished.
Applied some changes here too, just to make sure that `subscription.finalized_block_scale_encoded_header` is used at first, so we don't need to wait for `Finalized`, which can take some time.

lrubasze: To shed more light on the refactor here.
`runtime_service` tracks every block received in an internal tree. To emit `runtime_service::Notification::Finalized` event for a block it looks up the block hash in that tree. The blocks are provided by `sync_service` via `sync_service::Notification::Finalized` event, which is not the case for warp synced blocks. If `sync_service` sent `Finalized` event for warped-synced block, then `runtime_service` wouldn't find such block in the tree and by design panic (lines 2613, 2621):
https://github.com/paritytech/smoldot/blob/19050210879dc7a9f10aa3236672aa7500ebea5f/light-base/src/runtime_service.rs#L2562-L2626

Instead of sending the event `sync_service` kills the subscription after warp sync is finished. Subscribers reconnect and get a fresh tree starting from the post-warp head (`subscription.finalized_block_scale_encoded_header`) and they receive finalized blocks as they are being synced.

Assumed that faking finalized event for warp sync block would be less clean than simply using `subscription.finalized_block_scale_encoded_header`, especially after we ensure that warp sync is finished before `subscribe_all()` returns.

bkchr: IMO this is still too complicated and I think something like [this](https://github.com/paritytech/smoldot/commit/9291582eaee8a9417f6777042e3ea2e92932d096) should be enough to achieve the same.

lrubasze: Refactored it in a suggested manner, by sending "synthetic" Finalized upon new subscription.

It was not possible to apply it sync_service via `sync_service::Notification::Finalized` via `light-base/src/sync_service/substrate_compat.rs` as suggested [here](https://github.com/paritytech/smoldot/commit/9291582eaee8a9417f6777042e3ea2e92932d096), because it caused panic as described [above](https://github.com/paritytech/smoldot/pull/3210#discussion_r3184365907).

bkchr: Can we not just announce the block and directly followed by teh finalized notification? This should be possible?

lrubasze: Check https://github.com/paritytech/smoldot/pull/3246

lrubasze:
Closing in favor of https://github.com/paritytech/smoldot/pull/3246

PR #3212, opened 2026-04-21
fix(database): stop storing runtime code blob in database

### LINKED ISSUE #3211: Warm start is slower than cold start due to base64-decoding runtime code in database
## Problem

Since #863 (July 2023), `encode_database` stores a base64-encoded copy of the ~2 MiB runtime Wasm blob in the database JSON string (`runtimeCode` field). On warm start, `decode_database` base64-decodes this blob.

In the Wasm (browser) environment, this base64 decode costs **~1.1 seconds per chain**. For a parachain setup (relay chain + parachain), the total overhead is **~2.3 seconds**, making warm start consistently slower than cold start.

## Evidence

Waterfall comparison (parachain on Polkadot):

| Metric | Cold start (no DB) | Warm start (with DB) |
|---|---|---|
| `chain-added` duration | 2329ms | 4618ms |

The ~2.3s difference is the base64 decode of two runtime blobs during `decode_database`.

## Why the decoded code is unused

On warm start, the decoded runtime code is never consumed:

- **Relay chains**: The database contains chain information, so warp sync is skipped entirely. The `code_trie_node_hint` passed to `AllSync` is never consumed because the warp sync phase doesn't run.
- **Parachains**: `ConfigParachain` has no `runtime_code_hint` field — the decoded code is dropped immediately.

The runtime code hint only helps during warp sync, which only runs on cold start (or when the database is too stale). On warm start — the exact scenario where the database is loaded — warp sync doesn't run.

## Reproduction

1. Start smoldot with a parachain chain spec (e.g. Polkadot Asset Hub)
2. Wait for sync, call `chain.databaseContent()` to save the database
3. Restart smoldot, passing the saved database as `databaseContent`
4. Measure `chain-added` timing and compare with a restart without `databaseContent`

Expected: warm start should be faster than or equal to cold start.
Actual: warm start is ~2x slower than cold start.

## Impact

Affects every smoldot user who persists and restores the database via `databaseContent` — including polkadot.js apps using localStorage, and any application using the `databaseContent` API.

## Fix

Remove the `runtimeCode` field from the database. Keep the small merkle metadata (`codeMerkleValue`, `codeClosestAncestor`) for potential future use. Database size drops from ~5 MB to ~3 KB per chain.

Old databases containing a `runtimeCode` field still parse correctly (the field is ignored).

## Future: binary code cache for slow networks

For users on slow connections (< 5 Mbps), caching the runtime code avoids a costly download on each startup. This should be implemented as a **separate binary cache** (`Uint8Array`) rather than a base64 blob inside JSON. See discussion in the linked PR.

replghost:
## Summary

Fixes #3211. The database JSON stores a base64-encoded ~2 MiB runtime Wasm blob that is decoded on every warm start but never used. This makes warm start slower than cold start.

- Stop writing \`runtimeCode\` to the database (set \`code_storage_value: None\`)
- Stop base64-decoding \`runtimeCode\` on load (match on merkle value + ancestor only)
- Remove \`code\` field from \`DatabaseContentRuntimeCodeHint\`
- Remove \`ConfigSubstrateCompatibleRuntimeCodeHint\` and plumbing
- Set \`code_trie_node_hint: None\` in warp sync config
- Old databases with \`runtimeCode\` still parse (backwards compatible, field ignored)

## Why the runtime code was unused

On warm start, warp sync is skipped (chain info from DB). The \`code_trie_node_hint\` is only consumed during warp sync. For parachains, the hint was never wired into \`ConfigParachain\` at all.

### Database size (fix vs main)

Same chain (Paseo), same sync state, different builds:

| Build | DB size | Has runtimeCode |
|-------|---------|-----------------|
| main | 2342 KB | yes |
| fix | 44 KB | no |

### Decode cost (Rust unit test, native)

\`\`\`
with 2 MiB code:    0.4ms/iter (JSON size: 2,666,889 bytes)
without code:       0.0004ms/iter (JSON size: 205 bytes)
ratio: 812x slower with code
\`\`\`

### addChain A/B (Node.js, same binary, 5 runs, median)

| DB variant | addChain | 
|------------|----------|
| main DB (2342 KB, with runtimeCode) | 11.2ms |
| fix DB (44 KB, no runtimeCode) | 7.6ms |

Node.js runs JSON.parse and base64 as native C++. In browser Wasm these are ~1000x slower, consistent with the user's observed 2.3s regression.

### Browser Wasm (smolbench, Playwright, Paseo)

| Scenario | init (ms) | DB loaded |
|----------|-----------|-----------|
| main cold | 2607 | none |
| main warm | 6850 | 2341 KB (with runtimeCode) |

After the fix, new databases are 44 KB. After one restart cycle, warm start loads a 44 KB JSON instead of 2.3 MB.

## Test plan

\`\`\`bash
cargo test -p smoldot-light -- database::tests --nocapture
\`\`\`

- \`decode_db_without_code_is_fast\` — 100 decodes in <100ms
- \`decode_db_with_legacy_code_field_still_works\` — old 2 MB databases parse correctly
- \`new_encode_omits_runtime_code\` — new databases omit runtimeCode
- \`decode_with_code_much_slower_than_without\` — 812x slowdown measured
- \`database_size_comparison\` — 2.7 MB vs 205 bytes


replghost:
## Benchmark results (Node.js, 3 iterations, relay chains only)

Note: these run in **Node.js** where JSON parse and base64 run as native C++. In the browser **Wasm** environment (where smoldot actually runs), base64 decode is ~1000x slower. The user's waterfall showed 4618ms warm vs 2329ms cold — a 2.3s gap.

### Warm-cold delta (what matters)

Measured by loading a 2.4-2.7 MB database containing the \`runtimeCode\` field:

| Network | main (warm-cold) | fix (warm-cold) | Improvement |
|---------|-----------------|-----------------|-------------|
| Polkadot | +9.0ms | +5.1ms | 43% less overhead |
| Kusama | +9.5ms | +6.8ms | 28% less overhead |
| Westend | +4.5ms | +2.2ms | 51% less overhead |

The fix reduces the warm start overhead because the 2.7 MB \`runtimeCode\` base64 string is no longer decoded. The remaining overhead is JSON parsing the large string (which is still present in old databases for backwards compatibility).

### Database size after fix

New databases produced by the fix branch will **not** contain \`runtimeCode\`:

| | main | fix |
|---|---|---|
| Polkadot DB | 2451 KB | ~170 KB |
| Kusama DB | 2707 KB | ~170 KB |
| Westend DB | 2488 KB | ~170 KB |

After one save-restore cycle on the fix branch, the database shrinks to ~170 KB and the warm-cold delta drops to near zero.

### Extrapolated browser impact

The Rust unit test shows 812x decode slowdown for 2 MiB base64 in native. In Wasm, JSON.parse + base64 of a 2.7 MB string takes ~1.1s per chain (measured in the user's waterfall). With a ~170 KB database (no runtimeCode), the parse is <1ms.

replghost:
## Proper A/B benchmark: same binary, DB with vs without runtimeCode

Same Wasm binary for all tests. Only variable is database content. 5 runs each, median reported.

### Results (Node.js, per relay chain)

| Network | No DB | With runtimeCode | Without runtimeCode | Overhead | DB size reduction |
|---------|-------|------------------|---------------------|----------|-------------------|
| Polkadot | 15.3ms | 20.3ms | 14.0ms | **+6.4ms (+46%)** | 2483 KB → 168 KB |
| Kusama | 17.3ms | 18.4ms | 17.0ms | **+1.5ms (+9%)** | 2707 KB → 276 KB |
| Paseo | 14.4ms | 14.1ms | 10.2ms | **+3.8ms (+38%)** | 2342 KB → 44 KB |

Key observations:
- Without runtimeCode, warm start is **equal to or faster than** cold start (chain info + known nodes help)
- With runtimeCode, warm start is consistently slower
- DB size drops 90-98%

### Why the overhead looks small here

These numbers are from **Node.js** where V8 handles JSON.parse and base64 as native C++ operations. The overhead is single-digit ms.

In the **browser Wasm** environment (smoldot's actual deployment target), JSON.parse and base64 decode run inside the Wasm VM. The user's waterfall measurements show:
- Cold \`chain-added\`: **2329ms**
- Warm \`chain-added\`: **4618ms**
- Delta: **+2289ms** (for two chains)

This is consistent with the ~1000x Wasm overhead factor: 6ms native × 1000 ≈ 6s for two chains.

### Reproducing

\`\`\`bash
# Build the Wasm:
cd wasm-node/javascript && npm run build

# Run the benchmark (uses fix branch binary):
node /tmp/bench-db-decode.mjs ./dist/mjs/index-nodejs.js
\`\`\`

The script syncs each chain to get a database with \`runtimeCode\`, then compares \`addChain\` timing with vs without the runtimeCode field (same chain info, same nodes).

replghost:
## Browser Wasm benchmark (Playwright, Chromium, Paseo relay chain)

Using [smolbench](https://github.com/paritytech/smolbench) browser benchmark infrastructure — real Chromium, real Wasm, real network.

### Main (before fix) — relay cold vs warm

| Scenario | init (ms) | chain-added → first proof | DB size |
|----------|-----------|---------------------------|---------|
| Cold | 2607 | 1635ms (warp sync) | — |
| Warm | 6850 | 4041ms (**4s gap**) | 2341 KB |

### Fix (after fix) — relay cold vs warm

The fix stops **writing** \`runtimeCode\` to the database. New databases are 44 KB instead of 2341 KB.

| DB variant | addChain median (Node.js) | DB size |
|------------|--------------------------|---------|
| Main DB (with runtimeCode) | 11.2ms | 2342 KB |
| Fix DB (no runtimeCode) | 7.6ms | 44 KB |

### What the fix changes

1. **New databases** produced after the fix: 44 KB (no \`runtimeCode\`). Warm start loads instantly.
2. **Old databases** from before the fix: still parse correctly (backwards compatible), but the 2341 KB JSON must still be parsed. The \`runtimeCode\` field is ignored during decode (no base64 decode).
3. **After one cycle**: user restarts → loads old DB → runs → saves new DB → next restart loads the small 44 KB DB.

### What we still can't prove in this PR

The 4s gap between \`chain-added\` and first storage proof in the warm timeline is **network latency** (waiting for a peer + proof response), not database decode. The decode overhead is real but small compared to network variance in single runs. A proper measurement needs many runs isolating just the decode time in Wasm — which the Rust unit test does (812x slowdown, extrapolating to ~1s in Wasm).

The strongest evidence remains:
- Rust unit test: 812x decode slowdown with 2 MiB code blob
- Node.js A/B test: 44 KB DB consistently ~3-4ms faster than 2342 KB DB
- User's original waterfall: 4618ms warm vs 2329ms cold

replghost:
Closing — the browser Wasm benchmarks (Polkadot, Kusama, Paseo) showed only 9-15ms overhead from the runtimeCode blob in the database. Not a meaningful performance issue on main. The original 2.3s regression was from the parachain hint probe logic on the fix/parachain-warm-start-hint branch, which was never merged.

PR #3213, opened 2026-04-21
perf(runtime-service): use short retry when no peers available

replghost:
## Summary

On warm start, the runtime service immediately tries to download the finalized block's runtime. No peers are connected yet. The download fails with `StorageQueryError { errors: [] }` (no peers queried), triggering a 4s generic cooldown. Every warm start wastes 4 seconds.

## Fix

Distinguish "no peers available" from "peers returned bad data":
- `StorageQueryError::is_no_peers()` — empty error list means nobody was queried
- On `is_no_peers()`: exponential backoff (200ms → 400ms → 800ms), then fall through to the normal 4s cooldown after 3 fast retries
- All other errors: full 4s cooldown (unchanged)
- Counter resets on successful download

Adds `async_op_failure_retry_at()` to `AsyncTree` for custom retry timing.

### Relay chain warm start (5 runs each)

| Network | Before (median) | After (median) |
|---------|-----------------|----------------|
| **Polkadot** | ~5-7s | **673ms** |
| **Kusama** | ~5-7s | **628ms** |
| **Paseo** | ~5-7s | **1405ms** |

### Parachain end-to-end (cold → save DB → warm)

| Network | Cold | Warm (with fix) |
|---------|------|-----------------|
| **Polkadot Asset Hub** | 13.7s | **7.0s** |
| **Paseo Asset Hub** | 19.2s | **7.7s** |

## Reproduction

```bash
cd smolbench
SMOLDOT_DIST=path/to/dist/mjs/index-nodejs.js \
  CHAIN_SPEC=chain-specs/polkadot.json WARM=1 RUNS=5 \
  node bench/runtime-download-count.mjs
```

## Test plan

- [x] `cargo check -p smoldot -p smoldot-light` clean
- [x] `cargo fmt` clean
- [x] Relay chain benchmarked on 3 networks, 5 runs each
- [x] Parachain end-to-end benchmarked on 3 networks
- [x] Exponential backoff prevents busy loop when genuinely offline
- [x] After 3 fast retries, falls through to normal 4s cooldown

### DIFF
--- a/light-base/src/runtime_service.rs
+++ b/light-base/src/runtime_service.rs
@@ -831,6 +831,7 @@ async fn run_background<TPlat: PlatformRef>(
             blocks_stream: None,
             runtime_downloads: stream::FuturesUnordered::new(),
             progress_runtime_call_requests: stream::FuturesUnordered::new(),
+            no_peers_retry_count: 0,
         }
     };
 
@@ -2766,6 +2767,7 @@ async fn run_background<TPlat: PlatformRef>(
                 };
 
                 // Insert the runtime into the tree.
+                background.no_peers_retry_count = 0;
                 match &mut background.tree {
                     Tree::FinalizedBlockRuntimeKnown { tree, .. } => {
                         tree.async_op_finished(async_op_id, runtime);
@@ -2810,16 +2812,16 @@ async fn run_background<TPlat: PlatformRef>(
                     );
                 }
 
-                if error.is_no_peers() {
-                    // No peers available yet — use a short retry (200ms) instead of
-                    // the full 4s cooldown. Peers typically connect within milliseconds.
-                    let short_retry = background.platform.now() + Duration::from_millis(200);
+                if error.is_no_peers() && background.no_peers_retry_count < 3 {
+                    let delay_ms = 200u64 << background.no_peers_retry_count;
+                    background.no_peers_retry_count += 1;
+                    let retry_at = background.platform.now() + Duration::from_millis(delay_ms);
                     match &mut background.tree {
                         Tree::FinalizedBlockRuntimeKnown { tree, .. } => {
-                            tree.async_op_failure_retry_at(async_op_id, &short_retry);
+                            tree.async_op_failure_retry_at(async_op_id, &retry_at);
                         }
                         Tree::FinalizedBlockRuntimeUnknown { tree, .. } => {
-                            tree.async_op_failure_retry_at(async_op_id, &short_retry);
+                            tree.async_op_failure_retry_at(async_op_id, &retry_at);
                         }
                     }
                 } else {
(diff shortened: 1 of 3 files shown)


light-base/src/runtime_service.rs
```diff
@@ -2810,12 +2810,26 @@ async fn run_background<TPlat: PlatformRef>(
                     );
                 }
 
-                match &mut background.tree {
-                    Tree::FinalizedBlockRuntimeKnown { tree, .. } => {
-                        tree.async_op_failure(async_op_id, &background.platform.now());
+                if error.is_no_peers() {
+                    // No peers available yet — use a short retry (200ms) instead of
+                    // the full 4s cooldown. Peers typically connect within milliseconds.
```
lexnv: Is this true? I believe smoldot is not driving kademlia discovery under the hood.

light-base/src/runtime_service.rs
```diff
@@ -2810,12 +2810,26 @@ async fn run_background<TPlat: PlatformRef>(
...
+                    let short_retry = background.platform.now() + Duration::from_millis(200);
```
lexnv: This effectively turns the 4s cooldown into a busy loop every 200ms if the peers are genuinely not reachable. Could probably consider an exponential backoff

light-base/src/sync_service.rs
```diff
@@ -1040,6 +1040,11 @@ pub struct StorageQueryError {
 }
 
 impl StorageQueryError {
+    /// Returns `true` if no peers were available to query.
+    pub fn is_no_peers(&self) -> bool {
```
lexnv, line 1044: Do we have a "no peers" path for parahead fetch?

light-base/src/runtime_service.rs
replghost: You're right — smoldot doesn't drive Kademlia. Removed the comment entirely. The backoff schedule (200ms → 400ms → 800ms → 4s) speaks for itself.

light-base/src/runtime_service.rs
replghost: Implemented. Exponential backoff: 200ms → 400ms → 800ms, then fall through to the normal 4s cooldown after 3 fast retries. Counter (`no_peers_retry_count`) lives in the background task, resets on successful download. See 66866ae.

light-base/src/sync_service.rs
replghost: No, and it doesn't need one. `fetch_parachain_head_from_relay` (after #3210) waits on relay chain `subscribe_all` notifications, not directly on peers. The relay chain runtime service handles peer connectivity — the parachain fetch blocks on relay chain events, so there's no peer-level retry to optimize there.

replghost:

PR #3209, opened 2026-04-17
feat(sync): add onSyncProgress callback for warp-sync progress

BigTava:
Closes #3206

PR #3215, opened 2026-04-22
add sync rpc

KarimJedda:


light-base/src/json_rpc_service/background.rs
```diff
@@ -2834,6 +2835,46 @@ pub(super) async fn run<TPlat: PlatformRef>(
                         }
                     }
 
+                    methods::MethodCall::sudo_unstable_syncStatus {} => {
```
lexnv, line 2838: Would it make more sense to have this as a subscription similar to `chainHead`? Then developers would have the information accessible instantly via events, as opposed to manually calling this method every now and then, which can create some blindspots?

lexnv:
Would love to hear what your use cases are for the sync RPC🙏 

Is this intended to be delivered to end-users? Is this something needed to work around another issue? Does this modify/change your normal API behavior of using smoldot? 

We'd also need a counterpart PR in https://github.com/paritytech/json-rpc-interface-spec. This is our spec for the RPC methods shared between substrate and smoldot. It would probably be better to create the draft there, since other external contributors might take part in the design decision (for example @josepot) 🙏

josepot:
I don't think this endpoint exposes an API that makes sense in the context of a light-client.

For a full-node, this API is definitely useful because it can indicate how close the node is to completing sync. A full node must import every block up to the chain tip in order to be considered synced, so progress can be expressed in a fairly natural way.

A light client, however, uses warp-sync, which works very differently. Because of that, I don’t see how the API proposed here maps cleanly to the light-client model.

I assume the intention is to let consumers track warp-sync progress. But that is tricky, especially if we want to avoid exposing implementation details of a particular light-client design.

For example, one valid implementation might:

- first import all headers where the authority-set changed, and then
- after reaching the last authority-set change, import the remaining headers up to the tip.

Another equally valid implementation could perform both steps in parallel. And beyond that, there are many other aspects of a light client’s warp-sync process that can affect perceived progress (provided snapshot of light-sync-state that may or may not be ignored, internal-db, etc).

So if the goal of this RPC is to report warp-sync progress, I think the API should avoid exposing low-level implementation details. Instead, it should allow each light client to provide its own estimated progress in a way that reflects its particular warp-sync strategy. In other words, probably the best would be to expose an aprox percentage of progress and let each implementer determine how to come up with that number.

That said, I personally haven’t run into a strong need for warp-sync progress reporting. That does not mean the use case does not exist. It is perfectly reasonable to want some indication of how much of the warp sync has completed. My main concern is simply that the API proposed in this RPC does not seem like the right abstraction for a light client.

bkchr:
This pull request will not be used in this way. We want to show some progress of warp sync. I written down on what we need for this here https://github.com/paritytech/smoldot/pull/3209#discussion_r3122083357.

I think this pr is not really bringing us anything and I'm closing it.

PR #3224, opened 2026-04-22
Add releasing guide

lrubasze:
## Summary

1. **`docs/RELEASING.md`** — new release runbook: 12 steps + two appendices,
   covering version bumps, branch/PR/merge flow, tag creation, and post-deploy
   sanity checks. 
   **Appendix A documents the manual dev-publish flow** (how to
   trigger the workflow, what version/dist-tag get produced, how to install a
   dev build).

3. **`docs/DEV-PUBLISHING.md`** — short companion guide for the manual
   dev-publish flow (when you want a testable npm build from a feature branch
   without editing any version files by hand).

2. **`.github/workflows/deploy.yml`** — harden the `workflow_dispatch`
   dev-publish path so it can't move the `latest` dist-tag and no longer
   needs a manual `package.json` version commit.

## Dev-publish hardening

Replaces `npm_tag` input with optional `npm_tag_suffix` (charset
`[A-Za-z0-9-]`). On `workflow_dispatch` the workflow now:

- Computes `TAG = dev-<YYYYMMDD>[-<suffix>]` — cannot equal `latest`.
- Computes `VERSION = <next-patch>-<TAG>.<N>` — so the dist-tag string
  appears verbatim inside the version. `N` is `max + 1` from
  `npm view --json smoldot versions`; first dispatch of a new
  `(next-patch, date, suffix)` tuple starts at `.0`. Every subsequent
  dispatch (including "Re-run jobs") bumps `N`.
- Rewrites `wasm-node/javascript/package.json` and `wasm-node/rust/Cargo.toml`
  in-memory before `npm pack`. Never committed.

Examples (dispatched 2026-04-22, current stable = 3.1.1):

| suffix     | VERSION                         | TAG                    |
|------------|---------------------------------|------------------------|
| `smoke`    | `3.1.2-dev-20260422-smoke.0`    | `dev-20260422-smoke`   |
| *(blank)*  | `3.1.2-dev-20260422.0`          | `dev-20260422`         |

Install the first one with:
`npm install smoldot@dev-20260422-smoke` (follows the tag) or
`npm install smoldot@3.1.2-dev-20260422-smoke.0` (pinned).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,7 +26,7 @@ on:
   workflow_dispatch:
     inputs:
       npm_tag_suffix:
-        description: 'Optional suffix for the dev dist-tag. Non-empty → tag "dev-<YYYYMMDD>-<suffix>", version "<next-patch>-dev.<YYYYMMDD>.<suffix>.<N>". Empty → tag "dev-<YYYYMMDD>", version "<next-patch>-dev.<YYYYMMDD>.<N>". Version always bumps patch (never minor/major). Cannot produce "latest". Allowed suffix chars: [A-Za-z0-9-].'
+        description: 'Optional suffix for the dev dist-tag. Non-empty → tag "dev-<YYYYMMDD>-<suffix>", version "<next-patch>-dev-<YYYYMMDD>-<suffix>.<N>". Empty → tag "dev-<YYYYMMDD>", version "<next-patch>-dev-<YYYYMMDD>.<N>". Version always bumps patch (never minor/major). Cannot produce "latest". Allowed suffix chars: [A-Za-z0-9-].'
         type: string
         required: false
         default: ''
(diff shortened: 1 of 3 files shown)


lrubasze:
Not sure, maybe the releasing doc is too verbose?

docs/RELEASING.md
```diff
@@ -0,0 +1,315 @@
...
+git tag -l '*<X.Y.Z>*'   # should list the CI-pushed light-js-deno-v tag plus the manual ones
+```
+
+On any post-publish failure, do not delete published versions — yank
+(`npm deprecate smoldot@<X.Y.Z> '...'`, `cargo yank --version <A.B.C>`) and
+cut a new release.
+
+---
+
+## Appendix A — Manual dev publish
```
skunert, line 262: I think we can keep this doc, but its a bit lengthy. We should have a small markdown just for the dev release which is super short TLDR. Like create branch, press run workflow, add this and that tag. If someone comes and just wants to do a dev release they should not dig through all the details here.

.github/workflows/deploy.yml
```diff
@@ -193,6 +194,57 @@ jobs:
           node-version: 18
       - uses: Swatinem/rust-cache@v2
 
+      # On workflow_dispatch: compute a unique dev version and a safe dist-tag,
+      # then rewrite them in-place into package.json and wasm-node/rust/Cargo.toml
+      # (never committed). On push-to-main this step is skipped; the committed
+      # versions are published to the default dist-tag (`latest`).
+      - name: Compute dev version and dist-tag
+        if: github.event_name == 'workflow_dispatch'
```
skunert, line 202: Nit: I think we should add shell: bash here to get pipefail for the pipelines below: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell

skunert (approved):

.github/workflows/deploy.yml
lrubasze: Done

lexnv (approved):

PR #3230, opened 2026-04-24
fix continous re-trying on statement store

alexggh:
... when we enable statement store if we try to connect to a peer that is not supporting the protocol, current behaviour is to continously retry which end creating a lot of events that consume unbounded memory.


skunert (approved):

lib/src/network/service.rs
```diff
@@ -2432,20 +2432,48 @@ where
...
                             );
 
-                            // If the substream failed to open, we simply try again.
-                            // Trying again means that we might be hammering the remote with
-                            // substream requests, however as of the writing of this text this is
-                            // necessary in order to bypass an issue in Substrate.
-                            // Note that in the situation where the connection is shutting down,
-                            // we don't re-open the substream on a different connection, but
-                            // that's ok as the block announces substream should be closed soon.
+                            // If the substream failed to open, we may try again. Re-trying
```
AndreiEres: Do we really need this long "Previously in smoldot..." comment?

AndreiEres (approved):

lib/src/network/service.rs
alexggh: Fixed.

PR #3233, opened 2026-04-27
Smoldot startup benchmarks

lrubasze:
Adds a startup benchmark for smoldot.

- Measures time-to-first-finalized-block, as seen by an app on top of smoldot.
- Two modes: **cold** (no DB, warp-sync) and **warm** (pre-saved DB skips warp-sync).
- Two targets: **relay** or **para**.
- Runs against a local zombienet network, or any real network via a chain spec.
- Reports per-phase timings and finalized-block drift; JSON output optional.

Example:
```bash
cargo run --release --bin startup -- \
   --mode cold  --target para  --relay-chain-spec paseo  --para-chain-spec paseo_asset_hub  --iterations 5
```
Output:
```
=== startup benchmark ===
mode                : Cold
source              : user-supplied spec
relay chain         : Paseo Testnet (paseo)
para chain          : Paseo Asset Hub (asset-hub-paseo)
target              : Para
iterations          : 5
warmup              : 0

finalized_ms (total)         mean=5229.0 median=5064.7 p95=8008.4 stddev=1752.5 min=2776.0 max=9790.7
  addChain relay             mean=43.9 median=43.3 p95=47.2 stddev=2.0 min=41.7 max=48.9
  addChain para              mean=13.0 median=13.4 p95=13.8 stddev=0.9 min=11.3 max=13.9
  wait finalized block       mean=5142.8 median=4981.0 p95=7921.9 stddev=1752.6 min=2690.9 max=9703.4
```

See `benchmarks/README.md` for more details and examples.


benchmarks/js/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "type": "module",
+  "private": true,
+  "dependencies": {
+    "polkadot-api": "^2.1.0",
```
lrubasze, line 5: Not sure if we want PAPI as dependency eventually. 
Using it like this right now to have setup as close as possible as dotli has.

benchmarks/README.md
```diff
@@ -0,0 +1,109 @@
...
+  --relay-chain-spec polkadot \
+  --para-chain-spec polkadot_asset_hub \
+  --iterations 5
+
+# warm, Polkadot relay
+cargo run --release --bin startup -- \
+  --mode warm --target relay --relay-chain-spec polkadot --iterations 5
+```
+
+### Flags
```
skunert: IMO we should not have this, will just get outdated. We can expect people to call --help

benchmarks/README.md
```diff
@@ -0,0 +1,109 @@
...
+cd benchmarks
+
+# cold, westend-local relay
+ZOMBIE_PROVIDER=native cargo run --release --bin startup -- --target relay --iterations 10
+
+# cold, people-westend-local parachain
+ZOMBIE_PROVIDER=native cargo run --release --bin startup -- --target para --iterations 10
+
+# warm, parachain
+ZOMBIE_PROVIDER=native cargo run --release --bin startup -- --mode warm --target para --iterations 10
```
skunert, line 45: I tried this one and got a panic

```
Thread 'main' panicked at 'Invalid value: "Value 'people-westend-1004' is not configured yet"', cumulus/polkadot-parachain/src/chain_spec/mod.rs:259
```

skunert (commented):
In general looking good. With real networks numbers are all over the place of course because they depend on which peer, network connection, when next block arrives etc.

In the future we could also look into how long the individual stages, would be interesting to know time to first peer, warp sync time spent, time from para finalized head to tip etc.

benchmarks/README.md
lrubasze: Probably old `polkadot-parachain` version, which tries to resolve legacy runtime using chain id.
```
  polkadot-parachain --chain /tmp/.../people-westend-custom.json
    └─> load_spec(<absolute path>)            // path-fallback, no panic
    └─> ChainSpec loaded, id = "people-westend-1004"
    └─> RuntimeResolver::runtime(spec)
          └─> LegacyRuntime::from_id("people-westend-1004")
                └─> "people-westend-1004".parse::<PeopleRuntimeType>() = Err
                    └─> .expect("Invalid value")  → PANIC
```
Recent versions don't do it.
Anyway I will change chain id to `people-westend-local`.

skunert (approved):

lrubasze:
> In general looking good. With real networks numbers are all over the place of course because they depend on which peer, network connection, when next block arrives etc.
> 
> In the future we could also look into how long the individual stages, would be interesting to know time to first peer, warp sync time spent, time from para finalized head to tip etc.

Good idea. TODO added

gab8i (approved):
Implementation looking good! Also really useful to compare different transport layer such as tcp, wss or webrtc! 

Another future feature could be to spawn smoldot within a headless browser instead of node and have numbers as close as possible to reality!

PR #3234, opened 2026-04-27
Add smoke zombienet test

AndreiEres:
- Adds an e2e smoke test that boots zombienet and asserts smoldot observes new parachain blocks via chainHead_v1_follow
- Fixes block production in statement-store tests

### DIFF
--- a/e2e-tests/tests/smoke.rs
+++ b/e2e-tests/tests/smoke.rs
@@ -1,0 +1,102 @@
+// Smoldot
+// Copyright (C) 2019-2026  Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


e2e-tests/tests/smoke.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    let spawn_fn = zombienet_sdk::environment::get_spawn_fn();
+    let network = spawn_fn(config).await?;
+    network.detach().await;
+    network.wait_until_is_up(120).await?;
+
+    // The parachain producing a block implies the relay is running too.
+    network
+        .get_node("alice")?
+        .wait_metric_with_timeout("block_height{status=\"best\"}", |h| h >= 1.0, 300u64)
```
skunert: We should be a bit more generous here and wait for a few blocks.

skunert (approved):
nice

lrubasze (approved):
LGTM!

I think worth adding smoke test to CI?

PR #3240, opened 2026-04-30
tests: Add zombienet test for bitswap_v1_get

### LINKED ISSUE #3232: Zombienet for `bitswap_v1_get`
We need an integration test running in CI. Specifically:
- A helper test that generates DB snapshots for the bulletin chain and relay chain at a sufficient height (> 1000 blocks). This test should not run in CI; it should only be invoked manually.
- Snapshots shall be stored on our Google Drive.
- During DB snapshot generation, a set of well-known CIDs must be injected into the bulletin chain.
- The main test spawns the relay chain + parachain, then smoldot, and fetches some of the known CIDs via bitswap.
- Additional cases should also be covered, for example:
  - Fetching non-existent CIDs.
  - CIDs with mixed availability across bulletin nodes (which can happen during syncing).

BigTava:
Closes #3232


BigTava:
@skunert you mention snapshots should live on Google Drive, but I haven't seen anything similar. polkadot-bulletin-chain stores its build outputs as GitHub artifacts (e.g. [release.yml#L98-L103](https://github.com/paritytech/polkadot-bulletin-chain/blob/main/.github/workflows/release.yml#L98-L103)). Should we do that instead?

skunert:
@BigTava An example can be seen here: https://github.com/paritytech/polkadot-sdk/blob/3501f2c3c8e46d0b726994065a05615995f03880/cumulus/zombienet/zombienet-sdk/tests/zombie_ci/full_node_warp_sync/common.rs#L26

I think artifacts can be of limited size only and have a retention period. These snapshots we should only generate very rarely and then always reuse.

skunert:
Hmm looks like I lost access to our cloud storage, @pepoviola what is the current way to store these snapshots, do you still have access?

pepoviola:
> Hmm looks like I lost access to our cloud storage, @pepoviola what is the current way to store these snapshots, do you still have access?

You should have access to `zombienet-db-snaps` bucket, let me know if doesn't works for you and I can re-check.
Thx!

BigTava:
> An example can be seen here

Thanks @skunert! I added the remaining code based on this example. The snapshots come out to ~50 MB total. Someone with bucket-write access needs to upload them for the CI job to pass.

e2e-tests/js/bulletin_fetch.js
```diff
@@ -0,0 +1,204 @@
...
+      code === ERR_INVALID_PARAMS,
+      code === ERR_INVALID_PARAMS
+        ? `code ${code}`
+        : `expected ${ERR_INVALID_PARAMS}, got ${code}`,
+    );
+  }
+
+  try {
+    // Given
+    const fullOnly = payloads.find((p) => !p.on_partial);
```
lrubasze: Why not test all fullOnly payloads?

e2e-tests/src/bulletin.rs
```diff
@@ -0,0 +1,195 @@
...
+/// `schema_version` is a breaking change.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BulletinManifest {
+    pub schema_version: u32,
+    pub snapshot_height: u64,
+    pub bulletin_release_tag: String,
+    pub polkadot_release_tag: String,
+    pub payloads: Vec<ManifestPayload>,
+    pub archives: ArchiveChecksums,
+}
```
lrubasze: newline missing.
`cargo fmt` not applied?

e2e-tests/tests/bulletin_fetch.rs
```diff
@@ -0,0 +1,188 @@
...
+use std::path::{Path, PathBuf};
+
+use anyhow::{Result, anyhow};
+use serde::Serialize;
+use smoldot_e2e_tests::{
+    bulletin, ensure_js_deps_installed, ensure_smoldot_built, resolve_base_dir, run_js_test,
+};
+use zombienet_sdk::{LocalFileSystem, Network, NetworkConfigBuilder};
+
+const RELAY_CHAIN: &str = "westend-local";
```
lrubasze: Duplicated constants

https://github.com/paritytech/smoldot/pull/3240/changes#diff-e1ca510123b5e139ebb68f3f1d3f49a63fd44618ae70f45d8fe3500cbcba9d1fR41-R43

How about moving them to the `bulletin.rs`

e2e-tests/js/bulletin_fetch.js
```diff
@@ -0,0 +1,204 @@
...
+
+const client = createSmoldotClient();
+let exitCode = 0;
+try {
+  const relay = await addChainFromSpec(client, relaySpecPath);
+  const bulletin = await addChainFromSpec(client, bulletinSpecPath, {
+    potentialRelayChains: [relay],
+  });
+
+  for (const payload of payloads) {
```
lrubasze, line 51: Nit... How about separating each test into a function, eg.
```js
async function testKnownPayloads(chain, payloads) { ... }
```

e2e-tests/js/bulletin_fetch.js
```diff
@@ -0,0 +1,204 @@
...
+      report(`known-${payload.label}`, false, err.message);
+    }
+  }
+
+  try {
+    // Given
+    const cid = missingCid;
+
+    // When
+    const hex = await sendRpcAndWait(bulletin, "bitswap_v1_get", [cid], 60_000);
```
lrubasze: Sometimes we use `bitswapGetWithRetry` and sometimes `sendRpcAndWait`.
What's the logic behind?

lrubasze (commented):
Nice work!
Few comments from my side

e2e-tests/js/bulletin_fetch.js
BigTava: Thanks!

e2e-tests/src/bulletin.rs
BigTava: Yes. Thanks!

e2e-tests/tests/bulletin_fetch.rs
BigTava: Thank you!

e2e-tests/js/bulletin_fetch.js
BigTava: I left them inline to match statement_store_submission.js

e2e-tests/js/bulletin_fetch.js
BigTava: Thanks, replaced with `bitswapGetWithRetry` for consistency. At the moment of `sendRpcAndWait` smoldot is already warm, so there was no need for retries.

michalkucharczyk:
Dumb question: does `e2e-tests/tests/bulletin_generate_snapshot.rs` really belongs to smoldot repo domain?

I can imagine such snapshot can be used for bulletin related functionality also in other repos.
And one more thought - should snapshot generation be part of zombienet sdk? Again I can imagine this code will be duplicated in every test that requires snapshot (cc: @pepoviola).

@lrubasze @skunert - does it feel right for you?
@karolk91 wdyt?

lrubasze:
> Dumb question: does `e2e-tests/tests/bulletin_generate_snapshot.rs` really belongs to smoldot repo domain?
> 
> I can imagine such snapshot can be used for bulletin related functionality also in other repos. And one more thought - should snapshot generation be part of zombienet sdk? Again I can imagine this code will be duplicated in every test that requires snapshot (cc: @pepoviola).
> 
> @lrubasze @skunert - does it feel right for you? @karolk91 wdyt?

I think this is fine.
We use similar approach in [full_node_warp_sync test](https://github.com/paritytech/polkadot-sdk/blob/3501f2c3c8e46d0b726994065a05615995f03880/cumulus/zombienet/zombienet-sdk/tests/zombie_ci/full_node_warp_sync/generate_snapshots.rs)

Tests that use snapshot and test that generates it need to have the same network setup. So i think it is good to keep them in one place. Unless we outsource zombienet network setup too.

lrubasze (approved):

skunert:
> should snapshot generation be part of zombienet sdk?

I thought about this in the past too, my 2c is that its okay to include this per-repo, because sometimes the snapshots need to replicate specific scenarios. For bulletin I can imagine scenarios with specific counts of renew/store tx etc.

.github/workflows/zombienet.yml
```diff
@@ -68,6 +68,9 @@ jobs:
           - job-name: "zombienet-smoldot-0003-statement_store_peer_connection"
             test: "statement_store_peer_connection"
             runner-type: "default"
+          - job-name: "zombienet-smoldot-0004-bulletin_fetch"
```
skunert: Lets make this 007 so that it does not conflict with https://github.com/paritytech/smoldot/pull/3242

BigTava:
@lrubasze @skunert needs 2 approvals. Thanks

lrubasze (approved):

e2e-tests/js/bulletin_fetch.js
```diff
@@ -0,0 +1,200 @@
...
+    report(
+      "missing-invalid-cid",
+      code === ERR_INVALID_PARAMS,
+      code === ERR_INVALID_PARAMS
+        ? `code ${code}`
+        : `expected ${ERR_INVALID_PARAMS}, got ${code}`,
+    );
+  }
+
+  for (const payload of payloads.filter((p) => !p.on_partial)) {
```
skunert, line 119: One nit (don't think we need to change it right now). Would be nice to know if we actually hit the partial case where a CID is unavailable. Imagine if peer switching was broken and smoldot would always ask the same peer, then this test would become flaky depending on which peer is chosen for request right? Not sure we have a good way to handle this case currently.

skunert (approved):
Nice!

.github/workflows/zombienet.yml
BigTava: Thanks!

e2e-tests/js/bulletin_fetch.js
BigTava: Smoldot handles `DontHave` silently in [`bitswap_service.rs`](../blob/main/light-base/src/bitswap_service.rs#L538), so there was nothing for the test to scrape today. Feel free to file that as a follow-up issue to discuss and implement.

pepoviola:
> > should snapshot generation be part of zombienet sdk?
> 
> I thought about this in the past too, my 2c is that its okay to include this per-repo, because sometimes the snapshots need to replicate specific scenarios. For bulletin I can imagine scenarios with specific counts of renew/store tx etc.

I aggre, but we can offload the creation logic to zombienet-sdk itself, so per repo you only need to setup the network and call something like `network.generate_snap()`. I will add a new issue to track this in zombienet repo.

Thx!

PR #3242, opened 2026-05-05
More smoke tests

lrubasze:
## Summary
- Three scenarios under `e2e-tests/tests/`: 
  - `smoke_fresh` (genesis), 
  -   `smoke_cold` (snapshot + `lightSyncState`, no smoldot DB), 
  - `smoke_warm` (snapshot + `lightSyncState` + persisted `databaseContent`).
- Snapshot bundle published to `gs://zombienet-db-snaps/zombienet/smoldot_smoke_db/v1/bundle.tar.gz`, SHA-pinned in `src/snapshot.rs`. `ARTIFACTS_DIR_OVERRIDE` skips download for local iteration
- snapshot generator added as a dedicated test `smoke_generate_snapshots`

Full description in `e2e-tests/docs/smoke-scenarios.md`


AndreiEres (approved):
Statement store tests still work :-)

e2e-tests/js/smoke.js
skunert: Not sure if you are aware of this, but [`chainHead_v1_followEvent`](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_v1_follow.html#usage) emits on subscription first the finalized head and then a newBlock event per already known block. So if we want to count real new blocks since subscription time, we would need to wait for the initial burst. Just a thing to consider.

lrubasze: Good catch 😄 
Updated

gab8i (approved):

skunert (approved):

PR #3244, opened 2026-05-06
statement-store: Add browser sanity check zombienet test

AndreiEres:
## Summary

- Add a Playwright-driven zombienet test that runs smoldot's browser bundle (`index-browser.js`) inside headless Chromium against a real polkadot-sdk full node, exercising the statement-store ping-pong path that Node-based tests cannot reach (WebSocket-only transport, streaming wasm `fetch`, WebCrypto, the published browser entry point).
- The test follows the existing reception harness pattern: pre-populate stmt_B on alice/bob, launch the browser runner, wait for smoldot to peer with both collators, signal `READY`, then the page submits stmt_A (ping) and waits for stmt_B to arrive on its subscription (pong).
- Wire the new test into `.github/workflows/zombienet.yml` as `zombienet-smoldot-0004-statement_store_browser`, with caches for `e2e-tests/browser/node_modules` and `~/.cache/ms-playwright`, and a conditional `npx playwright install --with-deps chromium` step.

## Test plan

- [x] `cargo check --tests` clean for `smoldot-e2e-tests`
- [x] `node --check` for `e2e-tests/browser/run.mjs` and `helpers.js`
- [x] Local end-to-end run (zombienet native, polkadot-sdk binaries on PATH): `cargo test -p smoldot-e2e-tests --test statement_store_browser -- --nocapture` — five `PASS:` lines (bundle loaded, subscribe to topic_B, READY, ping, pong); test result: ok
- [x] CI: `zombienet-smoldot-0004-statement_store_browser` job passes


lrubasze (approved):

alvicsam (approved):

e2e-tests/browser/statement_store_browser.js
```diff
@@ -0,0 +1,215 @@
...
+  await waitForSyncMessage(process.env.SYNC_PATH, "READY", 120_000);
+  report("Rust signalled READY", true);
+
+  // Phase 3: ping (submit stmt_A) + pong (await stmt_B notification).
+  const result = await page.evaluate(
+    async ([stmtAHex, stmtBHex, subscriptionId]) => {
+      const { send, waitForResponse, client } = window.__t;
+
+      let pingResult = null;
+      for (let attempt = 0; attempt < 10; attempt++) {
```
P1sar: Why do we need to repeat submit/read 10 times? Should not is pass on first try?

e2e-tests/browser/page/index.html
```diff
@@ -0,0 +1,11 @@
+<!doctype html>
+<html>
+  <head><meta charset="utf-8"><title>smoldot browser tests</title></head>
+  <body>
+    <script type="module">
+      import * as smoldot from '/smoldot/dist/mjs/index-browser.js';
```
P1sar, line 6: very minor, but polkadot host uses Vite as a bundler maybe it make sense to also use it to build smoldot for this test. Bundlers can introduce their own problems

e2e-tests/browser/statement_store_browser.js
AndreiEres: You're right; it's too defensive.

e2e-tests/browser/page/index.html
AndreiEres: very don't want to add Vite here. I tried to keep the test very simple.

PR #3246, opened 2026-05-07
Skip post-warp-sync wait in parachain bootstrap

lrubasze:
## Problem

After a relay-chain warp sync, the runtime_service places the warp-synced block directly as the outer `finalized_block` — it never enters the inner `async_tree`, so subscribers never receive `Block`/`Finalized` notifications for it. `fetch_parachain_head_from_relay` waits for the next `Notification::Finalized`, which only fires when the relay finalizes the block *after* the warp-synced one (~6–12 s).

## Fix

Reshape the warp-sync `MustSubscribe` path so the warp-synced block flows through the normal `OutputUpdate::Block` → `OutputUpdate::Finalized` sequence:

- Previous finalized block (`pre_warp_finalized`) becomes the tree's initial finalized state — a tree-level predecessor only, not the real chain parent.
- Warp-synced block is inserted as a non-finalized child with its runtime pre-completed (no actual download), then `input_finalize`'d.
- Subscribers receive `pre_warp_finalized` as initial finalized, then `Block(warp_synced)` + `Finalized(warp_synced)` on the channel back-to-back.

`pre_warp_finalized` is pruned the moment warp-synced is finalized; its placeholder runtime is `Err(CodeNotFound)`.

This PR replaces https://github.com/paritytech/smoldot/pull/3210

## Tests

Unit-test coverage for the warp-sync-tree init introduced in the parent branch.
- Extracts `build_warp_sync_tree` from the `MustSubscribe` runtime path in `runtime_service.rs`.
- Added unit tests covering: pre-warp set / skipped (self-finalize, no input-finalized fallback), child attachment, and the `Block`+`Finalized` emission for the warp-synced block.


lrubasze:
## Benchmark results

Paseo + Asset Hub Paseo, time-to-finalized, n=10 iterations.

| Mode | Baseline `main` branch `f4f2b69e`  (median / p95 / max) | PR `f361eab2` (median / p95 / max) | Speedup (median / p95) |
|------|------------------------------------------|------------------------------------|------------------------|
| Cold | 11.4s / 56.4s / 59.1s                    | 5.5s / 9.8s / 12.3s                | **2.1× / 5.8×**        |
| Warm | 17.8s / 91.0s / 92.9s                    | 4.1s / 7.9s / 9.0s                 | **4.3× / 11.5×**       |

Highlights:
- Warm variance collapses: σ 30.0s → 1.7s.
- Cold variance: σ 18.2s → 2.2s; worst run 12.3s vs baseline 59.1s.
- Baseline warm had 1/10 run fail to finalize; PR has 10/10.

lib/src/chain/async_tree.rs
```diff
@@ -374,6 +374,15 @@ where
         self.input_best_block_index
     }
 
+    /// Returns the user data of the current "input" finalized block.
+    ///
+    /// Returns `None` if the input finalized block is the output finalized block (i.e. the
+    /// async_tree's output has caught up to the input).
```
bkchr:
```suggestion
```

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,148 @@ async fn run_background<TPlat: PlatformRef>(
...
+                            Arc::new(Runtime {
+                                runtime: Err(RuntimeError::CodeNotFound),
+                                runtime_code: None,
+                                heap_pages: None,
+                                code_merkle_value: None,
+                                closest_ancestor_excluding: None,
+                            })
+                        } else {
+                            runtime.clone()
+                        };
```
bkchr: This can just be moved into the `if` below. Then we don't need to clone `runtime` for nothing.

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,148 @@ async fn run_background<TPlat: PlatformRef>(
...
+                        let warp_synced_block = Block {
+                            hash: warp_synced_hash,
+                            height: warp_synced_height,
+                            scale_encoded_header: subscription.finalized_block_scale_encoded_header,
+                        };
+
+                        let (finalized_async_user_data, finalized_block) = if use_pre_warp_finalized
+                        {
+                            // use_pre_warp_finalized implies pre_warp_finalized.is_some().
+                            (pre_warp_finalized_runtime, pre_warp_finalized.unwrap())
```
bkchr:
```suggestion
                        let (finalized_async_user_data, finalized_block) = if let Some(finalized_block) = pre_warp_finalized.filter(|_| use_pre_warp_finalized)
                        {
                            (pre_warp_finalized_runtime, finalized_block)
```

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,148 @@ async fn run_background<TPlat: PlatformRef>(
...
+                            {
+                                async_tree::NextNecessaryAsyncOp::Ready(op) => op,
+                                async_tree::NextNecessaryAsyncOp::NotReady { .. } => unreachable!(
+                                    "warp-synced block just inserted with pending async op"
+                                ),
+                            };
+                            debug_assert_eq!(async_op.block_index, warp_synced_idx);
+                            new_tree.async_op_finished(async_op.id, runtime);
+                            new_tree.input_finalize(warp_synced_idx);
+                        }
```
bkchr: Can you explain what this changes?

bkchr (commented):
You have merged these two branches, can you explain on what is the difference to before? Why is this now sending the required events.

bkchr:
Don't we just need to set [this](https://github.com/paritytech/smoldot/blob/d34230fc7f9b90caccda4460b074b9dc6b8b3c49/light-base/src/runtime_service.rs#L1497) to `true`?  Then we don't need your `next_necessary_async_op` call?

light-base/src/runtime_service.rs
lrubasze: What I wanted here is to:
- attach the warp-synced runtime to the warp_synced_block
- mark the block as the next finalized

Not sure if this is fine to set `same_async_op_as_parent = true`, because it would implicate using the `finalized_async_user_data` we used to build the tree, which I deliberately set as Err(RuntimeError::CodeNotFound), because there is no actual runtime that could be associated with `pre_warp_finalized` block (there is no guarantee that pre_warp_finalized block uses the same runtime as `warp_synced_block`).
We could use warp-synced runtime as `finalized_async_user_data` when building a tree and then we could use approach suggested by you. Though not sure about the case when someone invokes runtime_call against `pre_warp_finalized` block and its runtime is supposed to be different than for `warp_synced_block`.
If this is a case, then better to return `RuntimeError::CodeNotFound`.

lrubasze:
> Don't we just need to set [this](https://github.com/paritytech/smoldot/blob/d34230fc7f9b90caccda4460b074b9dc6b8b3c49/light-base/src/runtime_service.rs#L1497) to `true`? Then we don't need your `next_necessary_async_op` call?

https://github.com/paritytech/smoldot/pull/3246#discussion_r3210688842

lib/src/chain/async_tree.rs
lrubasze: Done

light-base/src/runtime_service.rs
lrubasze: Done

light-base/src/runtime_service.rs
lrubasze: NIce! Done

lrubasze:
> You have merged these two branches, can you explain on what is the difference to before? Why is this now sending the required events.

Not sure what branches you mean. Is this about the difference between this PR and #3210 ?

light-base/src/runtime_service.rs
bkchr: > We could use warp-synced runtime as `finalized_async_user_data` when building a tree and then we could use approach suggested by you. Though not sure about the case when someone invokes runtime_call against `pre_warp_finalized` block and its runtime is supposed to be different than for `warp_synced_block`.

But `pre_warp_finalized` should not be part of the tree? 

Looking closer into the async tree. If we set `same_async_op_as_parent = true`, we set `AsyncOp::FInished`. Then we should [end up here](https://github.com/paritytech/smoldot/blob/37482d9a635a17f9e6a77aa245b4684b2f72e750/lib/src/chain/async_tree.rs#L1080) and next call to the same function we should go [here](https://github.com/paritytech/smoldot/blob/37482d9a635a17f9e6a77aa245b4684b2f72e750/lib/src/chain/async_tree.rs#L903) and finalize the block. Or what do I miss?

lrubasze: Ok, probably my approach was too defensive. I was assuming that `pre_warp_finalized` might be reported to subscribers, which might further invoke runtime call against it.
But this is not the case, so we can set the `warp_synced_block`'s runtime when building a tree and then use suggested approach.

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,121 @@ async fn run_background<TPlat: PlatformRef>(
...
+                                    hash: b.hash,
+                                    height: b.height,
+                                    scale_encoded_header: b.scale_encoded_header.clone(),
+                                })
+                            }
+                        };
+                        // Disable when missing or degenerately equal to warp-synced.
+                        let use_pre_warp_finalized = pre_warp_finalized
+                            .as_ref()
+                            .is_some_and(|b| b.hash != warp_synced_hash);
```
bkchr:
```suggestion
                        let pre_warp_finalized: Option<Block> = match &background.tree {
                            Tree::FinalizedBlockRuntimeKnown {
                                finalized_block, ..
                            } => Some(Block {
                                hash: finalized_block.hash,
                                height: finalized_block.height,
                                scale_encoded_header: finalized_block.scale_encoded_header.clone(),
                            }),
                            Tree::FinalizedBlockRuntimeUnknown { tree } => {
                                tree.input_finalized_user_data().map(|b| Block {
                                    hash: b.hash,
                                    height: b.height,
                                    scale_encoded_header: b.scale_encoded_header.clone(),
                                })
                            }
                        }
                        // Disable when missing or degenerately equal to warp-synced.
                        .filter(|b| b.hash != warp_synced_hash);
```

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,121 @@ async fn run_background<TPlat: PlatformRef>(
...
+
+                        log!(
+                            &background.platform,
+                            Debug,
+                            &background.log_target,
+                            "warp-sync-tree-init",
+                            use_pre_warp_finalized,
```
bkchr:
```suggestion
```

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,121 @@ async fn run_background<TPlat: PlatformRef>(
...
+                            // `Block` + `Finalized` without an actual download.
+                            // `pre_warp_finalized`'s real on-chain runtime is unknown and may
+                            // differ from the warp-synced runtime, but `pre_warp_finalized` is
+                            // tree-internal only (never reported to subscribers), so the outer-
+                            // finalized slot holds the warp-synced runtime and the insert below
+                            // copies it onto the warp-synced node via `same_async_op_as_parent`.
+                            let warp_synced_idx =
+                                new_tree.input_insert_block(warp_synced_block, None, true, true);
+                            new_tree.input_finalize(warp_synced_idx);
+                            pre_warp_finalized.unwrap()
```
bkchr:
```suggestion
                        let finalized_block = if let Some(pre_warp_finalized) = pre_warp_finalized {
                            // Pre-complete the runtime so the tree-advance loop emits
                            // `Block` + `Finalized` without an actual download.
                            // `pre_warp_finalized`'s real on-chain runtime is unknown and may
                            // differ from the warp-synced runtime, but `pre_warp_finalized` is
                            // tree-internal only (never reported to subscribers).
                            let warp_synced_idx =
                                new_tree.input_insert_block(warp_synced_block, None, true, true);
                            new_tree.input_finalize(warp_synced_idx);
                            pre_warp_finalized
```

light-base/src/runtime_service.rs
```diff
@@ -1355,10 +1355,10 @@ async fn run_background<TPlat: PlatformRef>(
 
                     // TODO: DRY below
                     if let Some(finalized_block_runtime) = subscription.finalized_block_runtime {
-                        let finalized_block_hash = header::hash_from_scale_encoded_header(
+                        let warp_synced_hash = header::hash_from_scale_encoded_header(
                             &subscription.finalized_block_scale_encoded_header,
                         );
-                        let finalized_block_height = header::decode(
+                        let warp_synced_height = header::decode(
```
bkchr: This rename is misleading, because this code path can also be called under different circumstances?

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,121 @@ async fn run_background<TPlat: PlatformRef>(
...
+                                None
+                            } else {
+                                Some(
+                                    new_tree
+                                        .input_output_iter_unordered()
+                                        .find(|b| b.user_data.hash == block.parent_hash)
+                                        .unwrap()
+                                        .id,
+                                )
+                            };
```
bkchr: Do we actually need the pre warp sync finalized? Because we never add it to the `tree`. 

The `pre_warp_finalized` is only used in this if branch to set the `parent_index` to `None`. Which could also just be achieved by not having this beautiful `unwrap` in here :D

bkchr (commented):
I think we are close. As discussed in the call, it would be ultra nice to have some sort of test :D

light-base/src/runtime_service.rs
lrubasze: Yes, I think it is needed.

Here is the flow after the `MustSubscribe` handler is done:

```
background.tree = Tree::FinalizedBlockRuntimeKnown {
    finalized_block: pre_warp_finalized,         // pre_warp_finalized used here
    tree: new_tree,                              // see below
    all_blocks_subscriptions: empty,             // subscriptions are wiped
    pinned_blocks: empty,
}
new_tree:
    output_finalized_async_user_data = warp_synced_runtime
    non_finalized_blocks = { warp_synced: async_op=Finished{warp_synced_runtime, reported:false} }
    input_finalized_index = Some(warp_synced_idx)
```

The main loop then selects `wake_up_reason`. Possible options in the `.or` chain:

1. Pending subscriptions ([L882](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L882)) — possible if someone subscribed after
`MustSubscribe` wiped the subscriptions, which is VERY unlikely.
2. Fallback to a tree-advance attempt (no other wake-up reasons are possible):
   a) Tree advance fires `OutputUpdate::Block(warp_synced)`.
   b) Tree advance fires `OutputUpdate::Finalized(warp_synced)`.

**Ad. 1.**

Build the subscription response. Decode `pre_warp_finalized` (a.k.a. `finalized_block`) at
[L1639](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1639) to pin it.

- Loop over tree ancestors ([L1658](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1658)):
  - `warp_synced` is skipped because its `async_op_user_data` is `None` (since `reported: false`).
  - There are no other entries in the tree at this stage.
- The subscriber(s) will receive ([L1732-L1748](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1732-L1748)):
  - `finalized = pre_warp_finalized`
  - `finalized_runtime = warp_synced_runtime` *(might not match `pre_warp_finalized`)*
  - `non_finalized = []`

Again, this scenario is highly unlikely.

**Ad. 2.**

a) Tree advance fires `OutputUpdate::Block(warp_synced)` ([L1115](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1115)).
   - Build `Notification::Block`:
     `parent_hash = finalized_block.hash` ([L1158-L1160](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1158-L1160)) - *pre_warp_finalized used here*. 
`tree.parent(warp_synced)` is `None`, so it falls back to `pre_warp_finalized.hash`.

b) Tree advance fires `OutputUpdate::Finalized(warp_synced)` ([L1021](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1021)).
   - `*finalized_block = new_finalized` (= `warp_synced_block`) at [L1038](https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base/src/runtime_service.rs#L1038) - *pre_warp_finalized dropped here*.

lrubasze: So `pre_warp_finalized` is used in couple of more places. The most important I think is the `Notification::Block`, where it is used as a `parent_hash` of the `warp_synced` block. 
We could make this field optional or add some other workaround but not sure if we want to go this way.

lrubasze:
> I think we are close. As discussed in the call, it would be ultra nice to have some sort of test :D

Added as a separate PR https://github.com/paritytech/smoldot/pull/3251 to this branch.
Test [notifies_subscribers_of_warp_synced_block](https://github.com/paritytech/smoldot/blob/34972e671b9690b4c8e715809a3f9d0a63ddc8d8/light-base/src/runtime_service/tests.rs#L132) covers the path exercised in this PR

light-base/src/runtime_service.rs
lrubasze: Done

light-base/src/runtime_service.rs
lrubasze: Makes sense

light-base/src/runtime_service.rs
lrubasze: Done

light-base/src/runtime_service.rs
lrubasze: Done

light-base/src/runtime_service.rs
bkchr: > So `pre_warp_finalized` is used in couple of more places. The most important I think is the `Notification::Block`, where it is used as a `parent_hash` of the `warp_synced` block. We could make this field optional or add some other workaround but not sure if we want to go this way.

Where is this?

lrubasze: https://github.com/paritytech/smoldot/blob/82b99cf61908aeca4fe39a08e2ebae59c5d18a23/light-base%2Fsrc%2Fruntime_service.rs#L1160

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,116 @@ async fn run_background<TPlat: PlatformRef>(
...
```
bkchr:
```suggestion
                            let parent_index = 
                                    new_tree
                                        .input_output_iter_unordered()
                                        .find(|b| b.user_data.hash == block.parent_hash)
                                        .map(|n| n.id);
```

bkchr (approved):

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,116 @@ async fn run_background<TPlat: PlatformRef>(
...
+                        let new_finalized_block = Block {
+                            hash: finalized_block_hash,
+                            height: finalized_block_height,
+                            scale_encoded_header: subscription.finalized_block_scale_encoded_header,
+                        };
 
-                                    let same_runtime_as_parent = same_runtime_as_parent(
+                        let mut new_tree =
+                            async_tree::AsyncTree::<_, Block, _>::new(async_tree::Config {
+                                finalized_async_user_data: runtime,
```
lexnv: dq: Could we end up with a situation where the runtime is newer than the older `pre_warp_finalized`? 

Then if the user makes a runtime API call to the first reported finalized block, we'll use the wrong runtime against `pre_warp_finalized` state?

light-base/src/runtime_service.rs
```diff
@@ -1412,64 +1412,116 @@ async fn run_background<TPlat: PlatformRef>(
...
+                            Tree::FinalizedBlockRuntimeUnknown { tree } => {
+                                tree.input_finalized_user_data().map(|b| Block {
+                                    hash: b.hash,
+                                    height: b.height,
+                                    scale_encoded_header: b.scale_encoded_header.clone(),
+                                })
+                            }
+                        }
+                        // Disable when missing or degenerately equal to warp-synced.
+                        .filter(|b| b.hash != finalized_block_hash);
```
lexnv: nit: Should we check that these blocks have heights smaller than our `finalized_block_height`?

lexnv (approved):

light-base/src/runtime_service.rs
lrubasze: This case is very unlikely. Described in point 1. In this comment 

https://github.com/paritytech/smoldot/pull/3246#discussion_r3220016313

skunert (approved):

light-base/src/runtime_service.rs
lrubasze: yes, that makes sense!

light-base/src/runtime_service.rs
lrubasze: Done

PR #3249, opened 2026-05-08
e2e-tests: Fix bulletin_fetch flake by staging the relay snapshot per-validator

BigTava:
## Problem

`bulletin_fetch` flakes intermittently with a panic in `zombienet-provider`'s `archive.unpack().unwrap()`:

```
called `Result::unwrap()` on an `Err` value: Custom { kind: UnexpectedEof,
  error: TarError { desc: "failed to unpack `data/chains/.../db/full/000008.log`", ... } }
```

Reproduced 3/5 in a rerun loop on a single commit. Failure path was inside zombienet's local copy/read/extract pipeline.

## Cause

`alice` and `bob` both passed the **same** relay tarball path to `with_db_snapshot(...)`. `zombienet-provider` keys its internal extraction cache by `sha256(path_string)`, so the two validators landed in the same cache slot and raced writing/reading the same intermediate `<hash>.tgz` in the namespace dir. When the read won the race against an in-progress write, tar parsing hit `UnexpectedEof` mid-entry.

### DIFF
--- a/e2e-tests/src/lib.rs
+++ b/e2e-tests/src/lib.rs
@@ -164,6 +164,59 @@ pub async fn run_browser_test(
     }
 }
 
+/// Download a snapshot to a stable local cache and return its path. If
+/// `override_var` is set, its value is used as the source instead of
+/// `default_url`; if the source is already a local path, it's returned
+/// as-is. URLs are fetched with `curl -fL --retry`. Cached under
+/// `e2e-tests/target/snapshot-cache/<basename>`.
+pub fn prefetch_snapshot(default_url: &str, override_var: &str) -> Result<String, anyhow::Error> {
+    let source = std::env::var(override_var).unwrap_or_else(|_| default_url.to_string());
+    if !source.starts_with("http://") && !source.starts_with("https://") {
+        return Ok(source);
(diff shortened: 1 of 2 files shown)


skunert:
Thanks! Can you maybe unify this with the other prefetchers we have in the tests?

I think the proper solution would be for zombiened-sdk to handle this gracefully, it seems totally normal to have multiple nodes reference the same snapshot. Opened an [issue](https://github.com/paritytech/zombienet-sdk/issues/540) on zombienet-sdk, in the meantime we should keep the workarounds.

cc @pepoviola

pepoviola:
Hi @BigTava / @skunert, thanks for reporting this issue. Looks like in native provider we took the decision of _not copy_ the snap per node  and we can have the described race condition when multiple nodes use the same snap. I easy workaround to use until I draft a new release with this fixed is to set in the global settings `spawn_concurrency = 1`, so this will make the spawning logic sequential an should works as expected.

I will work in the issue and ping you for bumping the version of zombienet when is ready.

Thx!

skunert:
Okay then lets use spawn_concurrency = 1 for now, thanks!

PR #3251, opened 2026-05-11
runtime_service: add unit tests for warp-sync-tree init

lrubasze:
## Summary
Unit-test coverage for the warp-sync-tree init introduced in the parent branch.

- Extracts `build_warp_sync_tree` from the `MustSubscribe` runtime path in `runtime_service.rs`.
- Added unit tests covering: pre-warp set / skipped (self-finalize, no input-finalized fallback), child attachment, and the `Block`+`Finalized` emission for the warp-synced block.


lrubasze:
The tests call `build_warp_sync_tree` directly. To test more of `runtime_service`, the next step would be to mock its dependencies (`SyncService`, `NetworkService`, `PlatformRef`) and drive the whole `run_background` loop.
But it'd require more time to add it, so maybe for now we could stick to the current shape.

light-base/src/runtime_service.rs
```diff
@@ -1355,10 +1355,10 @@ async fn run_background<TPlat: PlatformRef>(
 
                     // TODO: DRY below
                     if let Some(finalized_block_runtime) = subscription.finalized_block_runtime {
-                        let warp_synced_hash = header::hash_from_scale_encoded_header(
+                        let finalized_block_hash = header::hash_from_scale_encoded_header(
                             &subscription.finalized_block_scale_encoded_header,
                         );
-                        let warp_synced_height = header::decode(
+                        let finalized_block_height = header::decode(
```
bkchr: Please revert here as well.

light-base/src/runtime_service.rs
```diff
@@ -3209,6 +3157,97 @@ fn same_runtime_as_parent(header: &[u8], block_number_bytes: usize) -> bool {
...
+        let parent_index = if child.parent_hash == finalized_block.hash {
+            None
+        } else {
+            Some(
+                tree.input_output_iter_unordered()
+                    .find(|b| b.user_data.hash == child.parent_hash)
+                    .unwrap()
+                    .id,
+            )
+        };
```
bkchr: Same as in the main pr.

light-base/src/runtime_service.rs
```diff
@@ -1412,121 +1412,69 @@ async fn run_background<TPlat: PlatformRef>(
...
+                                    hash: header::hash_from_scale_encoded_header(
+                                        &b.scale_encoded_header,
+                                    ),
+                                    height: header::decode(
+                                        &b.scale_encoded_header,
+                                        block_number_bytes,
+                                    )
+                                    .unwrap()
+                                    .number,
+                                    scale_encoded_header: b.scale_encoded_header.clone(),
```
lexnv: nit: We could avoid this clone if inside the map we compute:

```rust
let same_runtime_as_parent = same_runtime_as_parent(
                                    &b.scale_encoded_header,
                                    block_number_bytes,
                                );
let hash = header::hash_from_scale_encoded_header(
                                        &b.scale_encoded_header,
                                    );
let height = height: header::decode(
                                        &b.scale_encoded_header,
                                        block_number_bytes,
                                    ) .unwrap()
                                    .number;

 WarpSyncTreeChild {
          block: Block { hash, height, scale_encoded_header: b.scale_encoded_header },
          parent_hash: b.parent_hash,
          same_runtime_as_parent,
          is_new_best: b.is_new_best,
      }

```

lexnv (approved):

light-base/src/runtime_service/tests.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    match result.tree.try_advance_output() {
+        Some(async_tree::OutputUpdate::Finalized {
+            user_data,
+            pruned_blocks,
+            ..
+        }) => {
+            assert_eq!(user_data.hash, new_finalized_hash);
+            assert!(pruned_blocks.is_empty());
```
skunert, line 169: DQ: The comment here says that the pre-warp block is pruned the moment the warp synced block is finalized: https://github.com/paritytech/smoldot/pull/3246/changes#diff-e877adbdbf0987ad30e3b4b12255071651f255a37a41483fb1950b75d9b0788fR1419

So we would expect it here right? Is the comment outdated or did I get it wrong?

skunert (approved):
Tests look good

light-base/src/runtime_service/tests.rs
lrubasze: The comment is imprecise (updated)

`pruned_blocks` lists blocks that were part of non-finalized blocks, but are being dropped now because they are not part of the canonical path. In other words, `pruned_blocks` lists forked blocks.

Additionally, `pre_warp_finalized` itself is not part of the async_tree. It is the outer-finalized — held by the wrapper struct. `pruned_blocks` only lists blocks that were in the tree.

PR #3252, opened 2026-05-13
set zombienet spawn concurrency to 1 as workaround

pepoviola:
This should fix the race condition that cause the _flakyness_ of the test.


skunert (approved):

skunert:
We should do the same for the other tests and remove the boilerplate snapshot prepare code.

lrubasze (approved):

PR #3254, opened 2026-05-13
Release process: auto-tag + /smoldot-release command

lrubasze:
## Summary

Two related changes:
1. Add a `/smoldot-release` slash command (`.claude/commands/smoldot-release.md`) that walks through the runbook in `docs/RELEASING.md` — useful when cutting a release with Claude Code.

2. Add `tags-publish` job to `deploy.yml` that pushes `npm-smoldot-v…`, `smoldot-v…`, and`smoldot-light-v…` on the merge commit, mirroring `deno-publish`. 
Updates `docs/RELEASING.md` accordingly

## Behavior (tags-publish)

- Runs on push to `main` after `npm-publish`, `deno-publish`, `crates-io-publish`.
- Per-tag existence check: non-bumping pushes are a no-op; works for any combination of
npm/lib/light-base bumps.
- Semver-regex validates parsed versions; `concurrency:` serializes back-to-back pushes.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -349,58 +349,107 @@ jobs:
         continue-on-error: true
 
   tags-publish:
-    # Creates the three release-marker tags (`npm-smoldot-v…`, `smoldot-v…`,
-    # `smoldot-light-v…`) on the merge commit. Per-tag existence check makes
-    # the job idempotent: pushes to main that don't bump a given version are
-    # a no-op for that tag because its name already points at a prior commit.
+    # Pushes `npm-smoldot-v…`, `smoldot-v…`, `smoldot-light-v…` — each only
+    # after the version is found on its registry. Guards against publish
+    # steps' `continue-on-error: true` letting a tag through on failure.
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     needs: [npm-publish, deno-publish, crates-io-publish]
-    # Serialize concurrent runs so two back-to-back pushes can't race on
-    # `git push` of the same new tag.
+    # Serialize runs to avoid `git push` races on the same new tag.
     concurrency:
       group: tags-publish
       cancel-in-progress: false
     runs-on: ubuntu-latest
     permissions:
-      contents: write   # Necessary in order to push tags.
+      contents: write   # to push tags
     steps:
       - uses: actions/checkout@v6
-        with:
-          fetch-depth: 0   # Necessary below for checking if tags exist.
       - id: tags
         name: Compute and validate tag names
         run: |
           set -euo pipefail
-          # First `version = "..."` line per Cargo.toml = [package].version.
-          # Yields empty on `version.workspace = true`; caught by the semver
-          # check below before any tag is pushed.
           NPM_VER=$(jq -r .version ./wasm-node/javascript/package.json)
-          SMOLDOT_VER=$(sed -nE '0,/^version = "(.+)"$/s//\1/p' ./lib/Cargo.toml)
-          LIGHT_VER=$(sed -nE '0,/^version = "(.+)"$/s//\1/p' ./light-base/Cargo.toml)
+          SMOLDOT_VER=$(cargo read-manifest --manifest-path=./lib/Cargo.toml        | jq -r .version)
+          LIGHT_VER=$(cargo  read-manifest --manifest-path=./light-base/Cargo.toml  | jq -r .version)
           for v in "$NPM_VER" "$SMOLDOT_VER" "$LIGHT_VER"; do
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -348,10 +348,64 @@ jobs:
...
+          done
+          echo "npm=npm-smoldot-v$NPM_VER"       >> "$GITHUB_OUTPUT"
+          echo "smoldot=smoldot-v$SMOLDOT_VER"   >> "$GITHUB_OUTPUT"
+          echo "light=smoldot-light-v$LIGHT_VER" >> "$GITHUB_OUTPUT"
+          printf 'Computed tags:\n  npm-smoldot-v%s\n  smoldot-v%s\n  smoldot-light-v%s\n' \
+            "$NPM_VER" "$SMOLDOT_VER" "$LIGHT_VER"
+      - name: Push missing release tags
+        run: |
+          set -euo pipefail
+          for TAG in \
```
lexnv: Should we also fetch tags before `git fetch --tags origin`?

.github/workflows/deploy.yml
```diff
@@ -348,10 +348,64 @@ jobs:
...
+              '${{ steps.tags.outputs.npm }}' \
+              '${{ steps.tags.outputs.smoldot }}' \
+              '${{ steps.tags.outputs.light }}'; do
+            if git rev-parse --verify --quiet "refs/tags/$TAG" >/dev/null; then
+              echo "Tag $TAG already exists; skipping."
+            else
+              git tag "$TAG"
```
lexnv: Could we be pushing tags even if the cargo publish fails:
- for example `crates-io-publish` runs with 

```
      - run: cargo publish --no-verify
        working-directory: ./light-base
        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
        continue-on-error: true

```

Which should continue on errors? So we could have crates unreleased but tags created?

.github/workflows/deploy.yml
```diff
@@ -348,10 +348,64 @@ jobs:
...
+      - id: tags
+        name: Compute and validate tag names
+        run: |
+          set -euo pipefail
+          # First `version = "..."` line per Cargo.toml = [package].version.
+          # Yields empty on `version.workspace = true`; caught by the semver
+          # check below before any tag is pushed.
+          NPM_VER=$(jq -r .version ./wasm-node/javascript/package.json)
+          SMOLDOT_VER=$(sed -nE '0,/^version = "(.+)"$/s//\1/p' ./lib/Cargo.toml)
+          LIGHT_VER=$(sed -nE '0,/^version = "(.+)"$/s//\1/p' ./light-base/Cargo.toml)
```
lexnv: We could use the read-manifest from cargo, its easier to reason than seds

```
cargo read-manifest --manifest-path=./Cargo.toml | jq -r .version
```

.github/workflows/deploy.yml
lrubasze: Not needed -`actions/checkout@v6` above should cover this:

https://github.com/paritytech/smoldot/pull/3254/changes/BASE..03a13104783d22d2d68d073cdc340426c070904f#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R367-R369

lexnv: Or we could use a ` git ls-remote --tags origin $TAG` and aavoid the fetch entirely

docs/RELEASING.md
```diff
@@ -238,31 +236,23 @@ Expected quirk: the `crates-io-publish` job's `smoldot` step fails with
 
 ---
 
-## 11. Create and push the manual tags
+## 11. Verify CI-pushed tags
```
lexnv, line 240: In step 6 we do a 

```
## 6. Verify locally
``` 

But in the smoldot-release.md we also mentioned `--alow-dirty`, should we propagate that to the step 6 from here?

.github/workflows/deploy.yml
lrubasze: I was thinking of running this job on the vanilla ubuntu GH runner, no toolchain, no cargo. Just to spare some time. Not sure, maybe it does not make sense. 
I will change it. It will read better.

lexnv (approved):

skunert (approved):

.github/workflows/deploy.yml
lrubasze: Nice idea! Done

.github/workflows/deploy.yml
lrubasze: Good point. Refactored, so the tags are pushed only if publish succeeded

.github/workflows/deploy.yml
lrubasze: Done

docs/RELEASING.md
lrubasze: Nice catch! Done

PR #3256, opened 2026-05-13
fix(webrtc): warp sync slow initialization and webrtc handshake bug

gab8i:
The following PR just extract work already done within https://github.com/paritytech/smoldot-archive/pull/2222.

The fixes are:
- Solving a deadlock within the multistream-select task loop
- Initializing warp sync faster by avoiding the need to wait for Grandpa notifications. This is especially needed by WebRTC because data channel creation introduces latency, and thus not waiting for the channel to open saves time during syncing.


lrubasze (approved):

dmitry-markin (approved):
Substream waking in WebRTC looks good.

The idea behind optimistically targeting warp sync at the best block and not finalized one is also good, but we need to rename `finalized_block_height` to something like `optimistic_finalized_block_height`, or a better name, so future code doesn't rely on this value as actually being finalized.

Also, the comment at `warp_sync.rs:768` likely needs updating and the corresponding logic checked for correctness.

light-base/src/network_service/tasks.rs
```diff
@@ -453,8 +464,16 @@ pub(super) async fn webrtc_multi_stream_connection_task<TPlat: PlatformRef>(
                 if let SubstreamFate::Continue = substream_fate {
                     when_substreams_rw_ready.push({
                         let platform = platform.clone();
+                        let write_ready = coordinator_write_ready.clone();
                         Box::pin(async move {
-                            platform.wait_read_write_again(socket.as_mut()).await;
+                            // Wait for either platform data or a coordinator write
+                            // notification (e.g. after AcceptInNotifications queues a
+                            // handshake response that needs to be flushed).
+                            let write_listener = write_ready.listen();
```
lexnv: Should we move the ` write_ready.listen()` outside the `Box::pin` future?

Currently, the write listener is registered once the the `when_substreams_rw_ready` is firstly polled.
This means we could arrive in a situation where `.notify(usize::MAX);` is called before registering the listener call.

For this to happen we need to:
- receive WakeUpReason::CoordinatorMessage(message)
- `message_sending.as_ref().as_pin_ref().is_none()` should have some data to effectively take the `None` path and ignore the `when_substreams_rw_ready.select_next_some()`

Moving the listener above `Box::pin` solves this regardless of which states call `when_substreams_rw_ready.poll_next` and avoids the problem entirely

lexnv (approved):
LGTM!

light-base/src/network_service/tasks.rs
gab8i: You are totally right, it should definitely be moved out so that it is registered and the async block later on depends on the registered event listener!

lexnv (approved):

PR #3257, opened 2026-05-13
fix(sync): avoid node ban after justification error UnknownTargetBlock

gab8i:
After warp sync finishes, the non finalized tree contains just the warp-sync target block. The node now needs to catch up from that target to the chain tip. During that window, peers will send justifications targeting blocks the local node hasn't yet downloaded.

Not banning the node avoids delays in receiving important data needed to continue the sync.

This has been extracted from [paritytech/smoldot-archive#2222](https://github.com/paritytech/smoldot-archive/pull/2222), but not being relevant only to WebRTC, it has been moved into its own pr.


lrubasze (approved):
Nice finding!

skunert (approved):

lexnv (approved):
Nice one! 🙏

Issue #3259, opened 2026-05-14
Smoldot reporting genesis parent on latest block for parachains


voliva:
Since smoldot 3.1.3, it reports wrong parents on the first block for parachains.

The `initialized` event contains the genesis block as the finalized:
```
{
  "jsonrpc": "2.0",
  "method": "chainHead_v1_followEvent",
  "params": {
    "subscription": "CsSX1byAaNda4afedSoyT5shcahqUBjExtft7ycXNJZv",
    "result": {
      "event": "initialized",
      "finalizedBlockHashes": [
        "0x68d56f15f85d3136970ec16946040bc1752654e906147f7e43e9d539d7c3de2f"
      ],
      "finalizedBlockRuntime": {
        …
      }
    }
  }
}
```

And then the next newBlock notification is one of the latest blocks on-chain (height > 15,700,000), but with "parent" the genesis block:

```
{
  "jsonrpc": "2.0",
  "method": "chainHead_v1_followEvent",
  "params": {
    "subscription": "CsSX1byAaNda4afedSoyT5shcahqUBjExtft7ycXNJZv",
    "result": {
      "event": "newBlock",
      "blockHash": "0xc03188c23080ca8c372c4dffbf235a43fe0bd0c43ae1d7af100cdc94334632f8",
      "parentBlockHash": "0x68d56f15f85d3136970ec16946040bc1752654e906147f7e43e9d539d7c3de2f",
      "newRuntime": null
    }
  }
}
```

If we request the header for this block, the header appropriately shows `parent: 0x84434125cd4ab10dc979633d6be31a9bca844d73328d92ebe32d6faabe396fcb`

This is problematic because PAPI doesn't request all the headers. Only does it for the initialized block, grabs the number, and then assumes any subsequent children has height `parent.height + 1`. So with 3.1.3, PAPI has incorrect block heights when using smoldot (meaning we can't submit transactions either as it breaks the period/phase params in mortality)

Attaching some logs level=4 (+what PAPI receives and blocks it emits) for 3.1.2, which was working, and 3.1.3, which is broken.

[genesis_3.1.2.txt](https://github.com/user-attachments/files/27752924/genesis_3.1.2.txt)

[genesis_3.1.3.txt](https://github.com/user-attachments/files/27752931/genesis_3.1.3.txt)

lrubasze:
Please check fix candidate:

[...]

PR #3260, opened 2026-05-15
runtime_service: reject genesis as pre_warp_finalized

lrubasze:
## Summary

`build_warp_sync_tree` could pick genesis as the tree's `pre_warp_finalized`, which surfaces to `chainHead_v1` subscribers as the `Initialized` event hash. Genesis state isn't reliably served by peers, so clients (e.g. PAPI) that use the Initialized hash for storage or runtime calls fail.

Above was introduced in #3246.

Observed PAPI failure on `next-asset-hub-paseo'
```
[FRAMELOG][host.localhost:5173/][debug] [smoldot +18.46s] [json-rpc-next-asset-hub-paseo] json-rpc-response-yielded; response={"jsonrpc":"2.0","method":"chainHead_v1_followEvent","params":{"subscription":"GhoxNtfsPRJMVuMrfkZCbN6Ufmg7bvjy8SPnFGc7ckc5","result":{"event":"initialized","finalizedBlockHashes":["0x173cea9df45656cf612c8b8ece56e04e9a693c69cfaac47d3628dae735067af8"]…
[FRAMELOG][host.localhost:5173/][debug] [smoldot +18.46s] [json-rpc-next-asset-hub-paseo] json-rpc-request-queued; request={"jsonrpc":"2.0","id":"1-2","method":"chainHead_v1_header","params":["GhoxNtfsPRJMVuMrfkZCbN6Ufmg7bvjy8SPnFGc7ckc5","0x173cea9df45656cf612c8b8ece56e04e9a693c69cfaac47d3628dae735067af8"]}
[FRAMELOG][host.localhost:5173/][debug] [smoldot +18.46s] [json-rpc-next-asset-hub-paseo] json-rpc-request-queued; request={"jsonrpc":"2.0","id":"1-3","method":"chainHead_v1_storage","params":["GhoxNtfsPRJMVuMrfkZCbN6Ufmg7bvjy8SPnFGc7ckc5","0x173cea9df45656cf612c8b8ece56e04e9a693c69cfaac47d3628dae735067af8",[{"key":"0x3a636f6465","type":"hash"}],null]}
[FRAMELOG][host.localhost:5173/][debug] [smoldot +18.47s] [json-rpc-next-asset-hub-paseo] json-rpc-response-yielded; response={"jsonrpc":"2.0","id":"1-3","result":{"result":"started","operationId":"DRUdj8pvY3EztntB1dfqKB5kiLYp6WcFJgkF2aa9EU4k","discardedItems":0}}
[FRAMELOG][host.localhost:5173/][debug] [smoldot +18.90s] [json-rpc-next-asset-hub-paseo] json-rpc-response-yielded; response={"jsonrpc":"2.0","method":"chainHead_v1_followEvent","params":{"subscription":"GhoxNtfsPRJMVuMrfkZCbN6Ufmg7bvjy8SPnFGc7ckc5","result":{"event":"operationInaccessible","operationId":"DRUdj8pvY3EztntB1dfqKB5kiLYp6WcFJgkF2aa9EU4k"}}}
```

## Fix

Add `b.height > 0` to the `build_warp_sync_tree` filter. When `pre_warp_finalized` would be genesis, fall back `finalized_block = new_finalized_block`.

## Test plan
- New unit tests cover both `prev_tree` variants:
  - `rejects_genesis_as_pre_warp_finalized` (FinalizedBlockRuntimeUnknown with genesis input-finalized)
  - `rejects_genesis_in_known_prev_tree` (FinalizedBlockRuntimeKnown with finalized=genesis)

### DIFF
--- a/light-base/src/runtime_service.rs
+++ b/light-base/src/runtime_service.rs
@@ -3217,9 +3217,11 @@ fn build_warp_sync_tree<TPlat: PlatformRef>(
             })
         }
     }
-    // Disable when missing, degenerately equal to the new finalized, or
-    // not strictly older.
-    .filter(|b| b.hash != new_finalized_block.hash && b.height < new_finalized_block.height);
+    // Disable when missing, degenerately equal to the new finalized, not strictly older,
+    // or pointing at genesis
+    .filter(|b| {
+        b.hash != new_finalized_block.hash && b.height < new_finalized_block.height && b.height > 0
+    });
 
     let pre_warp_finalized_hash = pre_warp_finalized.as_ref().map(|b| b.hash);
 
--- a/light-base/src/runtime_service.rs
+++ b/light-base/src/runtime_service.rs
@@ -3217,12 +3217,12 @@ fn build_warp_sync_tree<TPlat: PlatformRef>(
             })
         }
     }
-    // Disable when missing, degenerately equal to the new finalized, not strictly older,
-    // or pointing at genesis. Genesis state isn't reliably served by peers, so clients
+    // Disable when missing, pointing at genesis, degenerately equal to the new finalized,
+    // or not strictly older. Genesis state isn't reliably served by peers, so clients
     // would fail storage/runtime calls against the Initialized hash. Relay falls back
     // here when no checkpoint is set; parachain always falls back at first bootstrap.
     .filter(|b| {
-        b.hash != new_finalized_block.hash && b.height < new_finalized_block.height && b.height > 0
+        b.height > 0 && b.hash != new_finalized_block.hash && b.height < new_finalized_block.height
     });
 
     let pre_warp_finalized_hash = pre_warp_finalized.as_ref().map(|b| b.hash);
--- a/light-base/src/runtime_service.rs
+++ b/light-base/src/runtime_service.rs
@@ -3218,7 +3218,9 @@ fn build_warp_sync_tree<TPlat: PlatformRef>(
         }
     }
     // Disable when missing, degenerately equal to the new finalized, not strictly older,
-    // or pointing at genesis
+    // or pointing at genesis. Genesis state isn't reliably served by peers, so clients
+    // would fail storage/runtime calls against the Initialized hash. Relay falls back
+    // here when no checkpoint is set; parachain always falls back at first bootstrap.
     .filter(|b| {
         b.hash != new_finalized_block.hash && b.height < new_finalized_block.height && b.height > 0
     });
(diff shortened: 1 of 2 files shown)


light-base/src/runtime_service.rs
```diff
@@ -3217,9 +3217,13 @@ fn build_warp_sync_tree<TPlat: PlatformRef>(
...
     }
-    // Disable when missing, degenerately equal to the new finalized, or
-    // not strictly older.
-    .filter(|b| b.hash != new_finalized_block.hash && b.height < new_finalized_block.height);
+    // Disable when missing, degenerately equal to the new finalized, not strictly older,
+    // or pointing at genesis. Genesis state isn't reliably served by peers, so clients
+    // would fail storage/runtime calls against the Initialized hash. Relay falls back
+    // here when no checkpoint is set; parachain always falls back at first bootstrap.
+    .filter(|b| {
+        b.hash != new_finalized_block.hash && b.height < new_finalized_block.height && b.height > 0
```
lexnv: nit: Maybe we could place `b.height > 0` as first in the chain of checks

lexnv (approved):

light-base/src/runtime_service.rs
lrubasze: yes, that makes sense! Thanks :)

skunert (approved):

voliva:
I reported this here #3259

voliva:
I'm afraid whether this is hiding a bigger issue... If I understood it correctly, this change just waits for the `pre_warp_finalized` to reach any block whose height isn't 0?

The problem with #3259 though is that smoldot is somehow reporting that a block with height 17,###,### is a parent of block 0. This PR probably fixes this particular case, but I wonder whether there could be other cases where this could surface again. Wouldn't it be better to prevent any other case like that?

Historically, smoldot would sometimes send an `initialized` event with an ancient block, specially with adverse network conditions. Sometimes it would be the genesis block, sometimes it would be a sync-warp checkpoint.

However, as it fully synced with the rest of the network, it would send a `stop` event, notifying there was a discontinuity between that last announced block and the next one that it can't fulfill. The client would then restart a new follow subscription that starts with the correct block (or not, who knows). But most importantly, this behaviour is spec-compliant.

I also think that maybe this could break on environments for networks that produce blocks on-demand (so not every X seconds, but when e.g. there's a new transaction). Granted this is an edge case which probably happens only on test environments, but still... Correct if I'm wrong, but adding a condition "reject height=0" seems like a quick patch rather than a definitive fix.

bkchr:
@lrubasze instead of using this fake parent head in the notification, could we not just use the real parent hash of the block? We just use it to initialize the tree AFAIR? So, it should be fine and is the proper solution.

lrubasze:
Closing in favor of #3266

PR #3263, opened 2026-05-19
Implement `ext_trie_blake2_256_verify_proof` host functions

x3c41a:
## Background

`pallet-transaction-storage` (deployed on Polkadot Bulletin Chain, runs every block on paseo-bulletin) calls `ext_trie_blake2_256_verify_proof_version_2` from its `apply_block_inherents` inherent. Smoldot stubs both versions with `host_fn_not_implemented!()`, so chopsticks can't replay any block that includes the proof inherent — `run-block` panics on the runtime's `"Storage proof must be checked once in the block"` assertion.

Refs:
- paritytech/polkadot-bulletin-chain#341 — chopsticks issue
- paritytech/smoldot-archive#2221 — earlier attempt in the archived fork (used the regular Merkle proof decoder, which fails on compact proofs)

## Implementation

The verifier handles Substrate's **compact** proof format (`sp_trie::generate_trie_proof` → trie-db's `generate_proof`):

- Path children are replaced by an empty inline placeholder; the actual child node is the next proof entry.
- The target leaf's value is omitted; reconstructed from the caller-supplied expected value (hashed via blake2b for state version V1 values >= 32 bytes).

`verify_compact_trie_proof` in `trie::proof_decode` walks proof entries in order, descends into placeholders by consuming the next entry, and unwinds by re-encoding each node with reconstructed children. The final root node's hash must equal the caller-supplied root.

## Tests

Seven unit tests cover a real chain proof captured from `paseo-bulletin-rpc.polkadot.io` block #1503870: positive case + six rejection cases (wrong root / value / key / state version / tampered byte / truncated proof).

End-to-end: replaying six ProofChecked-emitting blocks (1503870, 1504069, 1504256, 1504290, 1504307, 1504375) through chopsticks `run-block` now finalizes cleanly. Previously all panicked.


lib/src/trie/proof_decode.rs
```diff
@@ -1792,3 +1792,174 @@ impl<'a, T> fmt::Binary for Children<'a, T> {
...
+}
+
+fn inject_compact_value(
+    expected_value: &[u8],
+    state_version: TrieEntryVersion,
+) -> CompactStorageValueOwned {
+    match state_version {
+        TrieEntryVersion::V0 => CompactStorageValueOwned::Unhashed(expected_value.to_vec()),
+        TrieEntryVersion::V1 => {
+            if expected_value.len() < 32 {
```
lexnv: There's a missmatch between this implementation and smoldot / substrate:
- values of length `> 32` should be hashed:

https://github.com/paritytech/smoldot/blob/cd8451d32de6eb948b63dae9256ca381e927b0e9/lib/src/trie/calculate_root.rs#L311-L313

lib/src/trie/proof_decode.rs
```diff
@@ -1792,3 +1792,174 @@ impl<'a, T> fmt::Binary for Children<'a, T> {
...
+        &key_nibbles[..],
+        &mut key_pos,
+        expected_value,
+        state_version,
+        true,
+    ) {
+        Some(m) => m,
+        None => return false,
+    };
+
```
lexnv, line 1837: We should double check after computing the merkle value that we have exhaused the proof entries.

In other words, we might call this function with an attacker crafter payload where the proof has some trailing entires: `[p0, p1, .., pn, pAttackVector1, pAttackVector2]`. In this case, smodot will consider the call successful. 


```suggestion
    if iter.next().is_some() {
        return false;
    }
```

lib/src/trie/proof_decode.rs
```diff
@@ -1792,3 +1792,174 @@ impl<'a, T> fmt::Binary for Children<'a, T> {
...
+        let child_index = u8::from(child_nibble) as usize;
+        let child = decoded.children[child_index]?;
+
+        // Compact encoding: an empty inline reference marks a path child whose subtree is
+        // the next proof entry; a sub-32-byte reference is the inlined child node itself;
+        // a 32-byte reference points outside the proven path, so descending into it means
+        // the proof can't prove this key.
+        let child_merkle = if child.is_empty() {
+            let next_node = proof_iter.next()?;
+            compact_proof_compute_merkle(
```
lexnv: We have seen in the past issues with exceeding the available stack, which then terminated the substrate node (cc @iulianbarbu).

Since the recursive depth is effectively a function influenced by the number of keys, which are controlled by the outside payload, we could:
- Use a stack to turn this algorithm from stack-based to heap-iteration-based
- Limit the depth of recursion by a sane upper limit (i think its bounded by maximum of 2 * num_keys)

lib/src/trie/proof_decode.rs
```diff
@@ -1792,3 +1792,174 @@ impl<'a, T> fmt::Binary for Children<'a, T> {
...
+    expected_value: &[u8],
+    state_version: TrieEntryVersion,
+) -> CompactStorageValueOwned {
+    match state_version {
+        TrieEntryVersion::V0 => CompactStorageValueOwned::Unhashed(expected_value.to_vec()),
+        TrieEntryVersion::V1 => {
+            if expected_value.len() < 32 {
+                CompactStorageValueOwned::Unhashed(expected_value.to_vec())
+            } else {
+                let h = blake2_rfc::blake2b::blake2b(32, &[], expected_value);
```
lexnv, line 2009: Lets also add a few more tets here, especially to capture the 32 len behavior 🙏

lib/src/trie/proof_decode.rs
bkchr: Are we doing the same in substrate?

lexnv: Yep: https://github.com/paritytech/trie/blob/69bfa688667dea1d8d35acc79b41082c415b2d20/trie-db/src/proof/verify.rs#L466-L468

```rust
					if proof_iter.next().is_some() {
						return Err(Error::ExtraneousNode)
					}
```

lib/src/trie/proof_decode.rs
lexnv: Impl example from substrate:  https://github.com/paritytech/trie/blob/69bfa688667dea1d8d35acc79b41082c415b2d20/trie-db/src/proof/verify.rs#L466-L468

lexnv:
@x3c41a To help make the review process a bit smoother, please use dedicated commits for fixes/ features / tests instead of force pushing 🙏 Otherwise, we'll need to double check previously reviewed work, thanks🙏

dr333ws:
> @x3c41a To help make the review process a bit smoother, please use dedicated commits for fixes/ features / tests instead of force pushing 🙏 Otherwise, we'll need to double check previously reviewed work, thanks🙏

sorry about that 😔 the force-push was comment edits only, no logic change. will keep dedicated commits going forward 🙏

lib/src/trie/proof_decode.rs
x3c41a: added

lib/src/trie/proof_decode.rs
x3c41a: good observation! changed to stack

lib/src/trie/proof_decode.rs
x3c41a: done in 67c5d68b

I also added `verify_compact_proof_rejects_extra_trailing_entry` which appends an extra entry to a valid proof and confirms it now rejects.

lib/src/trie/proof_decode.rs
```diff
@@ -1792,3 +1792,212 @@ impl<'a, T> fmt::Binary for Children<'a, T> {
...
+) -> Option<CompactMerkleValue> {
+    let mut stack: Vec<CompactFrame> = Vec::new();
+    let mut node_bytes: &[u8] = root_node_bytes;
+    let mut key_pos = 0usize;
+    let mut is_root = true;
+
+    let mut current = loop {
+        let decoded = trie_node::decode(node_bytes).ok()?;
+
+        let partial_key: Vec<nibble::Nibble> = decoded.partial_key.clone().collect();
```
lexnv, line 1888: nit: This allocates on the hot path. Could create an issue to takle this later since adding a lifetime on the `CompactFrame` looks quite involved atm

lexnv (approved):
LGTM! Thanks for contributing!

lib/src/trie/proof_decode.rs
x3c41a: tracked in #3275 for follow-up

lib/src/trie/proof_decode.rs
```diff
@@ -1792,3 +1792,212 @@ impl<'a, T> fmt::Binary for Children<'a, T> {
...
+            if key[key_pos] != *pk_nibble {
+                return None;
+            }
+            key_pos += 1;
+        }
+
+        let new_children: [Option<Vec<u8>>; 16] =
+            core::array::from_fn(|i| decoded.children[i].map(|c| c.to_vec()));
+
+        if key_pos == key.len() {
```
skunert, line 1902: One more thing here, substrate also reports ExtraneousValue if the value we are trying to verify is set. 

e.g. this test should be added and succeed:

```rust
#[test]
fn verify_compact_proof_rejects_present_leaf_value() {
    let mut proof = COMPACT_PROOF.to_vec();

    // COMPACT_PROOF is a SCALE-encoded Vec<Vec<u8>> with two trie-node entries.
    // At byte 74, the second entry starts with its SCALE compact length prefix:
    //
    //   0x08 0x40 0x00
    //
    //   0x08       = SCALE compact length 2 (2 << 2)
    //   0x40 0x00 = leaf node with an empty value placeholder
    //
    // In a compact proof, that empty value means "the real value was omitted and
    // must be supplied separately by the host-function argument".
    let leaf_entry_len_index = 74;
    assert_eq!(proof[leaf_entry_len_index], 0x08);
    assert_eq!(&proof[leaf_entry_len_index + 1..], &[0x40, 0x00]);

    // Change the second entry to:
    //
    //   0x0c 0x40 0x04 0xab
    //
    //   0x0c             = SCALE compact length 3 (3 << 2)
    //   0x40             = same leaf node header
    //   0x04             = SCALE compact value length 1 (1 << 2)
    //   0xab             = the one-byte value now present in the proof
    //
    // This turns the proof from "leaf exists, value omitted" into
    // "leaf exists, value is 0xab".
    // Substrate rejects this as ExtraneousValue, even though the verifier is
    // given the expected value separately.
    proof[leaf_entry_len_index] = 0x0c;
    proof[leaf_entry_len_index + 2] = 0x04;
    proof.push(0xab);

    assert!(!super::verify_compact_trie_proof(
        &proof,
        &COMPACT_PROOF_ROOT,
        COMPACT_PROOF_KEY,
        COMPACT_PROOF_VALUE,
        trie::TrieEntryVersion::V1,
    ));
}

```

Here a test against hte substrate implementation (should not be added in smoldot, but for reproduction purposes):
```rust
#[test]
fn substrate_verify_compact_proof_rejects_present_leaf_value() {
    let mut proof = COMPACT_PROOF.to_vec();

    // Same proof mutation as `verify_compact_proof_rejects_present_leaf_value`.
    let leaf_entry_len_index = 74;
    assert_eq!(proof[leaf_entry_len_index], 0x08);
    assert_eq!(&proof[leaf_entry_len_index + 1..], &[0x40, 0x00]);
    proof[leaf_entry_len_index] = 0x0c;
    proof[leaf_entry_len_index + 2] = 0x04;
    proof.push(0xab);

    let decoded_proof = super::decode_proof(&proof)
        .unwrap()
        .map(<[u8]>::to_vec)
        .collect::<Vec<_>>();
    let root = sp_core::H256::from(COMPACT_PROOF_ROOT);

    assert!(sp_trie::verify_trie_proof::<
        sp_trie::LayoutV1<sp_core::Blake2Hasher>,
        _,
        _,
        _,
    >(
        &root,
        &decoded_proof,
        &[(COMPACT_PROOF_KEY, Some(COMPACT_PROOF_VALUE))],
    )
    .is_err());
}
```

Otherwise looks good, fix this and we are ready to merge.

skunert: According to AI, this is the fix, but please double check, was short on time:

```rust
if key_pos == key.len() {
    let has_children = decoded.children.iter().any(Option::is_some);

    // Match trie-db's compact-proof verifier:
    // - leaf value must be omitted as an empty inline value
    // - branch value must be omitted entirely
    // - any present value/hash is ExtraneousValue, represented here as invalid proof
    match (has_children, decoded.storage_value) {
        (false, trie_node::StorageValue::Unhashed(v)) if v.is_empty() => {}
        (true, trie_node::StorageValue::None) => {}
        _ => return None,
    }

    let new_storage_value = inject_compact_value(expected_value, state_version);
    break compact_encode_node(partial_key, new_children, new_storage_value, is_root)?;
}
```

skunert: Why did you skip the children check? I think the difference is relevant.

Look at how leaf nodes skip values differently from branches:
- Leaf node encoding if omitted is inline empty value: https://github.com/paritytech/trie/blob/8fcd3911e5d2c66453f2770501a72b72203025d2/trie-db/src/proof/generate.rs#L91
- Branch node (we detect by looking at children) is encoded as `None`:  [here](https://github.com/paritytech/trie/blob/8fcd3911e5d2c66453f2770501a72b72203025d2/trie-db/src/proof/generate.rs#L83) and [here](https://github.com/paritytech/trie/blob/8fcd3911e5d2c66453f2770501a72b72203025d2/trie-db/src/proof/generate.rs#L112).

Or did you check and I am missing something?

x3c41a: That's what I was trying to understand yesterday evening too but decided to have a fresh look today. 

> Why did you skip the children check? I think the difference is relevant.

Root hash comparison catches it actually. Every node's hash covers its leaf/branch header too. So a swapped/malformed shape goes up to a root mismatch. 

Made the verifier reject malformed leaves/branches straight away instead of waiting for root-hash to catch it

skunert (approved):

skunert:
@x3c41a pls sign your commits

rockbmb (approved):
Tested by pointing the smoldot submodule in [AcalaNetwork/chopsticks#1037](https://github.com/AcalaNetwork/chopsticks/pull/1037) to this PR's tip (`970f4697`), rebuilding, and forking Paseo Bulletin Next (para 1501, `wss://paseo-bulletin-next-rpc.polkadot.io`). The Chopsticks PR adds a `SetTransactionStorageProof` inherent provider that Bulletin requires; the smoldot PR provides the host functions that Bulletin's user transactions need. Both are required together.

## Without this PR (smoldot `main` at `71438a0c`)

- Chopsticks `run-block` on a block containing `storeWithCidConfig`: crashes with `Host function not implemented: ext_trie_blake2_256_verify_proof_version_2`. The 3rd extrinsic (the `applyBlockInherents` inherent carrying a real proof) can't execute.
- Chopsticks `dev_newBlock` with a submitted `storeWithCidConfig`: the extrinsic triggers `ext_transaction_index_index_version_1` at runtime, which panics on `unreachable` (i64/i32 param mismatch). Chopsticks process crashes, WebSocket drops.
- Chopsticks `dev_newBlock` with empty blocks only: works, because neither broken code path is hit.

## With this PR (smoldot at `970f4697`)

- Chopsticks `run-block`: all 4 extrinsics (3 inherents + 1 signed `storeWithCidConfig`) execute without crashes.
- Chopsticks `dev_newBlock`: forked the chain, seeded test accounts, and dispatched 9 Bulletin-specific extrinsics through the tx pool: `store`, `storeWithCidConfig` (two CID configs), `forceRenew`, `authorizeAccount`, `authorizePreimage`, `refreshAccountAuthorization`, `refreshPreimageAuthorization`. All dispatched successfully with correct events, zero errors in logs.

dr333ws:
Apologies for force-pushing (again) but I had to do it to sign the commits. 
No logic change, only signing + rebasing onto main (had to resolve CHANGELOG conflicts)

PR #3264, opened 2026-05-19
`bitswap_unstable_stream` added

michalkucharczyk:
This PR adds `bitswap_unstable_stream` method defined in https://github.com/paritytech/json-rpc-interface-spec/pull/186

It also improves the e2e test for `bitswap_unstable_get` which was added in https://github.com/paritytech/smoldot/pull/3240


light-base/src/json_rpc_service/background.rs
```diff
@@ -1154,6 +1181,80 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
 
+                    methods::MethodCall::bitswap_unstable_stream { cids } => {
+                        log!(
+                            &me.platform,
+                            Debug,
+                            &me.log_target,
+                            format!("Bitswap stream subscription: {} cids", cids.len())
+                        );
+
+                        match me.bitswap_service.bitswap_stream(cids).await {
```
lrubasze: Not sure if I'm right but is client connection blocked while awaiting here?

michalkucharczyk: The await covers only the synchronous handoff (`parse_and_dedup` + a channel send + a oneshot for the `BatchId`), not the actual `Have` broadcast or per-CID events - those flow later via the events-pump task pushed to background_tasks at line 1220 (`light-base/src/json_rpc_service/background.rs`). 

The only backpressure could be the bitswap service's bounded inbound channel being saturated - I think this shall not happen.

Any ideas how to solve it in different way?

light-base/src/bitswap_service.rs
```diff
@@ -332,6 +606,181 @@ impl<TPlat: PlatformRef> BackgroundTask<TPlat> {
...
+            "sending cancel wantlist",
+            batch_id = batch_id.0,
+            cids = ?pending_cids,
+            peers = batch.peers_for_cancel.len()
+        );
+
+        // One Cancel wantlist message containing all pending CIDs, sent to every peer this
+        // batch's Have broadcast reached. Cancel for an unknown CID is harmless on the receiver
+        // side — the peer no-ops.
+        let message = build_bitswap_cancel_message(pending_cids.iter());
```
dmitry-markin, line 769: As a side note — substrate Bitswap server implementation never tracks pending want-lists (i.e., if it doesn't have the requested CID, it will just respond "NO" and never get back to it again). So, currently, it doesn't make sense to send cancel messages. While being mostly harmful (nit the bandwidth usage), we should probably:

- document it here and write that the message is sent only for conformance with IPFS Bitswap protocol implementation in case it is later updated on substrate's side;
- consider "subscribing" to not yet available chunks in substrate to handle cases when the publisher submitted the transaction, communicated to the receiver the CID, and the receiver need to await while it is available in the blocks. This is the edge case that needs to be confirmed before implementing in substrate to not introduce extra code requiring maintenance in case it is not needed by anyone.

light-base/src/bitswap_service.rs
```diff
@@ -332,6 +606,181 @@ impl<TPlat: PlatformRef> BackgroundTask<TPlat> {
...
+                Ok(()) => {}
+                Err(async_channel::TrySendError::Closed(_)) => {
+                    // Receiver dropped — JSON-RPC client disconnected or unsubscribed.
+                    // Cancel the rest of the batch and emit a Bitswap Cancel wantlist.
+                    should_cancel = true;
+                }
+                Err(async_channel::TrySendError::Full(_)) => {
+                    // Invariant: parse_and_dedup caps entries.len() at MAX_CIDS_PER_REQUEST
+                    // and events_tx is bounded(MAX_CIDS_PER_REQUEST). At most `total` items
+                    // are ever pushed for a batch, so the buffer cannot saturate.
```
dmitry-markin, line 664: May be worth adding a note / TODO item here that not backpressuring network responses here leads to allocation of (64 * 2 MiB) = 128 MiB max, which is acceptable in a desktop environment, but can cause issue on mobile.

light-base/src/bitswap_service.rs
```diff
@@ -163,6 +176,90 @@ impl BitswapService {
...
+    /// Subscribe to a stream of Bitswap blocks. On success, returns a [`BitswapStreamHandle`]
+    /// whose `events_rx` yields one `(cid_string, BlockResult)` event per input CID, in arrival
+    /// order (the order in which each CID resolves), not input order.
+    ///
+    /// Top-level errors: only the batch-input validation cases — `-32801 TooManyCids`,
+    /// `-32802 EmptyCids`, `-32803 DuplicateCids` — are surfaced at the top level. Wholesale
+    /// Have-broadcast failures (no peers connected / network send queue full) are NOT top-level
+    /// errors per the `bitswap_unstable_stream` spec: the subscription opens normally and the
+    /// failure fans out as one `streamItemError(-32812 FailRetryBackoff)` per remaining valid
+    /// CID, followed by `streamDone`.
```
dmitry-markin: Why it was chosen to not fail fast the entire subscription in case the have request failed, delivering individual errors instead?

light-base/src/bitswap_service.rs
```diff
@@ -441,56 +898,317 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
-                        timeout: timeout.clone(),
-                        stage: RequestStage::Have(have_peers),
-                        cid: cid.clone(),
-                    },
+                // Pre-resolve invalid-CID slots by pushing them straight into the events channel.
+                // try_send cannot fail here: events_tx is bounded(MAX_CIDS_PER_REQUEST) and the
+                // total events ever pushed for this batch is bounded by entries.len(); the
+                // receiver is also held by BitswapStreamHandle so the channel cannot be Closed.
+                for (cid_str, err) in invalid_slots {
+                    batch.pending_count -= 1;
```
dmitry-markin: nit: for uniformity with other places
```suggestion
                    batch.pending_count = batch.pending_count.saturating_sub(1);
```

dmitry-markin (approved):
The implementation looks correct and carefully written.

One thing I would do differently though, is apply the fail-fast principle and return an error to the subscription request in case the `Have` request failed, instead of faning-out errors for all the individual CIDs. May be there is some justification for not doing so?

light-base/src/json_rpc_service/background.rs
lrubasze: Getting the `batch_id` back isn't a quick handoff. In the normal case (≥1 valid CID), `ready_tx` is  moved into the broadcast future (bitswap_service.rs:1003-1013) and only sent at 1202, after `broadcast_bitswap_message().await` returns.
So `ready_rx.await` (line 213) waits for the whole "Have" broadcast, not just `parse_and_dedup` + the channel send.

Maybe we could mirror `bitswap_unstable_get` approach, which hands the await to a background task and finish the work in an event.
- Add another event variant:
```rust
BitswapStreamReady {
    request_id_json: String,
    result: Result<bitswap_service::BitswapStreamHandle, bitswap_service::BitswapGetError>,
},
```
- Replace the handler with something like this
```rust
methods::MethodCall::bitswap_unstable_stream { cids } => {
    log!(&me.platform, Debug, &me.log_target,
        format!("Bitswap stream subscription: {} cids", cids.len()));

    me.background_tasks.push({
        let bitswap_service = me.bitswap_service.clone();
        let request_id_json = request_id_json.to_owned();
        Box::pin(async move {
            let result = bitswap_service.bitswap_stream(cids).await;
            Event::BitswapStreamReady { request_id_json, result }
        })
    });
}
```
- Add the new event handler
```rust
WakeUpReason::Event(Event::BitswapStreamReady { request_id_json, result }) => {
    match result {
        Ok(handle) => {
            let subscription_id = {
                let mut sub_id = [0u8; 32];
                me.randomness.fill_bytes(&mut sub_id);
                bs58::encode(sub_id).into_string()
            };
            let events_rx = handle.events_rx.clone();
            let _prev = me.bitswap_subscriptions.insert(
                subscription_id.clone(),
                BitswapSubscription { _handle: handle },
            );
            debug_assert!(_prev.is_none());
            let _ = me.responses_tx.send(
                methods::Response::bitswap_unstable_stream(Cow::Borrowed(&subscription_id))
                    .to_json_response(&request_id_json),
            ).await;
            me.background_tasks.push(Box::pin(async move {
                let event = events_rx.recv().await.ok();
                Event::BitswapStreamEvent { subscription_id, event, events_rx }
            }));
        }
        Err(error) => {
            let _ = me.responses_tx.send(error.to_json_rpc_error(&request_id_json)).await;
        }
    }
  
```

light-base/src/bitswap_service.rs
michalkucharczyk: Best effort approach. We could still deliver some of the requested CIDs. The stream will contain errors for individual failed items or the data if fetch was successful.

dmitry-markin: I mean if the have request failed, we fan-out the errors for all CIDs anyway, so why not fail the original JSON-RPC call entirely so the caller can retry the entire subscription?

michalkucharczyk: Good point, we could go this way, fixed here: c312d76

light-base/src/json_rpc_service/background.rs
michalkucharczyk: @lrubasze thanks, makes sense. Done here: 6cdbbc6

light-base/src/bitswap_service.rs
michalkucharczyk: I would skip the note here, we already have backpressure on the todol [list](https://github.com/paritytech/smoldot/blob/47d579ca777f531022322d788b30f37f9127c4db/light-base/src/bitswap_service.rs#L36). OK for you?

light-base/src/bitswap_service.rs
michalkucharczyk: Small note added here: 8fdfb4d

lrubasze (approved):
Nice work here 👌

PR #3266, opened 2026-05-21
Warp-sync-aware runtime service for parachain bootstrap speedup

lrubasze:
## Summary

Reintroduces the goal of #3246 (skip the post-warp-sync wait in parachain bootstrap)

**Why:** the previous attempt let a pre-warp runtime download finish after warp sync had started, pinning a stale checkpoint hash in `subscribe_all` and breaking the parachain head lookup.

**How:** the sync service now tracks where warp sync is (not started, in progress, finished). Subscribers see that state at subscription time, and the runtime service uses it to hold off on runtime downloads while warp sync is mid-flight. Once warp sync settles (or AllForks makes enough progress on its own that warp sync is unnecessary), downloads resume and subscribers get a post-warp header.

With that guarantee in place, parachain bootstrap can use the subscription's initial finalized header directly instead of waiting for the next finalized block.

## Test plan
- Unit tests for the warp sync state machine transitions.
- [WIP] JSON-RPC compliance tests

### DIFF
--- a/light-base/src/sync_service/substrate_compat.rs
+++ b/light-base/src/sync_service/substrate_compat.rs
@@ -17,7 +17,7 @@
 
 use super::{
     BlockNotification, ConfigSubstrateCompatibleRuntimeCodeHint, FinalizedBlockRuntime,
-    Notification, SubscribeAll, ToBackground,
+    FinishedReason, Notification, SubscribeAll, ToBackground, WarpSyncState,
 };
 use crate::{log, network_service, platform::PlatformRef, util};
 
@@ -103,6 +103,8 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
             platform.sleep(Duration::from_secs(10)),
         ))
         .fuse(),
+        warp_sync_state: WarpSyncState::NotStarted,
+        all_forks_verifies_while_not_started: 0,
         all_notifications: Vec::<async_channel::Sender<Notification>>::new(),
         log_target,
         from_network_service: None,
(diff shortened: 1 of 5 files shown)


voliva:
I tried this branch in a local papi-console and seems to be working fine, the initial block is also not the genesis anymore.

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -422,6 +445,21 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
             WakeUpReason::SyncProcess(all::ProcessOne::VerifyBlock(verify)) => {
                 // Header to verify.
                 let verified_hash = verify.hash();
+
+                // AllForks-only progress near chain tip: warp sync is unnecessary.
+                let effects = task
+                    .warp_sync
+                    .on_event(WarpSyncEvent::AllForksBlockVerified);
```
lexnv, line 452: We received a block from a malicious peer, while calling `verify.verify_header(now)` we detect that the `VerifyOutcome::Error` is an erorr.

However, we have called the `on_event` to signal that the we have processed a block here.
This method will terminate the warp sync process when we reach `WARP_SYNC_NOT_NEEDED_AFTER_ALLFORKS_VERIFIES`.

Should we call `.on_event(WarpSyncEvent::AllForksBlockVerified)` and `force_resubscribe` inside the `all::HeaderVerifyOutcome::Success` after verifying the header? In other words, I believe its possible for a malicious node to craft some random blocks which will disable the warpsync exposing us to some long range / eclipse attacks

lrubasze:
Closing, as it is replaced by #3268. Apologies for the noise.

PR #3268, opened 2026-05-25
sync_service: upfront bootstrap mode decision

lrubasze:
## Summary

Reintroduces the goal of https://github.com/paritytech/smoldot/pull/3246

Coordinates `warp_sync` and `all_forks` at startup so RPC consumers receive an authoritative finalized block instead of a stale chain-spec checkpoint hash.

On the first `GrandpaNeighborPacket`, compute `gap = peer_finalized - local_finalized`:
- `gap >= 32` → **WarpAhead**: hold subscribers; once `WarpSyncFinished` fires, reply with the post-warp finalized header.
- `gap < 32` → **AllForksOnly**: cancel warp-sync, announce GrandPa state, reply to subscribers with the chain-spec finalized header.

If no peers reports within 30s deadline then fallback to **AllForksOnly**.

After the commit both paths converge: normal all-forks block-by-block following. The decision is one-shot per startup.

Parachain bootstrap consumes the SubscribeAll snapshot's finalized header directly — the mode-decision gate makes it authoritative.

## Test plan
- Added 'chainhead_v1_follow' conformance tests in CI

### DIFF
--- a/lib/src/sync/all.rs
+++ b/lib/src/sync/all.rs
@@ -224,8 +224,22 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
                 max_requests_per_block: config.max_requests_per_block,
                 block_number_bytes: config.block_number_bytes,
                 allow_unknown_consensus_engines: config.allow_unknown_consensus_engines,
+                warp_completion_suppressed: false,
             },
         }
+    }
+
+    /// Gates the final `all_forks` rebuild in [`AllSync::process_one`]. Warp's verification,
+    /// build, and [`AllSync::desired_requests`] keep running so callers can probe warp
+    /// readiness; a result completed while suppressed sits in [`Self::ready_to_transition`]
+    /// until lifted or dropped via [`AllSync::discard_pending_warp_completion`].
+    pub fn set_warp_completion_suppressed(&mut self, v: bool) {
+        self.shared.warp_completion_suppressed = v;
+    }
+
+    /// Drops any pending warp-sync completion before un-suppressing avoids a stale rebuild.
+    pub fn discard_pending_warp_completion(&mut self) {
+        self.ready_to_transition = None;
     }
 
     /// Returns the value that was initially passed in [`Config::block_number_bytes`].
@@ -401,14 +415,10 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
         all_forks.non_finalized_blocks_ancestry_order()
     }
 
-    /// Drops the warp-sync state machine. After this call:
-    ///
-    /// - No new warp-sync requests are produced by [`AllSync::desired_requests`].
-    /// - In-flight warp-sync requests are removed; their user data is returned to the
-    ///   caller (typically abort handles for the outer futures).
-    ///
-    /// No-op if called twice; returns an empty vector.
-    pub fn cancel_warp_sync(&mut self) -> Vec<TRq> {
+    /// Aborts all in-flight warp-sync requests, returning their user data (typically
+    /// `AbortHandle`s for the outer futures). `warp_sync` remains alive and may re-engage
+    /// later. No-op if warp-sync is not in use.
+    pub fn abort_in_flight_warp_requests(&mut self) -> Vec<TRq> {
         if self.warp_sync.is_none() {
             return Vec::new();
         }
@@ -421,19 +431,10 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
             .map(|(id, _)| RequestId(id))
             .collect();
 
-        let aborted_handles: Vec<TRq> = warp_request_ids
+        warp_request_ids
             .into_iter()
             .map(|id| self.remove_request(id))
-            .collect();
-
-        for (_, source_mapping) in self.shared.sources.iter_mut() {
-            source_mapping.warp_sync = None;
-        }
-
-        self.warp_sync = None;
-        self.ready_to_transition = None;
-
-        aborted_handles
+            .collect()
     }
 
     /// Returns true if it is believed that we are near the head of the chain.
@@ -1016,14 +1017,15 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
             }
         }
 
-        if let Some(RuntimeInformation {
-            finalized_runtime: finalized_block_runtime,
-            finalized_body,
-            finalized_storage_code,
-            finalized_storage_heap_pages,
-            finalized_storage_code_merkle_value,
-            finalized_storage_code_closest_ancestor_excluding,
-        }) = self.ready_to_transition.take()
+        if !self.shared.warp_completion_suppressed
+            && let Some(RuntimeInformation {
+                finalized_runtime: finalized_block_runtime,
+                finalized_body,
+                finalized_storage_code,
+                finalized_storage_heap_pages,
+                finalized_storage_code_merkle_value,
+                finalized_storage_code_closest_ancestor_excluding,
+            }) = self.ready_to_transition.take()
         {
             let (Some(all_forks), Some(warp_sync)) =
                 (self.all_forks.as_mut(), self.warp_sync.as_mut())
@@ -2527,6 +2529,8 @@ struct Shared<TRq, TSrc> {
     block_number_bytes: usize,
     /// Value passed through [`Config::allow_unknown_consensus_engines`].
     allow_unknown_consensus_engines: bool,
+    /// See [`AllSync::set_warp_completion_suppressed`].
+    warp_completion_suppressed: bool,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -2613,10 +2617,41 @@ mod tests {
     }
 
     #[test]
-    fn status_reports_sync_after_cancel_warp() {
+    fn abort_in_flight_warp_requests_keeps_warp_alive() {
         let mut sync = fresh_sync();
-        let _ = sync.cancel_warp_sync();
-        assert!(matches!(sync.status(), Status::Sync));
+        let aborted = sync.abort_in_flight_warp_requests();
+        assert!(aborted.is_empty());
+        // warp_sync remains alive; status still reflects it.
+        assert!(matches!(
+            sync.status(),
+            Status::WarpSyncFragments { source: None, .. }
+        ));
+    }
+
+    #[test]
+    fn desired_requests_emits_warp_regardless_of_suppression() {
+        let mut sync = fresh_sync();
+        let source_id = match sync.prepare_add_source(100, [1; 32]) {
+            AddSource::UnknownBestBlock(s) => s.add_source_and_insert_block((), ()),
+            _ => unreachable!(),
+        };
+        sync.update_source_finality_state(source_id, 100);
+
+        // Suppression only gates the final rebuild; warp's desired_requests must still
+        // emit so callers can probe whether warp would make progress.
+        sync.set_warp_completion_suppressed(true);
+        assert!(
+            sync.desired_requests()
+                .any(|(_, _, rq)| matches!(rq, DesiredRequest::WarpSync { .. }))
+        );
+    }
+
+    #[test]
+    fn discard_pending_warp_completion_is_noop_when_empty() {
+        let mut sync = fresh_sync();
+        assert!(sync.ready_to_transition.is_none());
+        sync.discard_pending_warp_completion();
+        assert!(sync.ready_to_transition.is_none());
     }
 
     #[test]
--- a/lib/src/sync/all.rs
+++ b/lib/src/sync/all.rs
@@ -2562,3 +2562,65 @@ fn all_forks_request_convert(
         request_justification: true,
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::chain::chain_information;
+
+    fn aura_grandpa_genesis() -> chain_information::ValidChainInformation {
+        chain_information::ValidChainInformation::try_from(chain_information::ChainInformation {
+            finalized_block_header: Box::new(header::Header {
+                parent_hash: [0; 32],
+                number: 0,
+                state_root: [0; 32],
+                extrinsics_root: [0; 32],
+                digest: header::Digest::from(header::DigestRef::empty()),
+            }),
+            consensus: chain_information::ChainInformationConsensus::Aura {
+                finalized_authorities_list: Vec::new(),
+                slot_duration: NonZero::new(6000).unwrap(),
+            },
+            finality: chain_information::ChainInformationFinality::Grandpa {
+                after_finalized_block_authorities_set_id: 0,
+                finalized_triggered_authorities: Vec::new(),
+                finalized_scheduled_change: None,
+            },
+        })
+        .unwrap()
+    }
+
+    fn fresh_sync() -> AllSync<(), (), ()> {
+        AllSync::new(Config {
+            chain_information: aura_grandpa_genesis(),
+            block_number_bytes: 4,
+            allow_unknown_consensus_engines: false,
+            sources_capacity: 4,
+            blocks_capacity: 4,
+            max_disjoint_headers: 4,
+            max_requests_per_block: NonZero::new(1).unwrap(),
+            download_ahead_blocks: NonZero::new(1).unwrap(),
+            download_bodies: false,
+            download_all_chain_information_storage_proofs: false,
+            code_trie_node_hint: None,
+        })
+    }
+
+    // Regression: AllSync::status() used to be a stub returning Status::Sync, hiding the active
+    // warp phase from callers (notably sync_service's warp_sync_can_proceed).
+    #[test]
+    fn status_reports_warp_for_fresh_grandpa_chain() {
+        let sync = fresh_sync();
+        assert!(matches!(
+            sync.status(),
+            Status::WarpSyncFragments { source: None, .. }
+        ));
+    }
+
+    #[test]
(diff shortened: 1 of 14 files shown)


light-base/src/sync_service/substrate_compat.rs
```diff
@@ -103,6 +115,12 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
             platform.sleep(Duration::from_secs(10)),
         ))
         .fuse(),
+        mode: ModeState::Deciding,
+        mode_decision_deadline: future::Either::Left(Box::pin(
```
skunert, line 122: nit: Intuitively, I would prever an Option here ofer this `Either` construct.

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -784,6 +845,45 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                // peer_finalized - local >= 32, so deciding on best leaves warp_sync
+                // stalled waiting for finality to catch up (~6s/block on Polkadot).
+                if matches!(task.mode, ModeState::Deciding) {
+                    let local_finalized = task
+                        .sync
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!())
+                        .finalized_block_number();
+                    let gap = finalized_block_height.saturating_sub(local_finalized);
+                    if gap >= MODE_DECISION_WARP_GAP_THRESHOLD {
```
skunert: This can starve SubscribeAll. Once one peer moves us to AwaitingWarp, the decision timer is disabled and later below-threshold peers cannot make us fall back to all-forks. If the triggering peer disconnects, and remaining peers are below warp_sync_minimum_gap, warp_sync.desired_requests() emits no warp request, WarpSyncFinished never fires, and pending subscriptions stay queued indefinitely. We need an AwaitingWarp fallback/redecision path, ideally driven by AllSync/warp_sync readiness rather than duplicating the gap check here.


Also I think we should unify this logic and ask warp_sync maybe whether it has requests inflight or knows peers which have the `warp_sync_minimum_gap` correct. Did not verify this part, but having this constant here in addition to the warp sync one smells.

skunert:
I think what you propose here can work, but it changes the semantics slightly. 

Before our changes to Smoldot, the following would happen:
- Relay side starts up and starts syncing from LightSyncState: both AllForks and WarpSync run in parallel. 
- AllForks  and WarpSync basically "race". Warp requests are handled first always and make much faster progress, so assuming the snapshot is not super close to the tip, warp sync will eventually finish.
- To bridge the gap between initialized event and newly warped to block, we send stop event and everyone needs to resubscribe
- Internally, parachain side is also listening to relay subscription and needs to decide when to request runtimes from the relay chain. The signal before our PRs was a finalized notification, because these indicate that we are close to the tip and are getting actual grandpa commit messages.

Subscriptions would send the initialized event right after runtime download and and additional stop event once warp sync is finished.

So I would say the "race" between AllForks and WarpSync was by design, parachain waited for Finalized event because that indicated that no larger gap will be jumped soon (at least its an indicator). Maybe this is known, but my mental model only completed now about this.


Json-RPC spec-wise, the changes here should be compliant. [Spec says](https://paritytech.github.io/json-rpc-interface-spec/api/chainHead_v1_follow.html#the-withruntime-parameter):

```
Note: It is unlikely that high-level UIs built on top of a JSON-RPC client can do anything before the JSON-RPC server has access to the runtime. Consequently, they should consider the time before the initialized event is generated as a loading time. During this loading time, the JSON-RPC server might be performing downloads and CPU-intensive operations. This loading time can be considered as a replacement for the isSyncing field of the legacy system_health JSON-RPC call.
```

So to me it sounds like as an implementor we have freedom to decide when loading is finished and we send the initialized event. That might as well be once we finish warp syncing.

light-base/src/sync_service/substrate_compat.rs
lrubasze: I wanted to keep consistency with other timer used in `smoldot`.
eg. 
https://github.com/paritytech/smoldot/blob/8feb1326a684e5c8f485f6f79280db83e5e0fa07/light-base/src/sync_service/substrate_compat.rs#L114

If we change, let's change both. WDYT?

light-base/src/sync_service/substrate_compat.rs
lrubasze: Tried to address the above concerns in follow-up commits

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1489,3 +1630,104 @@ impl<TPlat: PlatformRef> Task<TPlat> {
...
+        non_finalized_blocks_ancestry_order,
+        new_blocks,
+    });
+}
+
+/// Warp is downloading/building, or a qualifying source is available for the next request.
+fn warp_sync_can_proceed(
+    sync: &all::AllSync<future::AbortHandle, (libp2p::PeerId, codec::Role), ()>,
+) -> bool {
+    match sync.status() {
```
lexnv, line 1778: Since the `sync.status` always returns `Status::Sync` we are effectively short-circuting the first 2 branches for now:

https://github.com/paritytech/smoldot/blob/a12be0447d7bf215f1639a369adaefa793fa541a/lib/src/sync/all.rs#L246-L251


Then we'd evaluate if any `desired_requests` have a `DesiredRequest::WarpSync`:
- A warp sync in-flight fragment download gates the `WarpSync::desired_requests` and returns `iter::empty()`
- We could also have only `StorageGetMerkleProof` and `RuntimeCallMerkleProof` requests inflight 

In those cases, `warp_sync_can_proceed` returns `false` although the warp sync is still healthy. Then, we'd rely on the 30s timeout to detect stalls. Then we cancel warp:

https://github.com/paritytech/smoldot/blob/a12be0447d7bf215f1639a369adaefa793fa541a/light-base/src/sync_service/substrate_compat.rs#L1518

Dont believe we should treat download in progress as starved here. Instead, we can properly clean up the `sync.status` however not sure how involved that might be

light-base/src/sync_service/substrate_compat.rs
skunert: If its a pattern that is well-established in the codebase, I am fine with keeping it as is.

light-base/src/runtime_service.rs
```diff
@@ -1412,71 +1412,64 @@ async fn run_background<TPlat: PlatformRef>(
...
-                            "warp-sync-tree-init",
-                            pre_warp_finalized_hash = if let Some(h) = pre_warp_finalized_hash {
-                                Cow::Owned(HashDisplay(&h).to_string())
-                            } else {
-                                Cow::Borrowed("<none>")
-                            },
-                            finalized_block_hash = HashDisplay(&finalized_block_hash),
-                        );
-
                         background.tree = Tree::FinalizedBlockRuntimeKnown {
```
lexnv, line 1417: nit: We have the above comment "// but the sync service guarantees that `subscribe_all` returns very quickly." We are sitllwaiting a bit for the 30s timer / warp to finish. Could have this comment changed in other places as well if present

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -784,6 +840,45 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
                     .update_source_finality_state(sync_source_id, finalized_block_height);
+
+                if matches!(task.mode, ModeState::Deciding) {
+                    let local_finalized = task
+                        .sync
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!())
+                        .finalized_block_number();
+                    let gap = finalized_block_height.saturating_sub(local_finalized);
+                    if warp_sync_can_proceed(task.sync.as_ref().unwrap_or_else(|| unreachable!())) {
```
lexnv: nit: Could the first peer force us into a slower block by block sync from the checkpoint towards the tip?

If the first peer sends out a `GrandpaNeighborPacket` that is within <32 blocks from our local tip, however the next packet would appropriately move us into warp. 

- We could wait for a few packets here and decide on the maximum reported finalized across the packets 
- Alternatiely, not sure if we can move back to warpsync after commiting to only forks

lib/src/sync/all.rs
```diff
@@ -407,6 +407,44 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
...
+            .requests
+            .iter()
+            .filter(|(_, rq)| rq.warp_sync.is_some())
+            .map(|(id, _)| id)
+            .collect();
+
+        let mut aborted_handles = Vec::with_capacity(warp_request_ids.len());
+        for id in warp_request_ids {
+            let request = self.shared.requests.remove(id);
+            debug_assert!(request.all_forks.is_none());
```
lexnv: nit: This might be leaking all fork requests since we don't expect it to happen (ie donwload_bodies is false, so we can't have a block request with `Headers | Body` flags set at the same time).

This might change in the future:

```rust

if let Some(id)  = request.all_forks {
    self.all_forks.finish_request(id);
}

File: light-base/src/sync_service/substrate_compat.rs [hunk 7]
```diff
@@ -41,6 +43,11 @@ use smoldot::{
     sync::all,
 };
 
+/// Maximum wait for the first GrandpaNeighborPacket before falling back to AllForksOnly.
+/// Sized for cold-start peer discovery (DNS + libp2p handshake + gossip-open), which can
+/// stretch to ~20s on light clients.
+const MODE_DECISION_TIMEOUT: Duration = Duration::from_secs(30);
```
lexnv, line 51: Could we add a smaller timeout for testing builds? Then add a test that ensures this fires during the warp process? This might have captured the `status` incomplete path

light-base/src/sync_service/substrate_compat.rs
lexnv: Same here, would be good to have a test to double check

light-base/src/sync_service/substrate_compat.rs
lexnv: In the `futures` crate we have an [OptionalFuture](https://docs.rs/futures/latest/futures/future/struct.OptionFuture.html) however that returns instantly IIRC. Ideally we'd have a custom 

```rust
fn poll {
   match self.inner {
        Option(future) => poll future
        None => return poll::Pending
   }
}
```

But yeap lets keep parity here with prior patterns 🙏

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1409,10 +1482,69 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                    } else {
+                        task.mode_decision_deadline =
+                            future::Either::Right(future::pending()).fuse();
+                        log!(
+                            &task.platform,
+                            Debug,
+                            &task.log_target,
+                            "mode-decision; committed=AllForksOnly (warp starved)",
+                        );
+                        commit_all_forks_only(&mut task);
```
skunert: I think in this branch we should go back to `ModeState::Deciding`.

If we are way behind the tip, it does not make sense to me to commit to all_forks_only, even if we can currently not make warp progress.

lib/src/sync/all.rs
```diff
@@ -407,6 +407,44 @@ impl<TRq, TSrc, TBl> AllSync<TRq, TSrc, TBl> {
...
+            self.shared.sources[request.source_id.0].num_requests -= 1;
+            aborted_handles.push(request.user_data);
+        }
+
+        for (_, source_mapping) in self.shared.sources.iter_mut() {
+            source_mapping.warp_sync = None;
+        }
+
+        self.warp_sync = None;
```
skunert: Are we sure that we are guarded against the panics encountered last time? Setting `warp_sync` to None was previously attempted in #3155 and then we looked into fixes like #3185 . We can have it, but need to make sure we don't assert on warp sync being present elsewhere.

light-base/src/runtime_service.rs
michalkucharczyk: please explain why unwrap is safe, e.g. use "qed" pattern:
```suggestion
                                            .expect("decoding always works, header cannot be malformed. qed")
```

michalkucharczyk: same here - why unwrap is safe here? (use expect + qed).

lib/src/sync/all.rs
lrubasze: Good catch! Done

light-base/src/runtime_service.rs
lrubasze: Comment updated

light-base/src/sync_service/substrate_compat.rs
lrubasze: Adding such test is quite significant change, lets do it in a separate PR.
Added TODO to not forget.

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -784,6 +840,45 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
                     .as_mut()
                     .unwrap_or_else(|| unreachable!())
                     .update_source_finality_state(sync_source_id, finalized_block_height);
+
+                if matches!(task.mode, ModeState::Deciding) {
+                    let local_finalized = task
+                        .sync
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!())
```
michalkucharczyk, line 876: nit: I think simple expect would be more readable here.  Any reason for using "unreachable pattern" instead of simply expect?

michalkucharczyk: hm, maybe expect requires formatting code and some string embedded into lib?

light-base/src/runtime_service.rs
lrubasze: Done

lrubasze: Done

light-base/src/sync_service/substrate_compat.rs
lrubasze: Seems so

light-base/src/sync_service/substrate_compat.rs
lrubasze: Done

lib/src/sync/all.rs
lrubasze: I remembered this infamous rollback of #3155. Did my best to make sure no such situation occurs again.

light-base/src/sync_service/substrate_compat.rs
lrubasze: Added a constant to specify how many packets to wait, equals 2 atm.
The thing is that it delays the mode decision moment.
By 700-1000ms per additional packet (at least for Paseo network), delaying the time to finalized event.

light-base/src/sync_service/substrate_compat.rs
lrubasze: What a miss from my side... 🙈 
Updated `status()` to return proper `Status`.
Also added some tests to cover some concerning points.

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -406,14 +446,20 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
                                 }
                             )
                         );
+                        // `BlockNumberNotIncrementing` means a finality proof overtook the
+                        // fragment mid-flight; the peer isn't at fault.
+                        let peer_at_fault =
+                            !matches!(err, all::VerifyFragmentError::BlockNumberNotIncrementing);
```
lexnv, line 468: Could it still be possible that peers just send this maliciously? Then we'd have no way to disconnect / ban them

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1409,10 +1513,75 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                        log!(
+                            &task.platform,
+                            Debug,
+                            &task.log_target,
+                            "mode-decision; awaiting-warp deadline re-armed",
+                        );
+                    } else {
+                        // Warp starved: drop back to Deciding so a future warp-eligible peer
+                        // can re-trigger CommitWarpAhead instead of locking in AllForksOnly.
+                        task.mode = ModeState::Deciding;
```
lexnv, line 1572: Could we still block the `subscribe_all` here if no peers can provide a fragment with `decoded.is_finished = true`?

lib/src/sync/all.rs
skunert: This is the main gripe currently IMO. To me its unclear if there is any benefit to setting warp sync to None. If for any reason we later fall behind the tip by 32 blocks, warp sync would start requests again before these changes. With this we now assume that we will never need warp sync again for the lifecycle of this instance. Maybe its not a problem, but its not obvious to me.

The way I see it the main reason why you do this is to prevent warp synced result to "disrupt" subscriptions by resetting them. The same could be achieved by ignoring warp sync results that are below the current finalized block. Take it with a grain of salt, my reasoning might be off.

lrubasze: Honestly, I didn't realize warp sync could be useful later on after syncing to the tip! 
Tried to refactor it a bit, so it can resume after bootstrap phase.

skunert (approved):

lexnv (approved):
LGTM! Nice job here @lrubasze 🚀

light-base/src/sync_service/substrate_compat.rs
lrubasze: added TODO to address in a follow-up PR

light-base/src/sync_service/substrate_compat.rs
lrubasze: added TODO to address in a follow-up PR

PR #3269, opened 2026-05-25
network_service: shorten handshake/ban timeouts to speed peer discovery

lrubasze:
Tightens three timeouts in `light-base/src/network_service.rs` that  dominated peer-discovery latency on cold and warm restarts:

- `handshake_timeout`: 8s → 4s.
A handshake either completes quickly or it is not going to complete; 8s mostly served as a slot-hold for dead peers.
- Pre-handshake ban (dial / connection-attempt failure): 5s → 2s.
Many parallel dials time out before any handshake completes, and a 5s ban keeps peer slots tied up across the whole discovery burst.  Banning only briefly is enough to avoid immediate retry storms.
- Gossip-open-failed ban: 15s → 3s. 
Transient gossip failures should not lock out otherwise healthy peers for 15s — that is long enough to materially delay reaching usable peer count after a restart.

Post-handshake failures keep the 5s ban (the `handshake_finished` branch), since at that point we have stronger evidence the peer is misbehaving rather than just unreachable.

NOTE!
Values have been adjusted arbitrarily basing on the observations from `smoldot` startup times.

### DIFF
--- a/light-base/src/network_service.rs
+++ b/light-base/src/network_service.rs
@@ -2511,7 +2511,10 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
                     peer_id,
                     ?error,
                 );
-                let ban_duration = Duration::from_secs(3);
+                // Must exceed polkadot-sdk's 5s notification-reject ban; otherwise we retry
+                // into a still-active remote ban. 0.5s margin covers network delay and
+                // clock skew between the two sides' ban timers.
+                let ban_duration = Duration::from_millis(5500);
 
                 // Note that peer doesn't necessarily have an out slot, as this event might happen
                 // as a result of an inbound gossip connection.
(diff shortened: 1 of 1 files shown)


lexnv (approved):

light-base/src/network_service.rs
```diff
@@ -2504,7 +2511,7 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
                     peer_id,
                     ?error,
                 );
-                let ban_duration = Duration::from_secs(15);
+                let ban_duration = Duration::from_secs(3);
```
lexnv: nit:

One thing worth considering: for notification protocols, when Substrate rejects a substream, the endpoint is banned for 5s (ie: polkadot-sdk rejects a connection to us, then bans smoldot for 5s).
- then we should avoid retrying connections at 3s pace intervals, when the banning is in place for 5s.

This is a bit problematic even today, but without insigts into notif-protocol rejection reason it's hard to quantify:
- assume Node A is banned at T0+0ms 
- Node B detects the rejection from Node A (without knowing the reason) and bans on its side Node A at T0+10ms (network delay)
- Then, 5s later, T0+5s Node A tries to establish a connection, Node B bans directly because T0+10ms+Ts ban is still in place.

skunert (approved):

light-base/src/network_service.rs
lrubasze: Ok, I will change it to 5.5s for now.

PR #3273, opened 2026-05-27
network_service: prefer bootnodes for gossip slots until a chain connects

lrubasze:
On warm rest, the relay out-slots get filled from the restored peer book, which is full of DHT-discovered nodes advertising unreachable private/link-local endpoints. They time out and churn the slots for many seconds, delaying the first relay gossip.

This is not the case for cold start, where relay out-slots get filled from the chain-spec bootnodes.

Fix: Attempt to mimic cold approach by picking bootnodes if possible while a chain has no open gossip link.

### DIFF
--- a/lib/src/network/basic_peering_strategy.rs
+++ b/lib/src/network/basic_peering_strategy.rs
@@ -462,6 +462,18 @@ where
         chain: &TChainId,
         now: &TInstant,
     ) -> AssignablePeer<'_, TInstant> {
+        self.pick_assignable_peer_filtered(chain, now, |_| true)
+    }
+
+    /// Same as [`BasicPeeringStrategy::pick_assignable_peer`], except that only peers for which
+    /// `filter` returns `true` are taken into account. The [`AssignablePeer::AllPeersBanned`] and
+    /// [`AssignablePeer::NoPeer`] return values thus concern only the filtered subset of peers.
+    pub fn pick_assignable_peer_filtered(
+        &mut self,
+        chain: &TChainId,
+        now: &TInstant,
+        mut filter: impl FnMut(&PeerId) -> bool,
+    ) -> AssignablePeer<'_, TInstant> {
         let Some(&chain_index) = self.chains_indices.get(chain) else {
             return AssignablePeer::NoPeer;
         };
@@ -478,6 +490,7 @@ where
                         usize::MAX,
                     ),
             )
+            .filter(|(_, _, peer_id_index)| filter(&self.peer_ids[*peer_id_index]))
             .choose(&mut self.randomness)
         {
             return AssignablePeer::Assignable(&self.peer_ids[*peer_id_index]);
@@ -495,6 +508,7 @@ where
                 )),
                 ops::Bound::Excluded((chain_index, PeerChainState::Slot, usize::MIN)),
             ))
+            .filter(|(_, _, peer_id_index)| filter(&self.peer_ids[*peer_id_index]))
             .next()
         {
             let PeerChainState::Banned { expires } = state else {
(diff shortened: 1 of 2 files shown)


lib/src/network/basic_peering_strategy.rs
```diff
@@ -461,6 +461,18 @@ where
...
+    }
+
+    /// Same as [`BasicPeeringStrategy::pick_assignable_peer`], except that only peers for which
+    /// `filter` returns `true` are taken into account. The [`AssignablePeer::AllPeersBanned`] and
+    /// [`AssignablePeer::NoPeer`] return values thus concern only the filtered subset of peers.
+    pub fn pick_assignable_peer_filtered(
+        &mut self,
+        chain: &TChainId,
+        now: &TInstant,
+        mut filter: impl FnMut(&PeerId) -> bool,
```
lexnv: nit: `filter_fn` to avoid confusion with `.filter()` on the range

lrubasze: Good catch! Done

light-base/src/network_service.rs
```diff
@@ -1290,10 +1290,30 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                            let has_open_gossip_link = task
+                                .open_gossip_links
+                                .keys()
+                                .any(|(link_chain_id, _)| *link_chain_id == chain_id);
+                            if !has_open_gossip_link {
+                                if let basic_peering_strategy::AssignablePeer::Assignable(peer_id) =
+                                    task.peering_strategy.pick_assignable_peer_filtered(
+                                        &chain_id,
+                                        &now,
+                                        |peer_id| task.important_nodes.contains(peer_id),
```
lexnv: nit: This subtly changes the meaning behind `important_nodes` whichwere considered important for "logging purposes", now we'll use those peers until we get an open gossip link. 

Could change the doc to reflect this. At the same time, maybe we could refactor them into `important: HashMap<chain ID, HashSet< peer ID>>`

light-base/src/network_service.rs
```diff
@@ -1290,10 +1290,30 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
                             }
 
-                            match task
-                                .peering_strategy
-                                .pick_assignable_peer(&chain_id, &task.platform.now())
-                            {
+                            let now = task.platform.now();
+
+                            // Until the chain has an open gossip link, prefer a bootnode (curated
+                            // and reachable); afterwards fill slots from the general pool.
```
lexnv: nit: We don't know if they are either curated or reachable :D

light-base/src/network_service.rs
```diff
@@ -1290,10 +1290,30 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
...
+                            let has_open_gossip_link = task
+                                .open_gossip_links
+                                .keys()
```
lexnv: We are walking every key and every link in the `open_gossip_links` to find the current chain. We do this for every chain registered. This is `O(total links * chains)`.

The meaning behind the comment and our check here actually diverges a bit "Until the chain has an open gossip link". We are checking live links. In other words, this is not a one-time setup that we do on startup, we'll always prefer bootnodes when we drop links.

To tackle this we could:
- introduce a `task.chains_ever_gossip_connected: HashSet<ChainId>` 
- track the chain ID once we `task.open_gossip_link` in `GossipConnected`
- this check will be an amortized `O(1)` and we'd only prefer bootnodes initially, not every time we drop links: `if !task.chains_ever_gossip_connected.contains(&chain_id)`

lexnv (approved):
LGTM! 🚀

light-base/src/network_service.rs
lrubasze: yeah, good point :) Corrected

light-base/src/network_service.rs
lrubasze: That's a nice catch. I like the idea!

light-base/src/network_service.rs
lrubasze: Done

skunert (approved):

PR #3274, opened 2026-05-27
light-base: opt-in optimistic parachain bootstrap with gated reads

BigTava:
## Summary

- Bootstrap from most recent finalized block instead of next finalized blcok.
- Introduces optimistic parachain bootstrap. Example config:

```
client.addChain({
  chainSpec,
  potentialRelayChains: [relayChain],
  optimisticParachainBootstrap: {
    allowedStoragePrefixes: [
      concat(twox128("Revive"), twox128("AccountInfoOf")),
      new TextEncoder().encode(":child_storage:default:"),
    ],
  },
})
```
- Fixes `build_child_storage_proof_request`, which sent the bare child trie id where full nodes require the prefixed `:child_storage:default:<id>` key.

### DIFF
--- a/.config/smoldot.dic
+++ b/.config/smoldot.dic
@@ -1,4 +1,4 @@
-153
+155
 ABI
 ACK
 API
@@ -68,6 +68,7 @@ Westend
 Yamux
 aborter
 allocator
+allowlist/S
 amongst
 authorable
 backend/S
@@ -120,6 +121,7 @@ noncomprehending
 opcode
 orthogonally
 parachain/S
+parahead/S
 parathread/S
 parsers
 pluggable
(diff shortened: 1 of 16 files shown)


bkchr (changes requested):
Can you explain your reasoning behind this pr?

BigTava:
@bkchr thanks for taking a look. 

Smoldot waits for the relay to finalize a new block before it can derive the parachain head. As a first approach I attempted to use the relay best block instead of waiting. To mitigate the risk of products maliciously taking advantage of this during the bootstrap window, I added a JSON-RPC gate, so that callers can only read from a small allowlist of storage prefixes. This approach breaks the never-reverts-finalized rule. Therefore I switched to the relay current finalized block (already in `subscription.finalized_block_scale_encoded_header` after warp sync). I kept the bootstrapping gate, assuming it could still be misused. 

Fortunately, [#3268](https://github.com/paritytech/smoldot/pull/3268) already solved this with a much more elegant implementation, so this PR is ready to be closed. The only useful contribution is the child-trie support for `chainHead_v1_storage` and the codec fix in `build_child_storage_proof_request`, which can be addressed as follow-up PRs.

PR #3281, opened 2026-06-05
statement-store: index subscriptions by topic for efficient statement matching

### LINKED ISSUE #3144: statement-store: Processing statements received from network is inefficient on a big number of statements
TODO: This will be inneficient if number of statement_subscriptions is big. I assume it shouldn't.

_Originally posted by @alexggh in https://github.com/paritytech/smoldot/pull/3127#discussion_r2959219178_

https://github.com/paritytech/smoldot/blob/6b792e872e035f981b0ea64b2fc29beeb6181802/light-base/src/json_rpc_service/background.rs#L753-L754

### LINKED ISSUE #3147: statement-store: create a reverse index `topic` -> `subscription`
We need to create a reverse index `topic` -> `subscription` for adequate complexity. Even ruling out topics we are not interested in at all (none of the filters have them) can speedup things dramatically, and upon match we can do a more rigorous topic/statement match inspection. There are likely even better ways out there.

Currently it's `n_statements` x `n_filters` x `n_topics_in_filter` x `n_topics_in_statement` ...

We can merge for now, but should definitely address this in a follow up. Synchronous processing here can also stall the processing of other RPC methods..

_Originally posted by @dmitry-markin in https://github.com/paritytech/smoldot/pull/3127#discussion_r3001521786_

AndreiEres:
Closes #3144
Closes #3147

Statements from the network were matched against every active subscription, giving `O(n_statements × n_subscriptions × topic_factors)` complexity — a node re-scanned all subscriptions even for statements no one filtered on.

Add a reverse `topic -> subscriptions` index (plus a `wildcard` set for `Any` and empty `MatchAll`, which match everything). For each statement, matching now considers only the wildcard subscriptions and those sharing one of its topics, then applies the exact per-subscription filter and dedup via `accept`. Filter semantics are unchanged; the index only narrows the candidate set.

### DIFF
--- a/light-base/src/json_rpc_service/statement.rs
+++ b/light-base/src/json_rpc_service/statement.rs
@@ -478,10 +478,16 @@ mod tests {
         ]);
 
         let matches = subs.matching(&[batch_entry(0x01, vec![[7u8; 32]])]);
-        assert_eq!(matched_ids(&matches), vec!["all".to_string(), "any".to_string()]);
+        assert_eq!(
+            matched_ids(&matches),
+            vec!["all".to_string(), "any".to_string()]
+        );
 
         let matches = subs.matching(&[batch_entry(0x02, vec![])]);
-        assert_eq!(matched_ids(&matches), vec!["all".to_string(), "any".to_string()]);
+        assert_eq!(
+            matched_ids(&matches),
+            vec!["all".to_string(), "any".to_string()]
+        );
     }
 
     #[test]
--- a/light-base/src/json_rpc_service/statement.rs
+++ b/light-base/src/json_rpc_service/statement.rs
@@ -220,7 +220,7 @@ impl StatementSubscriptions {
             for id in &candidates {
                 let sub = subscriptions
                     .get_mut(*id)
-                    .expect("by_topic/wildcard only reference live subscriptions; qed");
+                    .expect("`candidates` is a subset of `subscriptions`; qed");
                 if sub.accept(hash, statement) {
                     let encoded = encoded.get_or_insert_with(|| {
                         HexString(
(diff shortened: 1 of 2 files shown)


light-base/src/json_rpc_service/statement.rs
```diff
@@ -78,28 +78,189 @@ impl StatementSubscription {
...
+        id: String,
+        topic_filter: TopicFilter,
+        max_seen: Option<NonZero<usize>>,
+    ) {
+        match &topic_filter {
+            TopicFilter::Any => {
+                self.wildcard.insert(id.clone());
+            }
+            // An empty `MatchAll` filter matches every statement.
+            TopicFilter::MatchAll(topics) if topics.is_empty() => {
```
alexggh, line 131: I wonder if we should just reject this one.

AndreiEres: why?

alexggh: because, feels like an unintended usage of the API, I'm fine with keep it as it is.

alexggh (approved):

light-base/src/json_rpc_service/statement.rs
```diff
@@ -78,28 +78,189 @@ impl StatementSubscription {
...
+                    candidates.extend(ids.iter().map(String::as_str));
+                }
+            }
+
+            // Re-encoded lazily on first match and reused for every matching subscription.
+            let mut encoded: Option<HexString> = None;
+            for id in &candidates {
+                let sub = subscriptions
+                    .get_mut(*id)
+                    .expect("by_topic/wildcard only reference live subscriptions; qed");
```
dmitry-markin: nit:
```suggestion
                    .expect("`candidates` is a subset of `subscriptions`; qed");
```

dmitry-markin (approved):

PR #3282, opened 2026-06-09
Report full finalized ancestry in chainHead Finalized events

lrubasze:
### Motivation

The `sync_service` `Finalized` notification carried only the highest finalized block hash, so when several blocks finalize at once `chainHead_v1_follow` emitted events that skipped the intermediate ones.

This wasn't strictly a spec violation (the spec only pins the last element + the `newBlock`-first rule), but the stream was lossy - consumers tracking every finalized block missed the skipped ones. This PR makes it complete and contiguous. While testing against the relay chain, a few related bugs surfaced and were fixed here too.

### Changes

1. Full finalized ancestry (original fix): `Notification::Finalized` now carries `finalized_blocks_hashes` - all newly-finalized blocks in ascending order, never empty - forwarded whole by the JSON-RPC layer. All were already reported via `newBlock`, so it stays spec-compliant.
2. Don't discard a finished warp sync: a warp result finishing while the sync mode was still being decided got parked, then thrown away when all-forks was picked. It's now kept.
3. Yield the first pending set-change block in `finality_checkpoints`: it was excluded, so all-forks requested a justification it couldn't verify yet (TooFarAhead), banned the peer, and finality starved across the set change.
4. Drop a full-buffer subscriber instead of silently skipping its notification (which gapped the stream); its channel is now closed per the `subscribe_all` contract.

### Testing
- New `chainHead_v1_follow` e2e coverage against the relay chain (relay-only and `with_runtime=false`); the relay-only test reproduces the stall and now passes.
- Multi-block finalization needed elastic scaling, hence new snapshots. Every relay validator is snapshotted so availability recovery can reconstruct in-flight candidates after restore.

### DIFF
--- a/e2e-tests/tests/smoke_generate_snapshots.rs
+++ b/e2e-tests/tests/smoke_generate_snapshots.rs
@@ -37,12 +37,12 @@ use std::path::{Path, PathBuf};
 use anyhow::anyhow;
 use serde_json::Value;
 use smoldot_e2e_tests::{
-    ensure_js_deps_installed, ensure_smoldot_built, resolve_base_dir, run_js_test,
-    FINALIZED_METRIC, PARA_ID,
+    elastic_scaling_genesis_overrides, ensure_js_deps_installed, ensure_smoldot_built,
+    resolve_base_dir, run_js_test, ELASTIC_VALIDATOR_COUNT, FINALIZED_METRIC, PARA_ID,
 };
 use zombienet_sdk::{
-    subxt::ext::subxt_rpcs::rpc_params, LocalFileSystem, Network, NetworkConfig,
-    NetworkConfigBuilder, NetworkNode,
+    snapshot::BundleBuilder, subxt::ext::subxt_rpcs::rpc_params, Bundle, LocalFileSystem, Network,
+    NetworkConfig, NetworkConfigBuilder, NetworkNode,
 };
 
 const DEFAULT_TARGET_FINALIZED: u32 = 2500;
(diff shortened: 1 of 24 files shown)


skunert:
I think if we are really pedantic the current implementation does not contradict the spec.

Per spec, the following requirements need to be met:
 [chainHead_v1_follow.md](/home/sebastian/work/repos/json-rpc-interface-spec/src/api/chainHead_v1_follow.md:23):

- “When the node finalizes a block, generates a `finalized` notification indicating which blocks have been finalized and which blocks have been pruned.”
- `finalizedBlockHashes` “contains a list of blocks that are now part of the finalized block, ordered by increasing block number. The last element in this list is the current finalized block.” 
- All items in `finalizedBlockHashes` must have been reported earlier through `newBlock`

So if blocks A, B, C finalize together and only C is reported, it technically fulfills the requirements. Effectively, I assume that the spec is kept a bit vague by design so implementers have some freedom, consumers can really only rely on the last block.

But that being said, reporting more blocks does not really hurt.

e2e-tests/tests/smoke_generate_snapshots.rs
```diff
@@ -142,63 +142,92 @@ async fn smoke_generate_snapshots() -> Result<(), anyhow::Error> {
...
-        &network_base,
-        &args.out.join("parachain-db.tgz"),
-    )
-    .await?;
-
-    create_bundle(&args.out)?;
-    print_manifest(&args.out)?;
+    // Step 3: drain in-flight parachain candidates so the unfinalized relay
+    // suffix is candidate-free at snapshot time (see `drain_parachain_pipeline`).
+    drain_parachain_pipeline(&network).await?;
```
skunert: Maybe we should just snapshot both validators instead of having this hack. Seems more clean to me and afaik also what we do in polkadot-sdk

skunert (approved):
The logic fixes look good. I ngeneral I find that we still have a ton of zombienet boilerplate. Maybe we could improve the Bundle API a bit medium term 🤔

lexnv (approved):
Nicely done here! 🙏

e2e-tests/tests/smoke_generate_snapshots.rs
lrubasze: Indeed, I went too far with this hack. Will regenerate snapshots

lrubasze:
> I think if we are really pedantic the current implementation does not contradict the spec.
> 
> Per spec, the following requirements need to be met: [chainHead_v1_follow.md](/home/sebastian/work/repos/json-rpc-interface-spec/src/api/chainHead_v1_follow.md:23):
> 
>     * “When the node finalizes a block, generates a `finalized` notification indicating which blocks have been finalized and which blocks have been pruned.”
> 
>     * `finalizedBlockHashes` “contains a list of blocks that are now part of the finalized block, ordered by increasing block number. The last element in this list is the current finalized block.”
> 
>     * All items in `finalizedBlockHashes` must have been reported earlier through `newBlock`
> 
> 
> So if blocks A, B, C finalize together and only C is reported, it technically fulfills the requirements. Effectively, I assume that the spec is kept a bit vague by design so implementers have some freedom, consumers can really only rely on the last block.
> 
> But that being said, reporting more blocks does not really hurt.

Fair point, you're right - I claimed in the PR that the old behavior violated the spec, but it didn't. My bad.

Still worth it though: the single-hash notification was lossy - finalize A, B, C together and we'd only emit C, so anything tracking every finalized block missed A and B. I'll fix the PR description.
Thanks @skunert

PR #3284, opened 2026-06-09
Support proper Kademlia peer discovery

bkchr:
Right now Smoldot is only doing Kademlia queries to nodes for which it has "consensus streams" open aka sync etc. This means, if it connects to a bootnode that doesn't has any sync slots free (which is kind of normal for a bootnode), it will not be able to discover any new nodes. This is fixed by this pull requests. Besides that it also implements the "proper Kademlia walk".


lib/src/network/service.rs
```diff
@@ -272,6 +272,14 @@ pub struct ChainNetwork<TChain, TConn, TNow> {
...
         collection::SubstreamId,
     )>,
+
+    /// Peers for which an outbound Identify request has been sent (or completed) on at least
+    /// one connection. Used to ensure we issue Identify at most once per peer lifetime.
+    identify_requested_peers: hashbrown::HashSet<PeerIndex, fnv::FnvBuildHasher>,
+
+    /// Peers known to support a chain's Kademlia protocol, as determined by Identify responses.
+    /// Used by [`ChainNetwork::kademlia_capable_peers`].
+    kademlia_capable_peers: BTreeSet<(usize, PeerIndex)>,
```
dmitry-markin, line 282:
```suggestion
    kademlia_capable_peers_by_chain: BTreeSet<(usize, PeerIndex)>,
```

light-base/src/network_service.rs
```diff
@@ -2907,6 +2942,11 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
                 response: service::RequestResult::KademliaFindNode(Ok(nodes)),
                 ..
             }) => {
+                // Track whether this response taught us anything new. If so, we reset the
+                // chain's discovery backoff so that the next FindNode round runs at the
+                // initial 2s interval rather than continuing to back off — Kademlia is
+                // making progress, walk the DHT eagerly.
+                let mut any_new_peer = false;
```
dmitry-markin, line 2949: I would stop eager discovery as soon as we have all gossip slots occupied to not introduce the load on WSS-connected peers, which will be approx. the same (mostly bootnodes) for all light clients.

dmitry-markin (approved):
Worth keeping in mind this still differs from substrate approach, because we will send `FIND_NODE` requests only to gossip-connected peers AND attempted peers for gossip connection (i.e., if we fail to establish gossip connection due to no light client slots on the node). In contrast, substrate/litep2p establishes new connections specifically to perform `FIND_NODE` requests. Not sure implementing it worth the required smoldot refactoring though.

bkchr:
> requests only to gossip-connected peers AND attempted peers for gossip connection (i.e., if we fail to establish gossip connection due to no light client slots on the node)

What are gossip connected peers? You mean nodes that we are explicitly connected to for sync etc?

dmitry-markin:
> > requests only to gossip-connected peers AND attempted peers for gossip connection (i.e., if we fail to establish gossip connection due to no light client slots on the node)
> 
> What are gossip connected peers? You mean nodes that we are explicitly connected to for sync etc?

This is how smoldot calls peers with which we have notifications substreams opened.

light-base/src/network_service.rs
dmitry-markin: Looking at the logs — it takes some seconds (~<10) for smoldot to establish gossip connections and sync, but firing `FIND_NODE` at 2s intervals continues many minutes after the initial sync.

skunert (approved):

PR #3293, opened 2026-06-30
fix(network): skip stale Kad-capable peers in discovery

### LINKED ISSUE #3291: Panic: `internal error: entered unreachable code` in network_service (light-base)
I've been using `smoldot-light` 1.2.0 in my application connecting to a parachain on Kusama. I've been testing the integration and noticed that the client can occasionally panic due to reaching what's considered unreachable code:

```
panicked at light-base/src/network_service.rs:2310:78: 
internal error: entered unreachable code
```

It seems very rare, it's happened only once so far therefore I wasn't able to retrieve more details that could better help to debug it.

lrubasze:
## Problem

Closes #3291.

A peer's Kad-capable flag is learned from `Identify` and outlives the connection it was learned on (it's cleared only on full peer removal, not on disconnect). So a discovery candidate from `kademlia_capable_peers` may have no usable connection. `start_kademlia_find_node_request` then returns `NoConnection`, which the discovery loop treated as `unreachable!()` - and panicked.

## Fix

Extract the dispatch loop into `dispatch_find_node_requests`, which skips `NoConnection` candidates instead of panicking and keeps trying until `PARALLEL_FIND_NODE_PER_ROUND` requests have started. Stale entries no longer panic nor consume a discovery slot; gossip-connected peers backfill reliably.

The function is platform-free and generic over the network's type parameters, so it's covered by a unit test that fails (panics) if `unreachable!()` is reintroduced.

### DIFF
--- a/e2e-tests/run_chainhead_test.sh
+++ b/e2e-tests/run_chainhead_test.sh
@@ -39,8 +39,8 @@ case "${network}" in
   kusama)
     relay_spec="${SPECS_DIR}/ksmcc3.json"
     para_spec="${SPECS_DIR}/ksmcc3_asset_hub.json"
-    relay_rpc=""  # TODO: set Kusama relay RPC URL
-    para_rpc=""  # TODO: set Kusama Asset Hub RPC URL
+    relay_rpc="https://kusama-rpc.n.dwellir.com"
+    para_rpc="https://asset-hub-kusama-rpc.n.dwellir.com"
     ;;
   westend)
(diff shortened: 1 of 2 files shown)


lrubasze:
Issue reproduced manually: https://github.com/paritytech/smoldot/issues/3291#issuecomment-4845573120

dmitry-markin (approved):

bkchr (approved):

Issue #3290, opened 2026-06-29
Parachain consensus bootstrap gets stuck on RemoteCouldntAnswer and never recovers


jsamol:
I've been using `smoldot-light` 1.2.0 in my application connecting to a parachain on Kusama. While testing I noticed that when the client fails to bootstrap parachain consensus, it falls into an endless retry loop and never recovers unless killed and restarted:

```
Bootstrapping parachain consensus from block #10982914 (0xe5ad…395b)
Downloading parachain runtime from peer 12D3KooWRNJYkFhPsoHDa2qJcVsfsqsNyaNV2cdc1XEqC9aUk5Rd
Failed to bootstrap parachain consensus: Storage proof request failed: RemoteCouldntAnswer. Retrying in 5s...
```

The initial bootstrapping issue seems hard to reproduce, I guess it heavily depends on the network conditions, I was hitting it quite consistently the other day, but now it's gone. When it did happen, though, it would get stuck on the retry every time.

lrubasze:
Problem reproduced with the attached script [run_3290.sh](https://github.com/user-attachments/files/29556629/run_3290.sh).
Issue occurs indeed occasionally, 3 times per 200 script iterations.

It also confirms the fix introduced in https://github.com/paritytech/smoldot/pull/3294

```shell
❯ bat _logs/fix_confirmed_iter13_1782903197.log | grep  -E "trying next|Bootstra|Downloading|Downloaded"
[2026-07-01T10:52:45.727Z] [INFO] [sync-service-asset-hub-kusama] Bootstrapping parachain consensus from block #18538934 (0x7cbb…7c11)
[2026-07-01T10:52:45.727Z] [INFO] [sync-service-asset-hub-kusama] Downloading parachain runtime from peer 12D3KooWFgAuJagLKBf21ttyYroXhR3YxAe9gUhzkYsET6NzARq2
[2026-07-01T10:52:45.729Z] [DEBUG] [sync-service-asset-hub-kusama] Bootstrap via peer 12D3KooWFgAuJagLKBf21ttyYroXhR3YxAe9gUhzkYsET6NzARq2 failed: Storage proof request failed: RemoteCouldntAnswer; trying next peer
[2026-07-01T10:52:45.729Z] [INFO] [sync-service-asset-hub-kusama] Downloading parachain runtime from peer 12D3KooWMzvdGcUXxacLdMQzRVrsP1mJrZHcrz8LtGbhLzve84Qx
[...]

PR #3294, opened 2026-07-01
fix: rotate peers when bootstrapping parachain consensus

lrubasze:
## Problem
Parachain bootstrap always used the first connected peer to download the runtime. If that peer replied `RemoteCouldntAnswer` (it can't serve theblock), the request failed, and the 5s retry loop re-picked the same peer
and same block - so bootstrap could retry forever without recovering.

Reported in  #3290.

## Fix
Try each connected peer in turn, rotating past any that fails. Only return an error once all peers have been tried, so the caller retries against a possibly-changed peer set instead of spinning on one peer.

## Tests
Unit tests for the rotation driver: rotate past a failing peer, error (not loop) once all peers fail, empty-peer handling, single-attempt on success.

Also verified end-to-end against live Kusama Asset Hub: warp sync + parachain follow pass with no regression.

### DIFF
--- a/light-base/src/sync_service/parachain.rs
+++ b/light-base/src/sync_service/parachain.rs
@@ -1330,6 +1330,16 @@ async fn bootstrap_parachain_consensus<TPlat: PlatformRef>(
 
     let peers = connected_peers_or_wait(network_service).await;
 
+    log!(
+        platform,
+        Trace,
+        log_target,
+        format!(
+            "Attempting parachain runtime download from {} connected peer(s)",
+            peers.len()
+        )
+    );
+
     first_successful_peer(peers, |peer_id| {
         attempt_bootstrap_with_peer(
             log_target,
--- a/light-base/src/sync_service/parachain.rs
+++ b/light-base/src/sync_service/parachain.rs
@@ -1311,25 +1311,15 @@ async fn bootstrap_parachain_consensus<TPlat: PlatformRef>(
 
     let peers = connected_peers_or_wait(network_service).await;
 
-    first_successful_peer(peers, |peer_id| async move {
-        let result = attempt_bootstrap_with_peer(
+    first_successful_peer(peers, |peer_id| {
+        attempt_bootstrap_with_peer(
             log_target,
             platform,
             network_service,
             chain_info,
             block_number_bytes,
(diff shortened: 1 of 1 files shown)


lrubasze:
Issue reproduced and fix confirmed: https://github.com/paritytech/smoldot/issues/3290#issuecomment-4856791581

light-base/src/sync_service/parachain.rs
```diff
@@ -1309,24 +1309,77 @@ async fn bootstrap_parachain_consensus<TPlat: PlatformRef>(
         )
     );
 
-    // Wait for a peer to connect.
-    let peer_id = {
-        let mut from_network = Box::pin(network_service.subscribe().await);
+    let peers = connected_peers_or_wait(network_service).await;
```
lexnv, line 1331: dq: If we enter here without any peers connected, we call into `connected_peers_or_wait` and loop through the first `Event::Connected`. I believe that guarantees the `network_service.peers_list()` contains a single peer unless another un-drained `Event::Connected` is queued in between.

In that case we only get a single peer back and the fix has parity with prior code logic. Would it be worth waiting for a number of peers before returning, ie maybe 3? Feel free to ignore

lexnv (approved):

light-base/src/sync_service/parachain.rs
lexnv, line 1331: nit: Might be worth adding a trace log for the peers

skunert (approved):

light-base/src/sync_service/parachain.rs
lrubasze: My idea was to attempt bootstrap with whatever's connected (even a single peer). `first_successful_peer` rotates through the collected connected peers, if all fail, the outer retry loop (parachain.rs:75-96) backs off 5s and then connected peers are recollected via `peers_list()`, so newly-connected peers join the rotation the next attempt. I wanted to avoid waiting for N peers upfront hoping that single peers could do the job and widen the peer pool only after a failure.
Would prefer to keep it as is unless I am missing something.

lrubasze: Done

PR #3299, opened 2026-07-07
test: refactor e2e-tests to run on both Node (TCP) and headless browser (WebRTC)

gab8i:
Refactor `e2e-tests/` so each test is written once as a host-agnostic body in `shared/<name>.js` and executed on two hosts, smoldot Node.js over TCP, and the smoldot browser in headless browser over WebRTC.
This replaces the duplicated `e2e-tests/js/` and `e2e-tests/browser/`.

The reason why this refactor was needed is that node.js has no native `RTCPeerConnection`, so smoldot cannot open WebRTC connections under Node.
The first attempt was to inject global WebRTC object but it was abandoned because the injected API deviate from the browser API leading to a non realistic e2e tests.

Instead, WebRTC now runs where it is 'native', the browser host drives the untouched browser build inside real headless Chromium, while the Node host covers the same test logic over TCP.

Changes are:
- Layout
  - `shared/`: test bodies + shared primitives
  - `hosts/node/`, `hosts/browser/`: host exucutors
- `ctx` contract: tests get a fixed interface  validated by both runners at startup, they only use web-platform APIs (no Node APIs), so they run unchanged on both hosts.
- Rust harness: `network.rs` starts nodes with `--experimental-webrtc` on `webrtc-direct` and writes chain specs whose `bootNodes` are dialable over both TCP and WebRTC.


e2e-tests/src/network.rs
```diff
@@ -331,45 +381,86 @@ async fn assign_elastic_cores(network: &Network<LocalFileSystem>) -> Result<(),
...
+    let mut out: Vec<String> = Vec::new();
+    for name in names {
+        let node = network.get_node(*name)?;
+        let rpc = node.rpc().await?;
+        let mut listen_addrs: Vec<String> = rpc
+            .request::<Vec<String>>("system_localListenAddresses", RpcParams::new())
+            .await
+            .map_err(|e| anyhow!("{name}: system_localListenAddresses failed: {e}"))?
+            .into_iter()
+            // Keep loopback addresses and only webrtc or tcp protocols.
```
lrubasze: nit.. comment suggests that webrtc and tcp check happens in filter below.
i'd reword the comment or refactor the filter, eg.

```rust
.filter(|addr| addr.contains("/ip4/127.0.0.1/") && (addr.contains("/tcp/") || addr.contains("/webrtc")))
```

e2e-tests/hosts/node/ctx.js
```diff
@@ -0,0 +1,88 @@
...
+    console.log(`[${ts}] PASS: ${name}${suffix}`);
+  } else {
+    console.log(`[${ts}] FAIL: ${name}${suffix}`);
+    process.exitCode = 1;
+  }
+}
+
+// Polls the Rust → JS sync file until a line equals `expected`. Pair with
+// `SyncFile` on the Rust side.
+async function waitForMessage(path, expected, timeoutMs = 120_000, pollMs = 100) {
```
lrubasze: Is it the same as https://github.com/paritytech/smoldot/pull/3299/changes#diff-16e25dd9c2af6bd505db3fb17ce7de7ad9f1a7f857124445bd17cab4c2800d87R26?
if yes, then maybe we could move it to the shared part?

e2e-tests/hosts/browser/run.js
```diff
@@ -0,0 +1,152 @@
...
+
+// Resolve inputs in Node (the browser sandbox can't read disk or env).
+const mod = await import(new URL(`../../shared/${testName}.js`, import.meta.url));
+const fileInputs = mod.fileInputs ?? [];
+const files = {};
+for (const name of fileInputs) {
+  const p = process.env[name];
+  files[name] = p ? await fs.readFile(p, "utf8") : null;
+}
+const env = { ...process.env };
```
lrubasze: according to Claude whole CI runner env is serialized into Chromium page. Anything sensitive in the CI env becomes readable by page JS. Maybe we could serialize only required data?

lrubasze (approved):
LGTM with some minor comments!
Nice work @gab8i

e2e-tests/hosts/browser/run.js
gab8i: Oh, thanks! Yes, it is definitely better to restrict the serialized envs to what is really needed!

e2e-tests/hosts/node/ctx.js
gab8i: Oh yes, it is kind of the same!! 

The thing is that they include fs differently so I will try to join them, not sure if I will manage to do it though

lrubasze: If joining complicates things too much, then maybe just leave it as is with some comment only?

gab8i: It turned out to be pretty easy, pushing everything soon!

gab8i:
Temporarily disable browser test, there is an upstream fix required to make WebRTC work. It makes sense to merge in the meantime to ensure no change is introduced to the tests executed within Node by this refactor.

e2e-tests/shared/smoke.js
```diff
@@ -15,37 +15,42 @@
...
+  "RELAY_CHAIN_SPEC",
+  "PARA_CHAIN_SPEC",
+  "SMOLDOT_DB_RELAY",
+  "SMOLDOT_DB_PARA",
+];
+
+export const envInputs = ["REQUIRED_BLOCKS", "EXPECTED_INITIAL_FINALIZED"];
+
+// Hex → bytes without Node's Buffer, so this runs unchanged in the browser.
+function hexToBytes(hex) {
```
dmitry-markin: nit: we can import it from `codec.js` as other files do. Also may be `decodeHeaderNumber`.

dmitry-markin (approved):
Did an AI-assisted review, and the PR looks good.

Minor nits:
- `docs/RELEASING.md` still mentions `e2e-tests/js`, `e2e-tests/browser`
- `cd e2e-tests && cargo +nightly fmt` is not a no-op

PR #3300, opened 2026-07-07
sync_service: bound AwaitingWarp wait with warp progress detection

lrubasze:
## Summary

Resolves a TODO from #3268 ([comment](https://github.com/paritytech/smoldot/pull/3268#discussion_r3319656011)): a warp sync that stays alive but never completes kept queued `subscribe_all` requests waiting forever - the `AwaitingWarp` deadline re-armed as long as warp looked alive, and the starve/re-trigger cycle through `Deciding` reset the timer each round.

## Changes

- Track warp's verified finalized height (`warp_sync_progress()`), baselined on `AwaitingWarp` entry; each 30s deadline firing with progress re-arms the deadline.
- One window without progress commits AllForksOnly: queued subscribers get the checkpoint header, like the other fallback paths. Committing is cheap - warp keeps running, and a later completion resets subscriptions at the warped head. A pending warp completion counts as progress so it is never discarded by the commit.
- The `AwaitingWarp` → `Deciding` starve fallback is removed; the stall commit covers that case.

Known edge: a warp that advances at least one fragment per window never triggers the fallback, so subscribers wait until it completes.


light-base/src/sync_service/substrate_compat.rs
```diff
@@ -54,6 +54,11 @@ const MODE_DECISION_TIMEOUT: Duration = Duration::from_secs(30);
 /// first peer locking us into the slow path when our local finalized is a stale checkpoint.
 const MODE_DECISION_MIN_PACKETS: usize = 2;
 
+/// Deadline firings (`MODE_DECISION_TIMEOUT` apart) without warp progress before committing
+/// AllForksOnly, so queued `SubscribeAll` requests always receive a response (the checkpoint
+/// header).
+const MODE_DECISION_MAX_WARP_STALLS: usize = 3;
```
lrubasze: Open question: how many stalled windows to wait before falling back?

Current draft waits 3 stalled windows (~90s). I'm considering 1 (~30s): a healthy warp normally shows progress within one window, and committing "too early" is cheap - subscribers get the checkpoint header, warp keeps running, and on completion they get `Stop` + resubscribe at the warped head anyway. It also removes the stall counter and the `AwaitingWarp` → `Deciding` fallback entirely.

WDYT?

lrubasze: Went with 1 window. Committing early costs nothing: warp keeps running and its completion still resets subscriptions at the warped head. So waiting longer has no benefit. The stall counter and the fallback to Deciding are removed.

Small edge that remains: if warp makes some progress every window, we never commit and subscribers wait until warp finishes. If we ever want a hard cap on that wait, we can commit on the first deadline regardless of progress.

Issue #3302, opened 2026-07-20
smoldot crashes in a Worker/SharedWorker when dialing a webrtc-direct address (RTCPeerConnection undefined)


BigTava:
#### Reproduction steps

1. See https://github.com/paritytech/dotli-community/pull/85
2. Open https://host-playground.westend.li/

#### Expected behaviour

Same as https://host-playground.paseo.li/

#### Current behaviour

```
Uncaught (in promise) ProtocolInitFailedError: Init failed: SharedWorker error: chainHead follow stopped (cause: ChainHead stopped)
```
#### Explanation

The browser connect code (`no-auto-bytecode-browser.js`, `config.address.ty === "webrtc"` branch) calls `RTCPeerConnection.generateCertificate(...)` unconditionally. `RTCPeerConnection` exists only on `Window` and is `undefined` in worker scopes, so the call throws, and a connect binding throwing synchronously panics the whole client.

| scope | `typeof RTCPeerConnection` |
|---|---|
| Window | `function` |
| Dedicated / Shared Worker | `undefined` |

(`WebSocket` is defined in all three, so WSS bootnodes are unaffected. On the main thread the same spec works.)

Run smoldot via `start()` in a worker, `addChain` with a spec whose `bootNodes` include a `/webrtc-direct/` multiaddr. On the dial: `Uncaught Error: Smoldot has panicked` and `chainHead_v1_follow` emits `stop`. One webrtc bootnode among healthy WSS ones is enough, since the throw is fatal to the entire client.

#### Suggestion 

In the webrtc branch, if `typeof RTCPeerConnection === "undefined"`, call `config.onConnectionReset(...)` and skip instead of constructing the connection. Verified locally: this makes a worker-hosted smoldot fall back to WSS with no panic.

lrubasze:
The `ReferenceError: RTCPeerConnection is not defined` described above unwinds through the Wasm while `connect_multistream` holds the global `STATE` lock, so the lock stays held forever. The exception is then silently swallowed by the `catch` around `advance_execution()`, and every later call into the Wasm panics at `STATE.try_lock().unwrap()`.

Sentry issues attributable to this cascade:

- [DOTLI-B1](https://paritytech.sentry.io/issues/DOTLI-B1/) (#3306) - `connection_reset` entry point
- [DOTLI-B2](https://paritytech.sentry.io/issues/DOTLI-B2/) (#3307) - `stream_message` entry point
[...]

PR #3303, opened 2026-07-21
fix: forbid webrtc if RTCPeerConnection undefined

gab8i:
This fix addresses https://github.com/paritytech/smoldot/issues/3302, where using WebRTC within a worker thread (an environment where WebRTC cannot be accessed) led to a crash of the entire client.

### DIFF
--- a/wasm-node/javascript/src/no-auto-bytecode-browser.ts
+++ b/wasm-node/javascript/src/no-auto-bytecode-browser.ts
@@ -45,6 +45,11 @@ export {
 export function startWithBytecode(options: ClientOptionsWithBytecode): Client {
     options.forbidTcp = true;
 
+    // Browsers only expose the WebRTC API on `Window`, it doesn't exist in worker global
+    // scopes, and consequently there is no way to open WebRTC connections from here.
+    if (typeof RTCPeerConnection === 'undefined')
+        options.forbidWebRtc = true;
+
     // When in a secure context, browsers refuse to open non-secure WebSocket connections to
     // non-localhost. There is an exception if the page is localhost, in which case all connections
     // are allowed.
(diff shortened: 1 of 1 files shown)


lrubasze (approved):
LGTM

wasm-node/javascript/src/no-auto-bytecode-browser.ts
```diff
@@ -191,6 +196,28 @@ function connect(config: ConnectionConfig): Connection {
...
+        // addresses.
+        //
+        // Note: `startWithBytecode` sets `forbidWebRtc` in that situation, making
+        // this code path unreachable, but better be defensive.
+        if (typeof RTCPeerConnection === 'undefined') {
+            let cancelled = false;
+            setTimeout(() => {
+                if (!cancelled)
+                    config.onConnectionReset('RTCPeerConnection is not available in this environment');
+            }, 1);
```
dmitry-markin, line 212: Why do we need `setTimeout` instead of calling the function directly?

dmitry-markin (approved):

wasm-node/javascript/src/no-auto-bytecode-browser.ts
```diff
@@ -191,6 +196,28 @@ function connect(config: ConnectionConfig): Connection {
...
+        // report the connection as reset so that smoldot moves on to other
+        // addresses.
+        //
+        // Note: `startWithBytecode` sets `forbidWebRtc` in that situation, making
+        // this code path unreachable, but better be defensive.
+        if (typeof RTCPeerConnection === 'undefined') {
+            let cancelled = false;
+            setTimeout(() => {
+                if (!cancelled)
+                    config.onConnectionReset('RTCPeerConnection is not available in this environment');
```
lexnv, line 211: Would this require changes on the dotly side to change the environment? 🤔  In other words, this is not hiding the issue, but making it clearer to implementers that they need to change the environment?

wasm-node/javascript/src/no-auto-bytecode-browser.ts
gab8i: setTimeout moves the "this connection failed" report from during registration to after registration.

Before the closure is effectively called it could happen that smoldot closes the connection calling the
reset callback which update the `cancelled` flag which will let the callback called at the and, the one within the timeout, being a no-op

wasm-node/javascript/src/no-auto-bytecode-browser.ts
gab8i: Yes, exactly!

I mean, as the comment says, this is a defensive guards, webrtc stuff will be completely skip if the environment doesn't allow to use it in the first place

Issue #3305, opened 2026-07-22
wasm: panic "same stream_id used multiple times in connection_stream_opened" - RTCDataChannel onopen not idempotent


lrubasze:
Seen via Sentry on smoldot npm **3.3.1** (browser, WebRTC connection): [DOTLI-B4](https://paritytech.sentry.io/issues/DOTLI-B4/). The same crash was also captured as [DOTLI-B0](https://paritytech.sentry.io/issues/DOTLI-B0/) and [DOTLI-4K](https://paritytech.sentry.io/issues/DOTLI-4K/) through `window.onerror` (same trace_id/user, message stripped to "Smoldot has panicked").

## Panic

```
panicked at wasm-node/rust/src/platform.rs:1115:13:
same stream_id used multiple times in connection_stream_opened
```

Fires when JS calls `connectionStreamOpened` twice with the same `(connection_id, stream_id)` on a multi-stream (WebRTC) connection.

## Analysis

Stream ids come from a per-connection incrementing counter (`state.nextStreamId` in `wasm-node/javascript/src/no-auto-bytecode-browser.ts`), so a duplicate id can only reach wasm if `config.onStreamOpened(streamId, ...)` is invoked twice for the same channel. The `onopen` handler allows exactly that:

```ts
dataChannel.onopen = () => {
    console.assert(!isOpen.value, "substream opened twice")
    isOpen.value = true;
    config.onStreamOpened(streamId, direction);
    config.onWritableBytes(65536, streamId);
};
```

`console.assert` only logs. If the browser fires `open` twice on an `RTCDataChannel` (SPECULATED trigger; the assert suggests this was considered possible), `onStreamOpened` runs again and wasm panics with exactly this message.

## Suggested fix

- Make the guard real: `if (isOpen.value) return;` in `onopen`.
- Optionally also dedupe in `remote-instance.ts` `stream-opened` handling (currently `add()`s to the per-connection Set without a `has()` check).

## Environment

- smoldot npm 3.3.1, Chrome 150, macOS, `paseo-next-v2` network, 1 user

lrubasze:
Fixed in #3309

PR #3309, opened 2026-07-23
wasm-node: ignore duplicate RTCDataChannel open events

lrubasze:
## Problem

Browsers can deliver the `open` event of an `RTCDataChannel` more than once (per MDN it fires when the underlying transport "is opened or reopened"). The browser platform code reported the same stream id to the wasm twice, which panics:
```
 same stream_id used multiple times in connection_stream_opened
```

Observed in production (smoldot npm 3.3.1, Chrome, WebRTC). The existing `console.assert` anticipated the double-fire but only logged it.

Observed in #3305 

## Fix

Make the `onopen` handler idempotent: return early if the substream was already reported open. Each `open` event past the first is ignored.

## Test

New browser-only e2e test `webrtc_double_open`: reuses the smoke body, and a browser-host prepare extension dispatches a second genuine `open` event on every data channel after the real one - through the real event machinery, indistinguishable from a native double-fire. Without the fix the first substream panics; with it the client must sync normally.

The CI matrix entry is added but commented out: browser test execution is currently disabled (`Host::Browser` arm in `e2e-tests/src/lib.rs`, pending an upstream fix), and the job would pass without executing anything. A reminder comment at the disable site links the two.


lexnv (approved):
Nice! Not an expert of the JS realm, but the rust test looks solid to me 🙏

gab8i (commented):
The `e2e-tests/hosts/browser/prepare/webrtc_double_open.js` file could probably be deleted and everything moved within `e2e-tests/shared/webrtc_double_open.js`

The resulting test file should look like this change here: https://github.com/paritytech/smoldot/commit/ffff934a3f3dc075cb292c0d3599269668412324#diff-611227a6a6abc2bec54c03b92b28509bfc70a6b0028b8403bc23c002cafed179

this should simplify the test structure

lrubasze:
> The `e2e-tests/hosts/browser/prepare/webrtc_double_open.js` file could probably be deleted and everything moved within `e2e-tests/shared/webrtc_double_open.js`
> 
> The resulting test file should look like this change here: [ffff934#diff-611227a6a6abc2bec54c03b92b28509bfc70a6b0028b8403bc23c002cafed179](https://github.com/paritytech/smoldot/commit/ffff934a3f3dc075cb292c0d3599269668412324#diff-611227a6a6abc2bec54c03b92b28509bfc70a6b0028b8403bc23c002cafed179)
> 
> this should simplify the test structure

Good point! Done :)

gab8i (approved):

Issue #3304, opened 2026-07-22
Native: panic in `Yamux::index_mut` from `single_stream_connection_task` (iOS, smoldot-light)


lrubasze:
Seen via Sentry in a native (Rust, smoldot-light) integration on iOS: [BREVITY-MORPHEUS-4C](https://paritytech.sentry.io/issues/BREVITY-MORPHEUS-4C) - 2 events, 2 users, substatus regressed.

## Crash

SIGABRT on thread `smoldot-light-2`, ~1 minute after app start:

```
smoldot_light::network_service::tasks::single_stream_connection_task::{{closure}}
smoldot::libp2p::connection::yamux::Yamux<T>::index_mut
core::panicking::panic
```

`Yamux::index_mut` (`lib/src/libp2p/connection/yamux.rs:1736`) panics when the `SubstreamId` is not in the substream map:

```rust
impl<TNow, TSub> ops::IndexMut<SubstreamId> for Yamux<TNow, TSub> {
    fn index_mut(&mut self, substream_id: SubstreamId) -> &mut TSub {
        ...
            .unwrap_or_else(|| panic!())
```

## Analysis

- The concrete caller is inlined away; candidates are the `self.inner.yamux[substream_id]` accessors in `lib/src/libp2p/connection/established/single_stream.rs` (~16 sites).
- SPECULATED: a coordinator message (request start, notification write, close, ...) is processed for a substream that has already died and been removed from the yamux state - a race between substream death and coordinator messages. Related past fix: #2869.

## Environment

- iOS 26.2, iPhone17,1, app `io.parity.brevity@1.0.0+39` (Flutter, native smoldot-light)
- smoldot version pinned by the app: unknown, needs confirmation from the app team

## Next steps

- [ ] Confirm smoldot(-light) version used by the reporting app
- [ ] Identify which accessor fires (debug log or better symbolication)

lrubasze:
Related Brevity Sentry issues :
- [BREVITY-MORPHEUS-4S](https://paritytech.sentry.io/issues/BREVITY-MORPHEUS-4S/) — same panic, from a macOS **debug** build (2 events, 2026-07-23). Symbolication resolves the frames that were inlined away in the iOS release build:

  ```
  inject_coordinator_message (single_stream.rs:488)
  → reject_in_notifications_substream
  → Yamux<T>::index_mut (yamux.rs:1736)
  ```

This answers the two open questions in the issue:

[...]

PR #3310, opened 2026-07-24
libp2p/yamux: include substream id in Index/IndexMut panic message

lrubasze:
The `Index`/`IndexMut` impls for `Yamux` panic with a bare "explicit panic" when a `SubstreamId` is missing from the map. Since these accessors are inlined, crash reports (e.g. #3304) show only `Yamux::index_mut` with no clue which substream. Add the id to the message:
```
    no substream with id <N> in yamux state
```

Refs #3304


gab8i (approved):

lexnv (approved):
This should help us identify the issue, nice!

PR #3316, opened 2026-07-30
light-base: use a dedicated channel for statement notifications

### LINKED ISSUE #3146: statement-store: use a dedicated statement notifications channel
We might want to use a dedicated statement notifications channel in the future instead of extending network events. The benefits:
1. Won't need to receive and discard network events in the json-rpc service.
2. If at some point statements become large or their amount becomes large, there is no point in dispatching them to all network subscribers.

At least worth estimating the overhead as a follow-up and may be decoupling them. @AndreiEres can you create an issue for this, please?

_Originally posted by @dmitry-markin in https://github.com/paritytech/smoldot/pull/3127#discussion_r3001473405_

AndreiEres:
Fixes https://github.com/paritytech/smoldot/issues/3146

Statement notifications rode the general network event stream, so every subscriber received them and the JSON-RPC service — their only consumer — had to loop over unrelated events and discard them. They now travel on their own channel, following the pattern already established for Bitswap messages.


light-base/src/json_rpc_service/background.rs
```diff
@@ -787,21 +787,20 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                let Some(rx) = &me.statement_events_rx else {
+                    return WakeUpReason::MustSubscribeStatementEvents;
+                };
+                let Ok(network_service::StatementEvent::StatementsNotification {
+                    statements, ..
+                }) = rx.recv().await
+                else {
+                    me.statement_events_rx = None;
+                    return WakeUpReason::MustSubscribeStatementEvents;
+                };
```
dmitry-markin: nit: it would be a bit more readable like:
```
match rx.recv().await {
    Ok(...) => WakeUpReason::NetworkStatementsReceived(statements),
    Err(_) => WakeUpReason::MustSubscribeStatementEvents,
}
```

light-base/src/network_service.rs
```diff
@@ -1203,8 +1254,10 @@ async fn background_task<TPlat: PlatformRef>(mut task: BackgroundTask<TPlat>) {
             let service_event = async {
                 if let Some(event) = (task.events_pending_send.is_empty()
                     && task.bitswap_event_pending_send.is_none()
+                    && task.statement_event_pending_send.is_none()
```
dmitry-markin, line 1257: Something to think about for a follow-up — if the statement traffic is high, we might want to discard statements and not backpressure the entire network to not stop smoldot networking completely. Another option is backpressuring only statement substreams, which is possible with WebRTC/QUIC, but implementing this is likely quite involved.

dmitry-markin (approved):

DenzelPenzel (approved):

PR #3317, opened 2026-07-30
light-base: add lifecycle_unstable_follow RPC subscription

### LINKED ISSUE #3301: Smoldot lifecycle event stream in light-base
## Motivation

Several consumers need to know what the light client is doing (bootstrap progress, sync mode, warp state), and today the only source is debug log text. Log matching is fragile - wording changes silently break benchmarks and tests.

Current and near-future consumers:

- **embedder UX**: cold start takes seconds to tens of seconds and the dapp gets nothing until `chainHead_follow` answers - the page looks frozen, the most-cited objection to light clients vs RPC endpoints. Events enable a real progress indicator ("found 3 peers" → "warp syncing 1.2M / 29.4M" → "ready") and, after startup, distinguishing "no new blocks" from "disconnected". substrate-connect could render this once for every chain.
- **health detection**: failures like "no peers", "warp stalled", or a 90s bootstrap are invisible to the embedder - the only symptom is an RPC call that never resolves. Events let a watchdog react to the cause (retry other bootnodes, notify the user, fall back to RPC) instead of a blunt restart-after-timeout, and let operators export stall rates and bootstrap times as metrics.
- **benchmarks**: phase markers (time-to-first-peer, warp duration, first finalized event) are matched from log lines
- **e2e tests**: asserting fallback paths (e.g. mode-decision commits from #3268) means grepping logs
- **telemetry**: constants like the 30s mode-decision timeout were picked without field data
- **bug reports**: a compact event trace beats megabytes of debug logs

## Proposal

A per-chain broadcast of typed lifecycle events, published by sync_service / network_service.

Sketch of events:

```rust
pub enum LifecycleEvent {
    Connecting,
    FirstPeer,
    ModeDecision { mode: SyncMode },
    WarpSyncProgress { at: u64, target: u64 },
    WarpSyncFinished { finalized: u64 },
    BootstrapComplete,
    Stalled { reason: StallReason },
}
```

Exposure, layer by layer:

- **light-base (Rust)**: `lifecycle_events(chain_id) -> impl Stream<Item = LifecycleEvent>`; first item is a snapshot of current state; bounded buffers, lag marker on overflow (same pattern as `subscribe_all`)
- **JSON-RPC**: smoldot-specific unstable subscription (`smoldotLifecycle_unstable_follow`, chainHead-style). Rides the existing RPC pipe, so the npm package, substrate-connect, and e2e tests get it with no new surface
- **logs**: generate the existing debug lines from the events, so log text becomes a projection of the stream instead of a parallel truth

## Design points

- event schema is unstable and versioned
- snapshot-on-subscribe semantics (like chainHead `initialized`)
- slow consumers must never block syncing
- optional ring buffer of last N events for bug-report traces

## Alternatives considered

Structured key=value log events (cheaper, but still log-transport based and not reachable by dapps).

BigTava:
## Summary

- Adds `lifecycle_unstable_follow` / `lifecycle_unstable_unfollow` JSON-RPC methods, sharing the 32-byte bs58 subscription-id format used by `transactionWatch_v1_submitAndWatch` and `bitswap_unstable_stream`.
- Defines seven `LifecycleEvent` variants (`Connecting`, `FirstPeer`, `ModeDecision`, `WarpSyncProgress`, `WarpSyncFinished`, `BootstrapComplete`, `Stalled`) with matching JSON schema and docstrings on both sides.
- Routes each subscription through the existing `background_tasks` `FuturesUnordered` with map-lookup unsubscribe-race handling, mirroring the `transactionWatch` and `bitswap` idioms.
- Closes #3301


lib/src/json_rpc/methods.rs
```diff
@@ -1142,6 +1152,21 @@ pub enum StatementEvent {
     },
 }
 
+/// Notification event for `lifecycle_unstable_follow` subscriptions.
+///
+/// The schema is unstable and versioned. See
+/// <https://github.com/paritytech/smoldot/issues/3301>.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(tag = "kind", rename_all = "camelCase")]
+pub enum LifecycleEvent {
```
lrubasze, line 1160: Is it useful with only 2 events? I mean maybe it would be better to add more events in this PR rather than postponing to a follow-up.

BigTava: I will let you know when it is a good moment to review. Thank you for peeking nonetheless!

lrubasze: yeah, missed it was a draft. sorry :)

light-base/src/lib.rs
```diff
@@ -1277,17 +1302,363 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+    // `ModeDecision`, so it cannot race ahead. `warp_done` gates the poller
+    // so no stray `WarpSyncProgress` lands after `WarpSyncFinished`.
+    platform.spawn_task("lifecycle-bootstrap".into(), {
+        let lifecycle_service = lifecycle_service.clone();
+        let sync_service = sync_service.clone();
+        let platform = platform.clone();
+        async move {
+            // `subscribe_all` returns after the sync service commits its
+            // bootstrap mode, so `bootstrap_mode()` answers immediately here.
+            let subscription = sync_service.subscribe_all(16, false).await;
```
skunert: This blocks until warp sync is done: the sync background only answers `SubscribeAll` once `ModeState::Ready`, which for warp chains is set in the `WarpSyncFinished` handler. So `modeDecision` arrives after warp finishes, and the progress poller below starts too late, sees `warp_sync_position() == None` and exits — `warpSyncProgress` is never emitted at all. I think these events need to be emitted from inside `substrate_compat.rs` where the transitions happen, not observed from the outside.

lib/src/json_rpc/service/client_main_task.rs
```diff
@@ -474,7 +474,8 @@ impl ClientMainTask {
                 | methods::MethodCall::transactionWatch_v1_submitAndWatch { .. }
                 | methods::MethodCall::sudo_network_unstable_watch { .. }
                 | methods::MethodCall::bitswap_unstable_stream { .. }
-                | methods::MethodCall::chainHead_v1_follow { .. } => {
+                | methods::MethodCall::chainHead_v1_follow { .. }
+                | methods::MethodCall::lifecycle_unstable_follow { .. } => {
```
skunert, line 478: `SubscriptionStartProcess::accept()` (around line 1215) has no arm for this method, so library embedders following the normal `.accept()` flow hit the `unreachable!()` and panic. The full node only survives via its catch-all `fail(ServerError)`.

light-base/src/lib.rs
```diff
@@ -1277,17 +1302,363 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+            let mut last_warp_position: Option<u64> = None;
+            let mut last_warp_advance_seen = started.clone();
+            // Reason latched by the most recent unmatched `Stalled`
+            // emission, or `None` if currently healthy.
+            let mut last_stall_reason: Option<lifecycle_service::StallReason> = None;
+
+            loop {
+                // Drain pending network events non-blockingly. A closed
+                // sender means the per-chain network task is gone. Exit.
+                loop {
```
skunert: This channel must be polled continuously (see the note in `network_service.rs`), but here it's only drained once per 5s tick. The network background delivers events with a blocking `send().await`, so once the 128-slot buffer fills, networking stalls for every chain until the next tick. Polling `peers_list()` each tick would avoid this and the manual peer counter.

light-base/src/json_rpc_service/background.rs
```diff
@@ -6139,6 +6199,157 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                                event: next,
+                                events_rx,
+                            }
+                        }));
+                    }
+                    None => {
+                        // Broadcaster gone. Emit a terminal `Stopped { ChainRemoved }`
+                        // so the client sees a clean shutdown rather than a silent drop,
+                        // then remove the subscription.
+                        let notification =
```
skunert, line 6359: This reports `chainRemoved` for any channel closure, including when `emit()` dropped the sender because this subscriber lagged. A slow client on a live chain gets told the chain was removed; `lagged` is never actually emitted.

light-base/src/lib.rs
```diff
@@ -1277,17 +1302,363 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+            // one so this task no longer contributes to keeping the chain
+            // alive.
+            drop(sync_service);
+
+            // The first item on `new_blocks` (either `Block` or `Finalized`)
+            // means live-block streaming has started. Waiting only for
+            // `Finalized` would push this out by the GRANDPA round cadence
+            // for no additional signal.
+            let new_blocks = subscription.new_blocks;
+            if new_blocks.recv().await.is_ok() {
```
skunert: If the channel closes before the first item arrives (full 16-slot buffer, or `substrate_compat` clearing subscriptions on a finality gap / re-warp), this task just exits and `warpSyncFinished`/`bootstrapComplete` are never emitted. The watchdog then reports a bogus `bootstrapTimeout`.

light-base/src/lib.rs
```diff
@@ -1277,17 +1302,363 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+                // reported instead of `BootstrapTimeout` when a finer signal
+                // is available.
+                let now = platform.now();
+                let no_peers = now.clone() - last_peer_seen.clone() >= NO_PEERS_TIMEOUT;
+                let bootstrap_stuck =
+                    !bootstrap_complete && now.clone() - started.clone() >= BOOTSTRAP_TIMEOUT;
+                let warp_stuck = !bootstrap_complete
+                    && last_warp_position.is_some()
+                    && now.clone() - last_warp_advance_seen.clone() >= WARP_NO_PROGRESS_TIMEOUT;
+                let sync_stuck = bootstrap_complete
```
skunert: `is_near_head_of_chain_heuristic()` documents that it must only be shown to the user, never used for logic. A client catching up after sleep is behind head for >60s while syncing fine and gets `syncNoProgress`. I'd drop the stall events from the wire entirely — thresholds like this are the consumer's call.

light-base/src/lib.rs
```diff
@@ -1277,17 +1302,363 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+
+    // Bootstrap sequence. Emissions are strictly ordered:
+    //   ModeDecision -> (WarpSyncProgress*)* -> WarpSyncFinished ->
+    //     BootstrapComplete
+    // The warp-progress poller is spawned from inside this task after
+    // `ModeDecision`, so it cannot race ahead. `warp_done` gates the poller
+    // so no stray `WarpSyncProgress` lands after `WarpSyncFinished`.
+    platform.spawn_task("lifecycle-bootstrap".into(), {
+        let lifecycle_service = lifecycle_service.clone();
+        let sync_service = sync_service.clone();
```
skunert: This holds a strong `Arc<SyncService>` while waiting, so after `remove_chain` the sync/network background stays alive until `subscribe_all` returns — which never happens if warp is stuck. The poller and watchdog use `Weak` for exactly this reason.

light-base/src/json_rpc_service/background.rs
```diff
@@ -3067,6 +3096,37 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                                }
+                            })
+                        });
+                    }
+
+                    methods::MethodCall::lifecycle_unstable_unfollow { subscription } => {
+                        // Removing the entry orphans the pump task: on its next event the
+                        // wake-up handler sees the missing subscription and drops the event
+                        // without re-arming. The call succeeds whether the id was known.
+                        me.lifecycle_subscriptions.remove(&*subscription);
```
skunert: Unfollow only removes the map entry. The pump task and its channel (sender still registered in the lifecycle service) are only freed after two further emits, which post-bootstrap may never come. Repeated follow/unfollow leaks tasks and channels without bound — unfollow should tear the pump down directly.

light-base/src/lib.rs
```diff
@@ -1277,17 +1302,363 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+                }
+
+                // Sample the near-head heuristic and warp position together
+                // to avoid a second round-trip through the sync service.
+                {
+                    let Some(sync_service_arc) = sync_service.upgrade() else {
+                        return;
+                    };
+                    let near_head = sync_service_arc.is_near_head_of_chain_heuristic().await;
+                    let warp_position = sync_service_arc.warp_sync_position().await;
```
skunert: This loop runs forever at 5s per chain even with zero subscribers, and once `bootstrap_complete` is set the `warp_sync_position()` result is never used by any branch. Could skip the warp query after bootstrap and pause the loop while nobody is subscribed.

light-base/src/lib.rs
BigTava: Thanks, you're right. I moved the emissions into the sync service where the state transitions happen.
`sync_service` now exposes `subscribe_bootstrap_status()`, which delivers a `BootstrapStatus` stream.

lib/src/json_rpc/service/client_main_task.rs
BigTava: You are right. It didn't surface in dotli since this is reachable by an embedder building on `smoldot::json_rpc::service`.

light-base/src/lib.rs
BigTava: Switched to `peers_list()`. Thanks!

light-base/src/json_rpc_service/background.rs
BigTava: Thanks, fixed. Broadcaster now flags evicted senders through the `lagged` boolean. Let me know if this is what you were looking for.

light-base/src/lib.rs
BigTava: Fixed, thanks!

light-base/src/lib.rs
BigTava: Agree, thanks!

light-base/src/lib.rs
BigTava: Already fixed as part of the [previous comment](https://github.com/paritytech/smoldot/pull/3317#discussion_r3784830271).

light-base/src/json_rpc_service/background.rs
BigTava: Thanks. The map now stores an `Arc<event_listener::Event>`. On unfollow the entry is removed and the event notified.

light-base/src/lib.rs
BigTava: > Could skip the warp query after bootstrap and pause the loop while nobody is subscribed.

Thanks! Already handled by the [previous rewrite](https://github.com/paritytech/smoldot/pull/3317#discussion_r3784830271). The watchdog derives warp progress from the `BootstrapStatus` stream it subscribes to at startup.

> This loop runs forever at 5s per chain even with zero subscribers

Left as-is because the watchdog `Stalled`/`Recovered` emissions feed the `LifecycleService` history ring. Pausing would make late subscribers miss stalls. Do you agree?

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -399,7 +402,19 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
                         &task.log_target,
                         "mode-decision; transition=Ready (WarpSyncFinished)",
                     );
+                    let finalized_height = task
+                        .sync
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!())
+                        .finalized_block_number();
                     drain_pending_subscriptions(&mut task);
+                    commit_bootstrap_mode(&mut task, BootstrapMode::WarpSync);
```
lrubasze, line 411: `modeDecision` is emitted after `warpSyncProgress`, docs promise the opposite.

The docs (`lifecycle_service.rs:71`, `methods.rs:1165`, `sync_service.rs:66`) say `modeDecision` precedes any warp event. But the only `ModeCommitted { WarpSync }` emission is in the `WarpSyncFinished` path (`substrate_compat.rs:411`), so a subscriber actually sees:

```
warpSyncProgress ...
modeDecision { mode: warpSync }
warpSyncFinished
```

A client that waits for `modeDecision` before rendering progress (as the docs suggest) receives progress events with no mode announced yet.

Suggested fix: emit `ModeCommitted { WarpSync }` on the first verified warp fragment, or reword the docs.

light-base/src/lib.rs
```diff
@@ -1277,17 +1299,250 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+                let now = platform.now();
+                let no_peers = now.clone() - last_peer_seen.clone() >= NO_PEERS_TIMEOUT;
+                let bootstrap_stuck =
+                    !bootstrap_complete && now.clone() - started.clone() >= BOOTSTRAP_TIMEOUT;
+                let warp_stuck = !bootstrap_complete
+                    && last_warp_position.is_some()
+                    && now.clone() - last_warp_advance_seen.clone() >= WARP_NO_PROGRESS_TIMEOUT;
+                let active = if no_peers {
+                    Some(lifecycle_service::StallReason::NoPeers)
+                } else if bootstrap_stuck {
```
lrubasze: Should we swap this and below branch to be inline with above comment?
```rust
// `WarpNoProgress` is reported instead of `BootstrapTimeout` when a finer signal is available.
```

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -399,7 +402,19 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                    emit_bootstrap_status(
```
lrubasze: Bootstrap mode announcement is protected against repeated commits (early return in`commit_bootstrap_mode`).
This is not the case for bootstrap status announcements - they will be resend if re-warp occurs.
I think we should align the approach. Allow or disallow repeated announcements for both, mode and status.

light-base/src/sync_service/substrate_compat.rs
BigTava: Fixed. Thanks!

light-base/src/lib.rs
BigTava: Fixed. Thanks!

light-base/src/sync_service/substrate_compat.rs
BigTava: Fixed. Added a guard so `WarpSyncFinished` is skipped on a post-bootstrap re-warp.

lrubasze:
Proposed a simplified design (state snapshot + updates instead of an event log) in https://github.com/paritytech/smoldot/issues/3301#issuecomment-5410713200.

lrubasze:
Closed in favor of #3353

Issue #3285, opened 2026-06-10
Internal metrics / introspection API


michalkucharczyk:
### Rationale

Today the only way to inspect smoldot's runtime state is via logs (`system_health`/`system_peers` aside, which expose almost nothing). Logs are often disabled in the field, verbose, and require reconstructing aggregate state by eye. This is especially painful in the browser, where you can't attach a debugger and bug reports come from users you can't instrument.

A small metrics surface lets us judge node state at a glance — peer churn, request failures, ban rate, sync progress, runtime-compilation cost — without enabling debug logging. Several useful internals are *already measured* and only `log!`ed (bytes/CPU, runtime compilation time, ping latency); this mostly aggregates what already exists.

### Solution sketch

- A hand-rolled `Arc<Metrics>` of counters/gauges, **no new dependencies**, incremented at existing event sites.
- Surfaced over the existing JSON-RPC interface (works identically in wasm and full-node; no Prometheus HTTP server needed):
  - **Pull:** `sudo_unstable_metrics()` → snapshot of all metric families.
  - **Push:** `sudo_unstable_metricsWatch(intervalMs)` → periodic **absolute** snapshots (the client diffs for rates; this enables simple graphs in an enclosing app's debug window).
- **Naming policy = Prometheus conventions** (counter/gauge distinction, `Total` suffix, base units, dimensions as labels), **camelCase** to match the repo. **No high-cardinality labels** (no peerId/multiaddr/hash) — aggregate only; per-peer detail stays in `system_peers`.

### Sample metrics (illustrative, not exhaustive)

- Network: `networkConnectionsOpenedTotal{direction}`, `networkRequestsTotal{protocol,outcome}`, `networkPeerBansTotal{reason}`, `networkAddressesDroppedTotal{reason}`, gauge `networkGossipPeersConnected{chain}`
- Sync: `syncBlocksVerifiedTotal{outcome}`, `syncVerificationErrorsTotal{kind,reason}`, gauges `syncBestBlockHeight` / `syncFinalizedBlockHeight`
- Runtime: `runtimeCompilationSeconds`, `runtimeCacheHitsTotal` (compilation timing already measured today)
- JSON-RPC / tx: `jsonrpcRequestsTotal{method,outcome}`, `transactionsDroppedTotal{reason}`
- Errors are first-class: classified by **who's at fault** (peer misbehavior / transient / local bug).

### Usage scenarios

[...]

PR #3318, opened 2026-07-31
Add `sudo_unstable_metrics` JSON-RPC function

lrubasze:
Covers https://github.com/paritytech/smoldot/issues/3285

Adds `sudo_unstable_metrics`, returning a snapshot of internal counters and gauges as a flat list of families (`name`, `type`, `entries` with labels). The set of metrics and labels is unstable and may change between versions.

Covered: network connections and discovery drops (process-wide); per-chain request counts/durations per protocol, peer bans by reason, gossip peers, block/finality verification outcomes, best/finalized/warp heights, runtime compilations (including cold-start ones), dropped transactions, JSON-RPC request count.

Notes:
- Counters are relaxed atomics, incremented next to the existing log calls. Label values derive from enums via strum; tests pin the exact label sets.
- **Breaking:** `ban_and_disconnect` now takes a `BanReason` enum instead of `&'static str` (`network_service` is public API).

## Use in e2e-tests

The chainhead e2e test can sample the new function during a run and render the time series as an HTML report:
```
SMOLDOT_METRICS_OUT=metrics.json ./run_chainhead_test.sh kusama-ah
```
This polls `sudo_unstable_metrics` on every chain every 5s (`SMOLDOT_METRICS_INTERVAL_MS` to change), writes the dump to `metrics.json`, and renders `metrics.html` (self-contained, no dependencies): block heights, warp progress, gossip peers, per-protocol request rate/failures/duration, runtime compilations, peer bans and discovery drops by reason, connection churn. The dump is written even when the test fails, so a bad run still produces a report - one paseo run with a warp stall was diagnosed straight from the charts (gossip peers at 0, handshake rate dropping to zero). Entirely opt-in: without `SMOLDOT_METRICS_OUT` nothing is sampled.


lrubasze:
Metrics example generated with below command:
```
SMOLDOT_METRICS_OUT=metrics.json ./run_chainhead_test.sh kusama-ah
```

<details>
  <summary>Returned json with snapshot</summary>
 
  ```json
    {
      "t": 1785492166420,
      "chain": "para",
      "metrics": [
        {
          "name": "networkConnectionsStartedTotal",
          "type": "counter",
          "entries": [
            {
              "value": 123
            }
          ]
        },
        {
          "name": "networkConnectionsHandshakesFinishedTotal",
          "type": "counter",
          "entries": [
            {
              "value": 22
            }
          ]
        },
        {
          "name": "networkConnectionsShutdownsTotal",
          "type": "counter",
          "entries": [
            {
              "value": 112
            }
          ]
        },
        {
          "name": "networkDiscoveryAddressesDroppedTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "reason": "peer-id-mismatch"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "not-supported"
              },
              "value": 8
            },
            {
              "labels": {
                "reason": "invalid"
              },
              "value": 0
            }
          ]
        },
        {
          "name": "networkRequestsTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "outcome": "success",
                "protocol": "blocks"
              },
              "value": 5
            },
            {
              "labels": {
                "outcome": "failure",
                "protocol": "blocks"
              },
              "value": 0
            },
            {
              "labels": {
                "outcome": "success",
                "protocol": "warp-sync"
              },
              "value": 0
            },
            {
              "labels": {
                "outcome": "failure",
                "protocol": "warp-sync"
              },
              "value": 0
            },
            {
              "labels": {
                "outcome": "success",
                "protocol": "storage-proof"
              },
              "value": 1
            },
            {
              "labels": {
                "outcome": "failure",
                "protocol": "storage-proof"
              },
              "value": 0
            },
            {
              "labels": {
                "outcome": "success",
                "protocol": "call-proof"
              },
              "value": 2
            },
            {
              "labels": {
                "outcome": "failure",
                "protocol": "call-proof"
              },
              "value": 0
            }
          ]
        },
        {
          "name": "networkRequestSecondsTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "protocol": "blocks"
              },
              "value": 0.588458
            },
            {
              "labels": {
                "protocol": "warp-sync"
              },
              "value": 0
            },
            {
              "labels": {
                "protocol": "storage-proof"
              },
              "value": 1.112125
            },
            {
              "labels": {
                "protocol": "call-proof"
              },
              "value": 0.039935
            }
          ]
        },
        {
          "name": "networkPeerBansTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "reason": "bad-block"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "bad-block-announce"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "bad-child-trie-root"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "bad-grandpa-commit"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "bad-justification"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "bad-merkle-proof"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "bad-warp-sync-fragment"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "invalid-call-proof"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "blocks-request-failed"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "call-proof-request-failed"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "child-storage-request-failed"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "storage-request-failed"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "warp-sync-request-failed"
              },
              "value": 0
            }
          ]
        },
        {
          "name": "networkGossipPeersConnected",
          "type": "gauge",
          "entries": [
            {
              "value": 4
            }
          ]
        },
        {
          "name": "syncBlocksVerifiedTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "outcome": "success"
              },
              "value": 35
            },
            {
              "labels": {
                "outcome": "failure"
              },
              "value": 0
            }
          ]
        },
        {
          "name": "syncFinalityProofsVerifiedTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "outcome": "success"
              },
              "value": 0
            },
            {
              "labels": {
                "outcome": "failure"
              },
              "value": 0
            }
          ]
        },
        {
          "name": "syncWarpFragmentsVerifiedTotal",
          "type": "counter",
          "entries": [
            {
              "value": 0
            }
          ]
        },
        {
          "name": "syncBestBlockHeight",
          "type": "gauge",
          "entries": [
            {
              "value": 19687690
            }
          ]
        },
        {
          "name": "syncFinalizedBlockHeight",
          "type": "gauge",
          "entries": [
            {
              "value": 19687672
            }
          ]
        },
        {
          "name": "syncWarpSyncHeight",
          "type": "gauge",
          "entries": [
            {
              "value": 0
            }
          ]
        },
        {
          "name": "syncWarpSyncTargetHeight",
          "type": "gauge",
          "entries": [
            {
              "value": 0
            }
          ]
        },
        {
          "name": "runtimeCompilationsTotal",
          "type": "counter",
          "entries": [
            {
              "value": 1
            }
          ]
        },
        {
          "name": "runtimeCompilationErrorsTotal",
          "type": "counter",
          "entries": [
            {
              "value": 0
            }
          ]
        },
        {
          "name": "runtimeCompilationSecondsTotal",
          "type": "counter",
          "entries": [
            {
              "value": 0.148646
            }
          ]
        },
        {
          "name": "runtimeCacheHitsTotal",
          "type": "counter",
          "entries": [
            {
              "value": 0
            }
          ]
        },
        {
          "name": "transactionsDroppedTotal",
          "type": "counter",
          "entries": [
            {
              "labels": {
                "reason": "gap-in-chain"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "max-pending-transactions-reached"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "invalid"
              },
              "value": 0
            },
            {
              "labels": {
                "reason": "validate-error"
              },
              "value": 0
            }
          ]
        },
        {
          "name": "jsonrpcRequestsTotal",
          "type": "counter",
          "entries": [
            {
              "value": 75
            }
          ]
        }
      ]
    }
  ```
</details>

<details>
  <summary>Rendered metrics using 5s snapshots as above</summary>

image
</details>

light-base/src/metrics.rs
```diff
@@ -0,0 +1,447 @@
...
+                    .zip(self.counters.iter().map(Counter::get)),
+            )
+            .finish()
+    }
+}
+
+/// Metrics of the network service. One instance per process, shared between all chains.
+#[derive(Debug, Default)]
+pub struct NetworkMetrics {
+    pub connections_started: Counter,
```
lexnv: These are peers that connected on the webrtc and the `connections_handshakes_finished` represent the peers that completed the handshakes? 🤔

lrubasze: These counters apply to any transport type (websocket, WebRTC, TCP).
`connection_started` is incremented upon each dial attempt, no matter the result.
`connections_handshakes_finished` is the subset of those that completed the handshake.
And the counters are per-connection, not per-peer - one peer can account for several attempts.

I'll update the doc comment.
Thanks!

light-base/src/transactions_service.rs
```diff
@@ -320,11 +324,15 @@ pub enum TransactionStatus {
 }
 
 /// See [`TransactionStatus::Dropped`].
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, strum::EnumDiscriminants)]
+#[strum_discriminants(name(DropReasonKind))]
+#[strum_discriminants(derive(strum::EnumIter, strum::IntoStaticStr))]
+#[strum_discriminants(strum(serialize_all = "kebab-case"))]
```
lexnv, line 333: This is for label info in grafana?

lexnv (approved):
Looks good on the Rust side (dont have much context on JS realm) 🙏

light-base/src/transactions_service.rs
lrubasze: Sort of - it's the `reason` label of the `transactionsDroppedTotal` metric in the `sudo_unstable_metrics` response (not Grafana-specific; anything consuming the metrics sees it, e.g. a Prometheus adapter or the e2e HTML report). `EnumDiscriminants` generates a payload-free `DropReasonKind` usable as a label, `EnumIter` gives the snapshot the full label set (zero counts included), `IntoStaticStr` + kebab-case turn variant names into the label strings, and the `disabled` variants (`Finalized` = success path, `Crashed`) are deliberately not counted. A test pins the exact label strings so a rename can't silently change the metric.

Added docs for this. Thanks!

light-base/src/network_service.rs
```diff
@@ -238,16 +247,23 @@ impl<TPlat: PlatformRef> NetworkService<TPlat> {
         Arc::new(NetworkService {
             messages_tx: main_messages_tx,
             platform: config.platform,
+            metrics,
         })
     }
 
+    /// Returns the process-wide network metrics.
+    pub fn metrics(&self) -> &Arc<metrics::NetworkMetrics> {
```
michalkucharczyk, line 255: do we actually need `Arc` here?

e2e-tests/render_metrics_html.mjs
```diff
@@ -0,0 +1,632 @@
...
+    zeroBase: true,
+    series: PROTOCOLS.map((p, i) => ({
+      name: p,
+      slot: i,
+      points: meanSeries(
+        chain,
+        "networkRequestSecondsTotal", { protocol: p },
+        "networkRequestsTotal", { protocol: p, outcome: "success" },
+      ),
+    })),
```
michalkucharczyk, line 213: `networkRequestSecondsTotal`  includes failures, but we only divide by `success`.

michalkucharczyk: Related code: https://github.com/paritytech/smoldot/blob/6fabafe786e670223cbacb4b37774d5882a0ab46/light-base/src/metrics.rs#L97-L105

light-base/src/network_service.rs
```diff
@@ -550,7 +618,11 @@ impl<TPlat: PlatformRef> NetworkServiceChain<TPlat> {
             .await
             .unwrap();
 
-        rx.await.unwrap()
+        let result = rx.await.unwrap();
+        self.metrics
+            .storage_proof_requests
```
michalkucharczyk, line 623: `child_storage_proof_request` ? Or do we intentionally merge them?

light-base/src/network_service.rs
```diff
@@ -315,6 +336,30 @@ pub enum BanSeverity {
...
+/// Reason for banning a peer. Printed in the logs and used as the `reason` label
+/// of the `networkPeerBansTotal` metric.
+// Strum derives are explained on [`crate::metrics::MetricLabel`]; `Display`
+// additionally prints the same string in log messages.
+#[derive(
+    Debug, Copy, Clone, PartialEq, Eq, strum::Display, strum::EnumIter, strum::IntoStaticStr,
+)]
+#[strum(serialize_all = "kebab-case")]
+pub enum BanReason {
+    BadBlock,
```
michalkucharczyk, line 348: never used?

light-base/src/metrics.rs
```diff
@@ -0,0 +1,462 @@
...
+    fmt,
+    marker::PhantomData,
+    sync::atomic::{AtomicU32, Ordering},
+};
+use smoldot::json_rpc::methods;
+use strum::IntoEnumIterator;
+
+// 32-bit atomics so that the metrics also compile on targets without 64-bit
+// atomics (e.g. `thumbv7m-none-eabi`); durations are tracked in milliseconds
+// to make the range acceptable.
```
michalkucharczyk, line 40: Nit: not sure if limiting to 32-bits on all platforms is OK here. Not saying it shall be 64bit - but maybe this deserves a second though if we can do something smarter here. E.g. 2^32 ~ 51 days, technically duration could accumulate to such number. But maybe it is ok.

michalkucharczyk (approved):
Skimmed the implementation, looks good. Left some nits.

light-base/src/network_service.rs
lrubasze: Yes, the `Arc` is needed: the counters are one live shared object - the network task writes them, and the JSON-RPC background task reads them on every `sudo_unstable_metrics` request. A clone would be a frozen copy, and a reference can't be held by a spawned task. `metrics()` itself is only called once, at chain startup, to wire the handle through. 
I changed it to return the `Arc` by value instead of by reference so the code reads better.

e2e-tests/render_metrics_html.mjs
lrubasze: Good catch! Thanks 👍

light-base/src/network_service.rs
lrubasze: Missed it. `child_storage_proof_request` metric added.

light-base/src/network_service.rs
lrubasze: Good catch - it was only referenced from a commented-out TODO block; removed.

light-base/src/metrics.rs
lrubasze: Alternatively we could widen the counters to `AtomicU64` with `#[cfg(target_has_atomic = "64")]` with the current `AtomicU32` as fallback. Every realistic deployment target (`wasm32`, `x86-64`, `aarch64`, 32-bit ARM/x86 on mobile) has 64-bit atomics, so the limit effectively goes away; the fallback stays only for the niche no-std targets we intentionally keep building (e.g. microcontrollers (Cortex-M), 32-bit RISC-V, and old MIPS/PowerPC/ARMv5 Linux).
@michalkucharczyk WDYT?

lrubasze: Other ideas:
- `portable-atomic` crate, which emulates `AtomicU64` everywhere. Adds a dependency, and on `thumbv7m` needs a critical-section provider or the single-core flag. Heavy for this.
- Replace the duration sums with latency histograms (per-protocol). Buckets hold counts, so u32 is enough, and they give percentiles and timeout visibility that a mean cannot. It's a new metric type to be implemented. Slightly heavier than counter or gauge type.

skunert: My 2c are that we should for sure widen this to U64 with U32 as fallback. I expect 99% of usage supports the wider format. 

Some use-cases are for example that you could use smoldot and embed it in some rust app. Or have a long running instance as target for the `--relay-chain-rpc-url` CLI param.

lrubasze: That makes sense. Thanks!

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -988,6 +1002,10 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
                     unreachable!()
                 };
 
+                task.metrics
+                    .sync_best_block_height
```
skunert, line 1018: Does this work correctly when starting from a checkpoint?

light-base/src/metrics.rs
```diff
@@ -0,0 +1,489 @@
...
+    network_service::{BanReason, DiscoveredAddressDropReason},
+    transactions_service::DropReasonKind,
+};
+use alloc::{borrow::Cow, boxed::Box, collections::BTreeMap, vec::Vec};
+use core::{fmt, marker::PhantomData, sync::atomic::Ordering};
+use smoldot::json_rpc::methods;
+use strum::IntoEnumIterator;
+
+// 64-bit atomics where the target has them (every realistic deployment target: wasm32,
+// x86-64, aarch64, 32-bit ARM/x86 with `std`), 32-bit saturating atomics as a fallback
```
skunert: In fact we don't saturate, but fetch_add wraps right?

lrubasze: You're right. FIxed, now we use saturating_add for both atomic types.

light-base/src/sync_service/substrate_compat.rs
lrubasze: Good catch, it did not. The gauges were only set when the best or finalized block changed, so they read 0 during the whole warp sync from a checkpoint (and for the parachain until the relay was synced). Fixed, now both sync tasks set them from the sync state at startup.

skunert (approved):

Issue #3322, opened 2026-08-03
wasm-node: uncaught InvalidStateError when sending on an RTCDataChannel that is no longer open


BigTava:
We hit this in production with smoldot 3.3.2 running in Chrome 149 on macOS:

```
InvalidStateError: Failed to execute 'send' on 'RTCDataChannel': RTCDataChannel.readyState is not 'open'
    at <object>.send (smoldot/dist/mjs/no-auto-bytecode-browser.js:461)  channel.channel.send(new Blob(data));
    at o (smoldot/dist/mjs/internals/client.js:190)  connection.send(event.data, event.streamId);
    at t.onmessage (smoldot/dist/mjs/internals/remote-instance.js:144)  config.eventCallback(message);
```

The WebRTC `send` handler in `wasm-node/javascript/src/no-auto-bytecode-browser.ts` calls `channel.channel.send()` without checking `readyState`:

```ts
send: (data: Array<Uint8Array>, streamId: number): void => {
    const channel = state.dataChannels.get(streamId)!;
    for (const buffer of data) {
        channel.bufferedBytes += buffer.length;
    }
    channel.channel.send(new Blob(data));
},
```

A data channel can leave the `open` state before its `close` event is dispatched, and the stream reset is only reported to the core from that event handler. Until it runs, the core still believes the stream is writable and can issue a send, which then throws. In our setup the core talks to the connection through a message port (`remote-instance.js` in the trace above), which widens that window. The exception escapes from the message event handler, so it surfaces as an uncaught error in the page.

This looks related to the `open` event race fixed in #3305, but on the closing side of the channel lifecycle.

Skipping the send when `readyState` is not `open` seems safe, since the `close` handler will follow up with `onStreamReset` for the same stream. I am happy to open a PR if that is the fix you would want.

lrubasze:
Reported in Sentry: [DOTLI-BF](https://paritytech.sentry.io/issues/DOTLI-BF) (smoldot 3.3.2, Chrome). Fixed by #3324, ships in the next npm release.

PR #3326, opened 2026-08-05
wasm-node: fix dropping of stale messages for stream id 0

lrubasze:
## Problem

In remote-instance mode, both sides of the MessagePort drop commands addressed to a connection or stream that was already reset locally, so messages that cross in flight don't act on dead objects. The client-side check tested `message.streamId && ...`, which skips stream id 0 — and stream ids start at 0, so the first substream of every WebRTC connection (the one carrying the Noise handshake) was exempt from the check. A `stream-send` racing that stream's reset then reached the platform after the `dataChannels` entry was removed and threw an uncaught `TypeError`.

Found while fixing #3322; same bug in the `stream-send-close` handler.

## Fix

Test `message.streamId !== undefined` instead, as the server-side handlers in the same file already do.

## Test

New Node unit test `test/remote-instance-already-reset.mjs` — it plays the server end of the `MessagePort` protocol by hand and asserts commands for already-reset streams never reach `eventCallback`. The two stream-0 cases fail without the fix and pass with it; a live-stream case and a stream-3 case pin that normal traffic still flows and the check itself works.

### DIFF
--- a/wasm-node/javascript/src/internals/remote-instance.ts
+++ b/wasm-node/javascript/src/internals/remote-instance.ts
@@ -130,8 +130,9 @@ export async function connectToInstanceServer(config: ConnectConfig): Promise<in
                 // The connection might have been reset locally in the past.
                 if (!state.connections.has(message.connectionId))
                     return;
-                // The stream might have been reset locally in the past.
-                if (message.streamId && !state.connections.get(message.connectionId)!.has(message.streamId))
+                // The stream might have been reset locally in the past. Note that stream
+                // ids start at 0, so a truthiness check would exempt the first substream.
+                if (message.streamId !== undefined && !state.connections.get(message.connectionId)!.has(message.streamId))
                     return;
                 break;
             }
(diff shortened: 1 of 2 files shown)


lexnv (approved):

wasm-node/javascript/src/internals/remote-instance.ts
```diff
@@ -130,17 +130,19 @@ export async function connectToInstanceServer(config: ConnectConfig): Promise<in
                 // The connection might have been reset locally in the past.
                 if (!state.connections.has(message.connectionId))
                     return;
-                // The stream might have been reset locally in the past.
-                if (message.streamId && !state.connections.get(message.connectionId)!.has(message.streamId))
+                // The stream might have been reset locally in the past. Note that stream
+                // ids start at 0, so a truthiness check would exempt the first substream.
```
bkchr:
```suggestion
                // The stream might have been reset locally in the past.
```

wasm-node/javascript/src/internals/remote-instance.ts
```diff
@@ -130,17 +130,19 @@ export async function connectToInstanceServer(config: ConnectConfig): Promise<in
...
                 break;
             }
             case "stream-send-close": {
                 // The connection might have been reset locally in the past.
                 if (!state.connections.has(message.connectionId))
                     return;
-                // The stream might have been reset locally in the past.
-                if (message.streamId && !state.connections.get(message.connectionId)!.has(message.streamId))
+                // The stream might have been reset locally in the past. Note that stream
+                // ids start at 0, so a truthiness check would exempt the first substream.
```
bkchr:
```suggestion
                // The stream might have been reset locally in the past.
```

bkchr (approved):
These claude comments are not really useful, at least for me. They just confused me more.

wasm-node/javascript/src/internals/remote-instance.ts
lrubasze: Done

wasm-node/javascript/src/internals/remote-instance.ts
lrubasze: Done

skunert (approved):

PR #3329, opened 2026-08-05
libp2p: fix panic on duplicate reject of a reset notifications substream

### LINKED ISSUE #3304: Native: panic in `Yamux::index_mut` from `single_stream_connection_task` (iOS, smoldot-light)

lrubasze:
Fixes #3304.
Fixes BREVITY-MORPHEUS-4C
Fixes BREVITY-MORPHEUS-4S

## Problem

When the remote resets an inbound notifications substream while a coordinator message about that substream is still in flight, the message can arrive after the substream was removed, and the connection panics.

Two cases:
- a duplicate `RejectInNotifications` - panics in `Yamux::index_mut` (yamux.rs:1736); this is the signature seen in the field via Sentry on iOS/macOS
- a CloseInNotifications sent right after an accept - hits an explicit panic in `close_in_notifications_substream`

## Fix

`reject_in_notifications_substream` and `close_in_notifications_substream` return early if the substream is gone. Applied to both single-stream and multi-stream connections.

## Test

New collection-layer tests reproduce both panics without the fix and assert the connection survives with it. Other coordinator message types were audited; no other case of this pattern found.

### DIFF
--- a/lib/src/libp2p/connection/established/single_stream.rs
+++ b/lib/src/libp2p/connection/established/single_stream.rs
@@ -891,8 +891,10 @@ where
             _ => panic!(),
         };
 
+        // The substream might have been reset by the remote (and thus removed from the yamux state
+        // machine) while the accept and close messages were in flight. See issue #3304.
         if !self.inner.yamux.has_substream(substream_id) {
-            panic!()
+            return;
         }
 
         self.inner.yamux[substream_id]
--- a/lib/src/libp2p/connection/established/single_stream.rs
+++ b/lib/src/libp2p/connection/established/single_stream.rs
@@ -771,6 +771,12 @@ where
             _ => panic!(),
         };
 
+        // The substream might have been reset by the remote (and thus removed from the yamux state
+        // machine) at the same time as the coordinator decided to reject it. See issue #3304.
+        if !self.inner.yamux.has_substream(substream_id) {
+            return;
+        }
+
         self.inner.yamux[substream_id]
             .as_mut()
             .unwrap()
--- a/lib/src/libp2p/collection.rs
+++ b/lib/src/libp2p/collection.rs
@@ -78,6 +78,9 @@ pub use single_stream::SingleStreamConnectionTask;
 mod multi_stream;
 mod single_stream;
 
+#[cfg(test)]
+mod tests;
+
 /// What kind of handshake to perform on the newly-added connection.
 pub enum SingleStreamHandshakeKind<'a> {
     /// Use the multistream-select protocol to negotiate the Noise encryption, then use the
--- a/lib/src/libp2p/collection/tests.rs
+++ b/lib/src/libp2p/collection/tests.rs
@@ -18,11 +18,11 @@
 #![cfg(test)]
 
 use super::{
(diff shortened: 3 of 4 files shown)


lib/src/libp2p/connection/established/single_stream.rs
```diff
@@ -771,6 +771,12 @@ where
             _ => panic!(),
         };
 
+        // The substream might have been reset by the remote (and thus removed from the yamux state
+        // machine) at the same time as the coordinator decided to reject it. See issue #3304.
+        if !self.inner.yamux.has_substream(substream_id) {
```
skunert, line 776: I think we have the same bug in multi_stream.rs. We can arrive here: https://github.com/paritytech/smoldot/blob/02d14856b435f164dbd0531049be5348b4dba63f/lib/src/libp2p/connection/established/multi_stream.rs#L778

When the stream has been reset we panic there.

lrubasze: Nice catch! Thanks 👍

lrubasze: Added guard and copied test harness for multi stream as well

dmitry-markin (approved):

lexnv (approved):

lexnv:
Should we fix it in `close_in_notifications_substream` as well?

lrubasze:
> Should we fix it in `close_in_notifications_substream` as well?

Nice catch @lexnv! Thanks!

PR #3331, opened 2026-08-05
executor: implement RFC-0163 elliptic curve host functions

### LINKED ISSUE #3328: unimplemented RFC-0163 EC host functions block bandersnatch ring-VRF runtimes
## Problem Statement

Tools that execute real Substrate runtimes on top of smoldot — chopsticks being the common case — cannot run any Polkadot chain that uses proof-of-personhood or ring VRFs. Those runtimes import a small family of elliptic-curve host functions from `sp-crypto-ec-utils` that smoldot does not implement, so every extrinsic that verifies a ring membership proof dies inside the block builder.

Because such tools normally enable `allow_unresolved_imports`, the chain starts up cleanly and unrelated transactions finalize normally. The failure only shows up as transactions that are accepted into the pool, never included in a block, and eventually time out client-side with no chain-side error — which reads as flakiness or a timing bug rather than a missing capability.

## Goal

smoldot implements the [RFC-0163 `ext_host_calls_*`](https://github.com/polkadot-fellows/RFCs/blob/main/text/0163-ec-host-functions.md) elliptic-curve host functions with semantics and wire format byte-compatible with `sp-crypto-ec-utils`, so that a runtime importing them performs ring-VRF verification correctly instead of raising `UnresolvedFunctionCalled`.

## Evidence: what happens today

Forking a People/personhood parachain and submitting an attestation extrinsic. The runtime instantiates fine; the block builder then fails on every attempt, and the same logical transaction is re-queued and retried on each subsequent block:

```
[block-builder]: Failed to apply extrinsic {} Error: Called unresolved function
  `env`:`ext_host_calls_ed_on_bls12_381_bandersnatch_mul_version_1`
```

The runtime's wasm import table (read from `:code`, zstd-unwrapped) contains 52 host-function imports, of which six are unimplemented by smoldot:

```
env::ext_host_calls_bls12_381_msm_g1_version_1
env::ext_host_calls_bls12_381_mul_g1_version_1
env::ext_host_calls_bls12_381_mul_g2_version_1
env::ext_host_calls_bls12_381_multi_miller_loop_version_1
env::ext_host_calls_bls12_381_final_exponentiation_version_1
env::ext_host_calls_ed_on_bls12_381_bandersnatch_mul_version_1
```

Only the last one was ever reached in a captured run — the five BLS12-381 symbols are imported but never called on the personhood path — so a single function unblocks the surface in practice.

## Evidence: why it is deferred to call time

`lib/src/executor/host.rs:834-845` at `paritytech/smoldot@355b3f42e63ee2524784b54c556162c3d433a6e0`:

```rust
let host_fn = match self.inner.common.registered_functions.get(id) {
    Some(FunctionImport::Resolved(f)) => *f,
    Some(FunctionImport::Unresolved { name, module }) => {
        return HostVm::Error {
            error: Error::UnresolvedFunctionCalled {
                function: name.clone(),
                module_name: module.clone(),
            },
            prototype: self.inner.into_prototype(),
        };
    }
    None => unreachable!(),
```

With `allow_unresolved_imports: true` the import is bound as `FunctionImport::Unresolved` at instantiation (`:368-375`) and only errors if actually called. The resulting `HostVm::Error` is terminal — there is no resumable variant an embedder could service in its own code, so this cannot be worked around outside smoldot.

## Orientation

- `lib/src/executor/host.rs` — the `HostFunction` enum and `HostFunction::by_name` registry, plus the per-function dispatch.
- Closest precedent in the same file: `ext_trie_blake2_256_verify_proof_version_2` (`:2112`, `:2115` at the pinned commit), added so that embedders could build blocks containing proof-verification inherents.
- Reference semantics live in `sp-crypto-ec-utils` in polkadot-sdk: a `#[runtime_interface] pub trait HostCalls`, which is what gives every symbol its `ext_host_calls_` prefix.
- Encoding is handled by the `ark-scale` crate; the curve implementation is `ark-ed-on-bls12-381-bandersnatch`.

The signature of the function that actually matters:

```rust
fn ed_on_bls12_381_bandersnatch_mul(
    base:   PassFatPointerAndRead<&[u8]>,      // ArkScale<EdwardsAffine>
    scalar: PassFatPointerAndRead<&[u8]>,      // ArkScale<BigInteger>
    out:    PassFatPointerAndWrite<&mut [u8]>, // ArkScale<EdwardsAffine>
) -> HostcallResult
```

Its body upstream is a single call to a twisted-Edwards multiply helper parameterised on `EdwardsConfig`.

## Definitions

**Degenerate point.** Bandersnatch is an *incomplete* twisted Edwards curve. With cofactor-admixed (non-prime-order-subgroup) inputs, the HWCD addition/doubling formulas can produce a projective point with `z = 0`. Such a point has no affine representative, and arkworks' `From<Projective> for Affine` panics on `z.inverse().unwrap()`. Upstream signals this across the FFI boundary as a dedicated degenerate-point error; the runtime side substitutes the all-zero projective point `(0, 0, 0, 0)`, which is not a valid curve point and is rejected by any downstream subgroup or validity check.

**ArkScale usage.** The runtime interface encodes every value with the `ArkScale` usage parameter set to *not-validated* and *not-compressed*. Any other combination is a wire-format mismatch.

## Non-Counting Outcomes

- **Resolving the imports without implementing them.** Binding the symbols so instantiation succeeds while returning a stub, zero, or the identity point. Ring proofs then fail verification *silently* — strictly worse than today's hard error, because a fixture would appear to work while producing wrong consensus results.
- **Implementing the short-Weierstrass multiply for the `_mul` entry point.** `ed_on_bls12_381_bandersnatch_mul` is twisted Edwards. Short-Weierstrass variants exist in the same upstream crate and are a different code path; using one for the other yields wrong results, not an error.
- **Encoding with `ArkScale` defaults.** Validated or compressed encoding is a byte-format mismatch against the deployed runtime side.
- **Unconditional projective-to-affine conversion.** Ignoring the `z = 0` case, which panics rather than returning the documented degenerate signal.
- **Self-referential test vectors.** Generating expected values with the same implementation under test passes by construction. Vectors must originate from `sp-crypto-ec-utils` or a native runtime execution.

## Acceptance Criteria

- [ ] **(gatekeeper)** A test in the executor test-suite feeds a `(base, scalar, expected)` triple produced by `sp-crypto-ec-utils` — not by this implementation — through the new host function and asserts byte equality of the encoded output. The test fails when the implementation is reverted.
- [ ] **(critical)** A test using a non-prime-order-subgroup input drives the result to `z = 0` and asserts the documented degenerate-point signal is returned, with no panic.
- [ ] Encoded output is byte-identical to `ArkScale<EdwardsAffine>` under not-validated / not-compressed usage, with no sentinel bit and no separate projective codec.
- [ ] `HostFunction::by_name` resolves `ext_host_calls_ed_on_bls12_381_bandersnatch_mul_version_1` with a signature matching the runtime's import, so a runtime importing it instantiates with `allow_unresolved_imports: false`.
- [ ] The five BLS12-381 symbols listed above are either implemented and covered by the same class of external-vector test, or explicitly named as out of scope in the change description. Silently leaving them unresolved without saying so does not satisfy this.
- [ ] A runtime performing bandersnatch ring-membership verification applies the extrinsic successfully under an embedder that executes full runtimes, where it previously raised `UnresolvedFunctionCalled`.

## Boundaries

**Never**

- Change the wire format or introduce a sentinel encoding. The runtime side is already deployed and expects `ArkScale<EdwardsAffine>` exactly.
- Substitute an approximate or short-circuit result for any input class; a wrong curve point is a consensus fault, and an error is preferable to a plausible-looking wrong answer.

**Ask first**

- Whether arkworks should sit behind an off-by-default feature. RFC-0163 exists precisely because these operations are too slow inside the runtime VM, and pulling the arkworks algebra stack into default builds adds real weight to a light client. A feature enabled only by embedders that execute full runtimes may be the better shape, and that trade-off is the maintainers' call, not the implementer's.

lrubasze:
Closes #3328

Implements the [RFC-0163](https://github.com/polkadot-fellows/RFCs/blob/main/text/0163-ec-host-functions.md) `ext_host_calls_*` host functions needed by runtimes that verify bandersnatch ring-VRF proofs (proof-of-personhood chains).

**Implemented:**
- `ed_on_bls12_381_bandersnatch_mul` and `_msm`
- `bls12_381_mul_g1`, `mul_g2`, `msm_g1`, `msm_g2`, `multi_miller_loop`, `final_exponentiation`
- The BLS12-381 functions are required, not optional: the personhood runtime routes its ring-proof KZG verification through them, right after the VRF step that calls the bandersnatch mul.
- Pallas and Vesta (rest of [RFC-0163](https://github.com/polkadot-fellows/RFCs/blob/main/text/0163-ec-host-functions.md)) have no caller today; registered with correct signatures, return `HostFunctionNotImplemented` if called.

**Compatibility:**
- Wire format byte-compatible with `sp-crypto-ec-utils` (ArkScale, not-validated / not-compressed).
- Degenerate-point error (code 4) for `z = 0` results on non-subgroup bandersnatch inputs, no panic.

**Tests (26), two independent vector sources:**
- Byte vectors from a replica of the `sp-crypto-ec-utils` helpers using the real `ark-scale` crate - checks the wire format
- EIP-2537 vectors from the Ethereum precompile suite - checks the curve arithmetic independently of arkworks.

**Size:** (wasm-node distribution build, `min-size-release`):
|          | raw .wasm | deflated |
|----------|-----------|----------|
| `main`   | 5,193 kB  | 2,298 kB |
| this PR  | 5,347 kB  | 2,351 kB |
| delta    | +154 kB (+2.96%) | +53 kB (+2.30%) |


ryanleecode (approved):

skunert (approved):
Double checked once more against substrate impl. Looks good.

lib/src/executor/host/elliptic_curves.rs
```diff
@@ -0,0 +1,174 @@
...
+/// an encoded target field element.
+pub(super) fn bls12_381_multi_miller_loop(
+    g1: &[u8],
+    g2: &[u8],
+    out_len: usize,
+) -> Result<Vec<u8>, u32> {
+    let g1 = decode::<Vec<ark_bls12_381::G1Affine>>(g1)?;
+    let g2 = decode::<Vec<ark_bls12_381::G2Affine>>(g2)?;
+    let result = Bls12_381::multi_miller_loop(g1, g2);
+    encode(&result.0, out_len)
```
davxy, line 165: We need the same fix as: https://github.com/paritytech/polkadot-sdk/pull/12834

ryanleecode:
Can we do this as a follow up. Would like a release now to unblock my CI :)

skunert:
> Can we do this as a follow up. Would like a release now to unblock my CI :)

Yes I think it is okay to merge here. If we introduce the fix in substrate we can do it here too.

lib/src/executor/host/elliptic_curves.rs
skunert: done

PR #3334, opened 2026-08-07
statement-store: Stop panicking on affinity updates when the statement protocol is disabled

AndreiEres:
Calling `statement_subscribeStatement` on a chain that runs without the statement protocol panics: inserting the subscription schedules a topic-affinity update, and the scheduler `expect`s the protocol config to be present. The scheduler now returns early instead — subscriptions can still be created in that case, and there is no affinity to advertise.


michalkucharczyk (approved):

light-base/src/json_rpc_service/background.rs
```diff
@@ -833,6 +835,8 @@ pub(super) async fn run<TPlat: PlatformRef>(
                 me.statement_affinity_stale = false;
                 me.last_statement_affinity_update = Some(me.platform.now());
 
+                // This wake-up only ever fires from `schedule_statement_affinity_update`, which
+                // returns without scheduling anything when the config is absent.
```
michalkucharczyk: not sure we need this comment?
```suggestion
```

lexnv (approved):

dmitry-markin (approved):

PR #3344, opened 2026-08-12
statement-store: validate statement_submit like a full node

### LINKED ISSUE #3349: statement-store: statement_* answers diverge from a full node's
smoldot's `statement_*` methods give answers a full node would not. A light client can't reproduce store-dependent answers (`rejected/*`, `storeFull`, a replay of stored statements), but every check and every answer that needs no chain state or store contents should match the full node, and the parity should be pinned by a test comparing both against the same inputs.

## statement_submit

Done in #3344. smoldot skipped the cheap validation a full node runs in `Store::submit` (expiry, size, proof presence) and reported broadcast success from the gossip-set size rather than actual sends, so invalid statements came back as `new` and undecodable payloads and zero-reach broadcasts surfaced as successes instead of JSON-RPC errors.

Left after #3344: smoldot still decoded the `Proof::OnChain` variant that polkadot-sdk removed, answered `new` and broadcast a statement every peer rejects. #3370 drops the variant, so the bytes fail to decode and the answer is `7001` like a full node's.

## statement_subscribeStatement

#3371. smoldot sent nothing after the subscription id, where a full node replays its store's matches in batches with `remaining` and ends with an empty batch with `remaining: 0`, so a client waiting for the end of the replay hung. It also forwarded every decodable gossiped statement, where a full node only delivers statements its store accepted, so a subscriber could receive an expired statement or one without a proof. The PR sends the empty initial batch and drops expired and proof-less statements before delivery, with the same checks `statement_submit` runs.

AndreiEres:
Closes https://github.com/paritytech/smoldot/issues/3349

`statement_submit` updated to behave like polkadot-sdk's `Store::submit`.

## Fixes

* **Added cheap validation.** Checks now run in `Store::submit`'s order — expiry, size, proof — so a statement failing several is given the same reason a full node would give.
* **Counted broadcast success on `sent`, not `total`.** A peer can sit in the gossip set with its statement substream absent or its queue full, so only `sent` tells whether the statement reached anyone. The error carries the connected-peer count, separating "no peers" from "peers took nothing".
* **Answered unreportable failures with JSON-RPC errors.** A payload that doesn't decode, or a statement that reached no peer, carries `7001` — polkadot-sdk's statement-store error code.

`StatementSubmitResult` becomes `New | Invalid(InvalidReason)`, mirroring upstream's `SubmitResult` minus the store-dependent statuses. `InvalidReason` mirrors upstream's, `snake_case` payload fields and all.

## Answer parity

A new zombienet job submits each statement below to both smoldot and a collator, then compares the answers. Divergences are pinned rather than skipped, so closing a gap — or opening one — fails the test instead of drifting.

| Case | Full node | smoldot | |
|---|---|---|---|
| `new` | `new` | `new` | ✅ |
| `known` | `new` | `new` | ✅ |
| `invalid_no_proof` | `invalid` / `noProof` | `invalid` / `noProof` | ✅ |
| `invalid_encoding_too_large` | `invalid` / `encodingTooLarge` | `invalid` / `encodingTooLarge` | ✅ |
| `invalid_already_expired` | `invalid` / `alreadyExpired` | `invalid` / `alreadyExpired` | ✅ |
| `decode_failure` | error `7001` | error `7001` | ✅ |
| `rejected_no_allowance` | `rejected` / `noAllowance` | `new` | allowance is chain state |
| `rejected_data_too_large` | `rejected` / `dataTooLarge` | `new` | allowance is chain state |
| `rejected_channel_priority_too_low` | `rejected` / `channelPriorityTooLow` | `new` | needs a store's contents |
| `rejected_account_full` | `rejected` / `accountFull` | `new` | needs a store's contents |
| `invalid_bad_proof` | `invalid` / `badProof` | `new` | verifying a signature costs CPU |

`known` matches because the status is reserved for a source that may not resubmit, and an RPC submission always may — so a full node re-accepts as `new` too, store or no store.

Three upstream answers get no case: `storeFull` needs the global 2 GiB total pushed through a collator, `internalError` reports a database failure a submission cannot provoke, and `knownExpired` needs a stored statement to expire mid-test.

`StatementSubmitResult`'s doc comment accounts for each declined answer and why, and the test above is its executable half.

## Breaking

Every fix above changes an answer a client already gets:

* Two shapes that were successful results are now JSON-RPC errors, so a resolved promise becomes a rejected one.
* Callers passing a small opaque `expiry` now get `alreadyExpired` rather than a broadcast, since the upper 32 bits are read as a UNIX timestamp.
* The `smoldot` crate drops `InternalError` and reshapes `Invalid` from a struct variant to a newtype — **a major bump** under `docs/RELEASING.md`.


light-base/src/json_rpc_service/statement.rs
```diff
@@ -71,32 +70,99 @@ impl StatementProtocolConfig {
     }
 }
 
+/// JSON-RPC error code answering a submission the statement store couldn't process.
+///
+/// Matches polkadot-sdk.
+pub const STATEMENT_STORE_ERROR_CODE: i64 = 7001;
```
dmitry-markin, line 76: Quickly skimming over https://github.com/paritytech/json-rpc-interface-spec/blob/main/src/api/statement_unstable_submit.md it looks like this error code is not listed there, but should be.

light-base/src/json_rpc_service/statement.rs
```diff
@@ -71,32 +70,99 @@ impl StatementProtocolConfig {
...
+
+/// Submission failure reported as a JSON-RPC error
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum StatementSubmitError {
+    /// The submitted bytes don't decode into a statement.
+    InvalidEncoding,
+    /// The statement is valid but reached none of the gossip-connected peers. A light
+    /// client keeps no store, so a statement nobody received is nowhere at all.
```
dmitry-markin:
```suggestion
    /// client keeps no store, so a statement nobody received is permanently lost.
```
?

dmitry-markin (approved):

light-base/src/json_rpc_service/statement.rs
AndreiEres: This is not about this API, we want to adapt the API existed before first.

alexggh (approved):

PR #3345, opened 2026-08-13
test: update and re-enable browser tests

gab8i:
This includes all the changes needed to adapt to this: https://github.com/paritytech/polkadot-sdk/pull/12870, which also includes a new str0m version which in its turn contains sctp fixes, thus allowing to re-enable browser tests.

NOTE: this requires a new `POLKADOT_SDK_MASTER_SHA` which will only be available once the polkadot pr will be merged.


lexnv (approved):

gab8i:
While trying to make CI to pass a couple of issues have been found. Solutions to them are here https://github.com/paritytech/smoldot/pull/3361!

e2e-tests/src/network.rs
```diff
@@ -35,41 +35,27 @@ use zombienet_sdk::{Arg, LocalFileSystem, Network, NetworkConfig, NetworkConfigB
...
-        _ => unreachable!("name: {name}, not associated to any udp port"),
-    }
-}
-
-/// Looks up the fixed WebRTC UDP port assigned to `name` and returns the CLI
-/// args that make a substrate node listen for WebRTC on it.
-pub fn listener_args(name: &str) -> Vec<Arg> {
-    let udp_port = webrtc_udp_port(name);
+/// First UDP port handed out to a `webrtc-direct` listener.
+const WEBRTC_BASE_UDP_PORT: u16 = 30333;
```
skunert: For multiple runs on the same machine this could still conflict. Maybe would be a better idea to use 0 and let the node pick a port, then read it  back through `system_localListenAddresses` RPC which we already query. Not sure if this is currently supported, but just an idea.

.github/zombienet-env
```diff
@@ -12,11 +12,14 @@ ZOMBIENET_SDK_DEFAULT_RUNNER_FOR_NATIVE=parity-zombienet-native-default
...
-# own zombienet CI runs on:
+# against. The tests need a build where WebRTC is available, so binaries
+# are extracted from the per-merge Docker images that polkadot-sdk's own
+# zombienet CI runs on:
 #  docker.io/paritypr/{polkadot-debug,polkadot-parachain-debug}:master-<sha8>
-POLKADOT_SDK_MASTER_SHA=64b05158
+#
+# Pinned to polkadot-sdk#12870 ("network: Enable WebRTC by default on full
+# nodes"), the first master merge whose binaries serve `webrtc-direct`.
+POLKADOT_SDK_MASTER_SHA=04132847
```
skunert: Since this is now merged, better would be to pin it to some commit on either the 2609 branch or master.

skunert (approved):
Makes sense, just some nits.

lrubasze (approved):

e2e-tests/src/network.rs
gab8i: It felt easier to just ask the kernel for a free UDP port and use it instead of going for the back and forth. I think this should reach exactly the same situation as asking the node do so and then fetch the port

PR #3353, opened 2026-08-28
light-base: add `lifecycle_unstable_follow`, a per-chain lifecycle state subscription

### LINKED ISSUE #3301: Smoldot lifecycle event stream in light-base

lrubasze:
Closes #3301. Supersedes #3317 (design discussion: https://github.com/paritytech/smoldot/issues/3301#issuecomment-5410713200).

## What

A small per-chain lifecycle state so embedders can show what the light client is doing without parsing logs:
```json
{
  "phase": { "kind": "syncing", "at": 31699566, "target": 34988240 },
  "hasPeers": true,
  "health": { "kind": "ok" }
}
```

- `phase`: `connecting` | `syncing {at, target}` | `ready` (not terminal: a later warp sync goes back to `syncing`)
- `hasPeers`: at least one peer connected
- `health`: `ok` | `stalled {reason: noPeers | noProgress}` (30s / 45s)

## API

- JSON-RPC: `lifecycle_unstable_follow` / `lifecycle_unstable_unfollow` / `lifecycle_unstable_followEvent`
- First notification is the current state, then one per change; each carries the whole state
- 2 follows per chain (same as `chainHead_v1_follow`); schema is unstable
- Rust: `Client::lifecycle_state(chain_id)`

## Design

- Latest-value broadcast, not an event log: no history, replay, lag handling or event ordering
- Phase is published by the sync service itself (`SyncService::subscribe_sync_status`)
- Progress batched at 500ms (Kusama cold start: ~50 notifications); warp start and `ready` never delayed
- Embedder policy (bootstrap timeouts, RPC fallback) deliberately left out
- Parachains never report `syncing`; subscribe on the relay for a progress bar

## Testing

- Unit tests: state broadcast, sync status emission, JSON shape
- New e2e tests `lifecycle_fresh` and `lifecycle_cold` (cold asserts `syncing` before `ready`), passing locally
- New example: `cargo run -p smoldot-light --features std --example lifecycle`
- Live runs on Polkadot and Kusama with their Asset Hubs


light-base/src/lib.rs
```diff
@@ -1283,11 +1302,145 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+                    return;
+                };
+
+                let now = platform.now();
+                if has_peers {
+                    last_peer_seen = now.clone();
+                }
+
+                let state = lifecycle_service.current().await;
+                let no_progress_for = match (state.phase, &last_progress) {
```
BigTava: The stall clock only runs while `phase` is `Syncing`, so a chain that has peers but never verifies a first warp fragment reports `health: ok` for as long as it stays stuck.

light-base/src/lib.rs
```diff
@@ -1283,11 +1302,145 @@ fn start_services<TPlat: platform::PlatformRef>(
...
+        let platform = platform.clone();
+        async move {
+            const FAST_POLL_WINDOW: Duration = Duration::from_secs(120);
+
+            let started = platform.now();
+            let mut last_peer_seen = started.clone();
+            // Warp sync height last observed, and when it was first observed.
+            let mut last_progress: Option<(u64, TPlat::Instant)> = None;
+
+            loop {
```
BigTava: This loop polls `peers_list` every 1 to 5 seconds for the whole life of every chain even when nobody has subscribed, so it would be worth gating on a subscriber count held by `LifecycleService`.

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -438,6 +442,24 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
                             verified_hash = HashDisplay(&fragment_hash),
                             verified_height = fragment_number
                         );
+
+                        if !task.sync_status_subscribers.is_empty() {
```
BigTava: The subscriber gate is always true.

light-base/src/json_rpc_service/background.rs
```diff
@@ -3115,6 +3249,41 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                subscription_id,
+                state,
+                subscription,
+            }) => {
+                let Some(cancel) = me.lifecycle_subscriptions.get(&subscription_id).cloned() else {
+                    // Unfollowed in the meantime.
+                    continue;
+                };
+
+                let Some(state) = state else {
```
BigTava: NIT: Nothing reaches this arm, because `Background` holds a strong `Arc<LifecycleService>` and unfollow removes the map entry before notifying `cancel`.

e2e-tests/shared/lifecycle.js
```diff
@@ -0,0 +1,187 @@
...
+        withGap.length > 0,
+        `${syncing.length} syncing states, ${withGap.length} with at < target`,
+      );
+    }
+    let monotonic = true;
+    for (let i = 1; i < syncing.length; i++) {
+      if (syncing[i].phase.at < syncing[i - 1].phase.at) monotonic = false;
+    }
+    report(`${label}: syncing height never decreases`, monotonic, `${syncing.length} syncing states`);
+    const readyIndex = states.findIndex((s) => s.phase.kind === "ready");
```
BigTava, line 81: NIT: With no `ready` state `findIndex` returns `-1` and `slice(0)` scans the whole array, so this fails claiming syncing followed a `ready` the run never reached.

lib/src/json_rpc/service/client_main_task.rs
```diff
@@ -474,7 +474,8 @@ impl ClientMainTask {
                 | methods::MethodCall::transactionWatch_v1_submitAndWatch { .. }
                 | methods::MethodCall::sudo_network_unstable_watch { .. }
                 | methods::MethodCall::bitswap_unstable_stream { .. }
-                | methods::MethodCall::chainHead_v1_follow { .. } => {
+                | methods::MethodCall::chainHead_v1_follow { .. }
+                | methods::MethodCall::lifecycle_unstable_follow { .. } => {
```
lexnv, line 478: nit: Do we have a countepart PR in rpc-v2 spec?

light-base/src/json_rpc_service/background.rs
```diff
@@ -527,6 +537,63 @@ enum Event<TPlat: PlatformRef> {
...
+
+/// Builds the task that waits for the next lifecycle state of a `lifecycle_unstable_follow`
+/// subscription, or for its cancellation.
+fn lifecycle_follow_task<TPlat: PlatformRef>(
+    subscription_id: String,
+    mut subscription: lifecycle_service::Subscription,
+    cancel: Arc<event_listener::Event>,
+) -> Pin<Box<dyn Future<Output = Event<TPlat>> + Send>> {
+    Box::pin(async move {
+        let cancelled = cancel.listen();
```
lexnv: We can place this outside of the boxed future to make sure that listener is registered before the future is polled (so a race with `unfollow` cannot happen)

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1118,6 +1140,15 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
                 let _ = send_back.send(out);
             }
 
+            WakeUpReason::ForegroundMessage(ToBackground::SubscribeSyncStatus { send_back }) => {
+                let (tx, rx) = async_channel::unbounded();
+                if task.bootstrap_complete {
+                    let _ = tx.try_send(SyncStatus::Ready);
```
lexnv, line 1117: nit: A subscriber joining after a boostrap complete and re-warp gets `Ready` instead of `WarpSyncing`. We could use last_sent_sync_status otherwise fallback to `SyncStatus::Ready` only when `task.bootstrap_complete == true`

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -438,6 +442,24 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                        if !task.sync_status_subscribers.is_empty() {
+                            let sync = task.sync.as_ref().unwrap_or_else(|| unreachable!());
+                            let peers_best = sync
+                                .sources()
+                                .map(|src| sync.source_best_block(src).0)
+                                .max()
+                                .unwrap_or(0);
+                            emit_sync_status(
+                                &mut task,
+                                SyncStatus::WarpSyncing {
```
lexnv: nit: We are emitting the warp sync status on the first verified fragment. If we have peers connected and we are warping, but the peers are never responding to fragement requests, we postpone the actual status. The easiest to fix this would be emitting a sync status from `CommitWarpAhead`

light-base/src/lifecycle_service.rs
```diff
@@ -0,0 +1,326 @@
...
+    Ok,
+    Stalled { reason: StallReason },
+}
+
+/// Lifecycle state of a chain.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct LifecycleState {
+    pub phase: Phase,
+    /// `true` if at least one peer is currently connected on this chain.
+    pub has_peers: bool,
```
lexnv: nit: Would it be useful to propagate back how many peers we have connected instead as u32?

wasm-node/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+- Add the `lifecycle_unstable_follow` and `lifecycle_unstable_unfollow` JSON-RPC functions. The subscription reports the lifecycle state of the chain: whether it is still connecting, warp syncing (with the current and target block heights), or ready, whether any peer is connected, and whether the built-in watchdog considers it stalled (no peers, or no warp sync progress). The first notification is the current state and every later notification carries the whole new state, so an embedder can show what the light client is doing without parsing log output. The schema is unstable. ([#3301](https://github.com/paritytech/smoldot/issues/3301))
```
lexnv: nit: The first notification currently lands after the first verified fragment

lexnv (approved):

light-base/src/sync_service/substrate_compat.rs
lrubasze: Done. The status is emitted when the first warp fragment request is sent, if the mode is no longer `Deciding`. On bootstrap this is right after `CommitWarpAhead`. It also covers a re-warp after `Ready`, which does not go through `CommitWarpAhead`.

wasm-node/CHANGELOG.md
lrubasze: Fixed, the entry now says syncing is reported as soon as fragments are requested.

light-base/src/lib.rs
lrubasze: Fixed. `Syncing` now starts when the first fragment is requested, so the no-progress clock runs even if no peer answers.

light-base/src/sync_service/substrate_compat.rs
lrubasze: Indeed, that makes much more sense!

light-base/src/sync_service/substrate_compat.rs
lrubasze: Good point! Gate removed

light-base/src/lib.rs
lrubasze: Nice catch! `LifecycleService` counts live subscriptions and the watchdog waits on an event while there are none. On the first subscriber it polls right away and restarts the stall clocks.

light-base/src/lifecycle_service.rs
lrubasze: Good idea. `hasPeers` is now `numPeers` (u32)

lib/src/json_rpc/service/client_main_task.rs
lrubasze: No. The functions are smoldot-specific and marked `_unstable`. If the shape settles and other clients want it, a spec PR can follow

light-base/src/json_rpc_service/background.rs
lrubasze: Right, the service outlives the JSON-RPC service and unfollow removes the entry before notifying. Folded both checks into one guard

light-base/src/json_rpc_service/background.rs
lrubasze: Nice catch! Done

e2e-tests/shared/lifecycle.js
lrubasze: Fixed, the check now requires a ready state to exist

BigTava (approved):
Thanks!

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1367,6 +1384,19 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                // Report the warp sync as soon as fragments are requested rather than once the
+                // first one verifies, so that peers never answering is visible as a stall.
+                // While `Deciding` the mode may still end up `AllForksOnly`; wait for the
+                // decision to avoid a spurious warp-syncing report.
+                if !matches!(task.mode, ModeState::Deciding) {
+                    let local_finalized = task
+                        .sync
+                        .as_ref()
+                        .unwrap_or_else(|| unreachable!())
+                        .finalized_block_number();
```
skunert: `finalized_block_number` will always return the value from all_forks, even if warp sync is already verifying fragments much higher, I think this is a bug.

light-base/src/lib.rs
```diff
@@ -1283,11 +1302,164 @@ fn start_services<TPlat: platform::PlatformRef>(
...
 
+    // The lifecycle service holds a small state describing what the chain is doing, for the
+    // benefit of embedders. Two tasks keep it up to date. Both hold only weak references so
+    // that they stop, rather than keep the chain alive, once the chain is removed.
+    let lifecycle_service = lifecycle_service::LifecycleService::new();
+
+    // Drives `LifecycleState::phase` from the sync service's own status: `Syncing` while a warp
+    // sync is in progress, `Ready` once the sync service serves the chain. Ends when the sync
+    // service is gone.
+    platform.spawn_task("lifecycle-phase".into(), {
```
skunert: IMO this code would be better off in lifecycle_service.rs

light-base/src/sync_service/substrate_compat.rs
lrubasze: Good catch! Added helper `warp_sync_finalized_number`, which gets `finalized_block_number` depending on `AllSync::status`

light-base/src/lib.rs
lrubasze: Done, both tasks moved to `lifecycle_service.rs`

skunert (approved):

PR #3355, opened 2026-09-02
fix: check pre-commit ancestry when verifying justifications

gab8i:
A GRANDPA pre-commit is signed over the block that pre-commit itself targets, never over the block the justification claims to finalize.

Walk `votes_ancestries` from each pre-commit's target down to the justification's target, and reject justifications carrying headers that no pre-commit needed.

### DIFF
--- a/lib/src/finality/verify.rs
+++ b/lib/src/finality/verify.rs
@@ -432,6 +432,33 @@ pub fn verify_justification<'a>(
         return Err(JustificationVerifyError::NotEnoughSignatures);
     }
 
+    // A pre-commit is signed over the block that this specific pre-commit targets, and not over
+    // the block that the justification as a whole claims to finalize. In order for the
+    // justification to actually prove anything, every pre-commit must therefore target either
+    // that block or one of its descendants. The `votes_ancestries` field contains the headers
+    // that link the targets of the pre-commits to the target of the justification.
+    //
+    // The map below contains, for each header found in `votes_ancestries`, the hash of its
+    // parent and whether the entry has been used by at least one of the pre-commits.
+    let num_votes_ancestries = decoded_justification.votes_ancestries.len();
+    let mut votes_ancestries = {
+        let mut map = hashbrown::HashMap::<[u8; 32], ([u8; 32], bool), _>::with_capacity_and_hasher(
+            num_votes_ancestries,
+            crate::util::SipHasherBuild::new({
+                let mut seed = [0; 16];
+                randomness.fill_bytes(&mut seed);
+                seed
+            }),
+        );
+        for header in decoded_justification.votes_ancestries.clone() {
+            map.insert(
+                header.hash(config.block_number_bytes),
+                (*header.parent_hash, false),
+            );
+        }
+        map
+    };
+
     // Verifying all the signatures together brings better performances than verifying them one
     // by one.
     // Note that batched ed25519 verification has some issues. The code below uses a special
@@ -456,7 +483,32 @@ pub fn verify_justification<'a>(
             }
         }
 
-        // TODO: must check signed block ancestry using `votes_ancestries`
+        // Check that the pre-commit targets either the block that the justification claims to
+        // finalize or one of its descendants. Without this check, signatures legitimately
+        // produced for one block could be re-used in a justification that claims to finalize a
+        // completely different block.
+        if precommit.target_hash == decoded_justification.target_hash {
+            if precommit.target_number != decoded_justification.target_number {
+                return Err(JustificationVerifyError::BadAncestry);
+            }
+        } else {
+            let mut current_hash = *precommit.target_hash;
+            let mut num_steps = 0;
+            while current_hash != *decoded_justification.target_hash {
+                let Some((parent_hash, used)) = votes_ancestries.get_mut(&current_hash) else {
+                    return Err(JustificationVerifyError::BadAncestry);
+                };
+                *used = true;
+                current_hash = *parent_hash;
+
+                // A valid chain of ancestors visits each entry of `votes_ancestries` at most
+                // once. Guarantees that the loop always terminates.
+                num_steps += 1;
+                if num_steps > num_votes_ancestries {
+                    return Err(JustificationVerifyError::BadAncestry);
+                }
+            }
+        }
 
         let mut msg = Vec::with_capacity(1 + 32 + 4 + 8 + 8);
         msg.push(1u8); // This `1` indicates which kind of message is being signed.
@@ -489,13 +541,19 @@ pub fn verify_justification<'a>(
         )));
     }
 
+    // Reject justifications that contain headers that no pre-commit has made use of.
+    if votes_ancestries.values().any(|(_, used)| !*used) {
+        return Err(JustificationVerifyError::UnusedAncestryEntry);
+    }
+
     // Actual signatures verification performed here.
     batch
         .verify(&mut randomness)
         .map_err(|_| JustificationVerifyError::BadSignature)?;
 
-    // TODO: must check that votes_ancestries doesn't contain any unused entry
-    // TODO: there's also a "ghost" thing?
+    // Note that the "ghost" of the pre-commits is intentionally not calculated. Doing so would
+    // additionally detect justifications that could have finalized a block higher than the one
+    // that they claim to finalize, which isn't a problem from a safety point of view.
 
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


lib/src/finality/verify.rs
lrubasze: nit: I guess no longer needed

lib/src/finality/verify.rs
```diff
@@ -517,4 +582,354 @@ pub enum JustificationVerifyError {
...
+        assert!(
+            matches!(result, Err(JustificationVerifyError::BadAncestry)),
+            "{result:?}"
+        );
+    }
+
+    /// Same as [`valid_justification_with_ancestries`], but the headers that link the pre-commits
+    /// to the target of the justification are missing.
+    #[test]
+    fn missing_ancestry_headers() {
```
lrubasze, line 980: Worth adding a test where the ancestry chain is only partially provided:
```
    A <- B <- C <- D
                   ^
                 vote(D)

    target = B
    votes_ancestries = [D]        C is missing

    walk:  D -> C      D marked used
           C -> ???    C not in the map  =>  BadAncestry
```
`missing_ancestry_headers` uses an empty list, so the walk fails on the very first lookup and
the loop never iterates. This case fails one step later, after a header was already marked
used, so it covers the loop actually looping and then failing.

lib/src/finality/verify.rs
```diff
@@ -489,13 +548,19 @@ pub fn verify_justification<'a>(
...
+    }
+
     // Actual signatures verification performed here.
     batch
         .verify(&mut randomness)
         .map_err(|_| JustificationVerifyError::BadSignature)?;
 
-    // TODO: must check that votes_ancestries doesn't contain any unused entry
-    // TODO: there's also a "ghost" thing?
+    // Note that the "ghost" of the pre-commits is intentionally not calculated. Doing so would
```
lrubasze, line 572: Worth adding a test that proves this comment:
```
    A <- B <- C
              ^
      vote1 vote2 vote3      all votes on C, none on B

    target = B
    votes_ancestries = [C]    C -> B for every vote, all headers used  =>  Ok
```
A test asserting `Ok(())` here makes it visible if a ghost check is ever added or the comment goes stale.

lib/src/finality/verify.rs
```diff
@@ -517,4 +582,354 @@ pub enum JustificationVerifyError {
...
+            &[],
+        );
+
+        let result = verify(&justification, &authorities);
+        assert!(
+            matches!(result, Err(JustificationVerifyError::NotEnoughSignatures)),
+            "{result:?}"
+        );
+    }
 }
```
lrubasze, line 1408: Worth adding a test where the votes are split across two forks that share the target as
parent. This is the normal shape when a round finalizes a block below the best one:
```
             .- C1
    A <- B
             `- C2

    target = B
    votes  = [vote1(C1), vote2(C1), vote3(C2)]
    votes_ancestries = [C1, C2]

    walk:  vote1:  C1 -> B      C1 used
           vote2:  C1 -> B      C1 used
           vote3:  C2 -> B      C2 used
    used = {C1, C2} == provided  =>  Ok
```
The existing tests all use one straight chain, so nothing checks that the walk works when two ancestry headers have the same parent.

lrubasze (approved):
Nice work here!

lib/src/finality/verify.rs
```diff
@@ -456,7 +488,32 @@ pub fn verify_justification<'a>(
...
+        // produced for one block could be re-used in a justification that claims to finalize a
+        // completely different block.
+        if precommit.target_hash == decoded_justification.target_hash {
+            if precommit.target_number != decoded_justification.target_number {
+                return Err(JustificationVerifyError::BadAncestry);
+            }
+        } else {
+            let mut current_hash = *precommit.target_hash;
+            let mut num_steps = 0;
+            while current_hash != *decoded_justification.target_hash {
```
skunert: This iteration happens before we validate signatures below. So someone could send us a large fabricated header chain and we would waste a lot of time walking it. We could first validate the signatures to make sure that the headers are properly signed. Also we are currently walking the same paths again for every precommit. But we should be able to continue once we enter a known path.  For this I am not sure if its worth it if we do signature verification first.

lib/src/finality/verify.rs
```diff
@@ -517,4 +580,448 @@ pub enum JustificationVerifyError {
...
+        let result = verify(&justification, &authorities);
+        assert!(matches!(result, Ok(())), "{result:?}");
+    }
+
+    /// No pre-commit targets the block that the justification claims to finalize, they all target
+    /// one of its descendants. Because the "ghost" of the pre-commits is intentionally not
+    /// calculated, such a justification is valid even though it could have finalized a higher
+    /// block.
+    #[test]
+    fn precommits_above_target_are_accepted() {
```
skunert, line 822: This is explicitly allowed, but we also need to check that the justification block number is correct. We verify the header path from all precommits to the justification target, but currently do not verify that the block number the justification claims is actually consistent with the headers.

lib/src/finality/verify.rs
```diff
@@ -456,7 +488,32 @@ pub fn verify_justification<'a>(
...
+        // completely different block.
+        if precommit.target_hash == decoded_justification.target_hash {
+            if precommit.target_number != decoded_justification.target_number {
+                return Err(JustificationVerifyError::BadAncestry);
+            }
+        } else {
+            let mut current_hash = *precommit.target_hash;
+            let mut num_steps = 0;
+            while current_hash != *decoded_justification.target_hash {
+                let Some(parent_hash) = votes_ancestries.get(&current_hash) else {
```
skunert: This code here requires that ALL precommits are descendants or equal to the justification target. But actually we only need 2/3+1 right? So this could fail valid justifications.

skunert (approved):
Nice work :)

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,93 @@ pub fn verify_justification<'a>(
...
+                (*header.parent_hash, header.number),
+            );
+        }
+        map
+    };
+
+    // For each header of `votes_ancestries` that a pre-commit has walked through, whether the
+    // path from that header reaches the target of the justification. Pre-commits can
+    // stop as soon as they reach a header that has already been walked through.
+    let mut walked_headers = hashbrown::HashMap::<[u8; 32], bool, _>::with_capacity_and_hasher(
```
lexnv, line 519: I believe we are relaxing two properties present on Substrate side:
- unlinkable precommits are now ignored: https://github.com/paritytech/polkadot-sdk/blob/238533bc0dbf8261b6ad1da2282e5ab239fad82c/substrate/client/consensus/grandpa/src/justification.rs#L249-L262
- unneeded headers are accepted: https://github.com/paritytech/polkadot-sdk/blob/238533bc0dbf8261b6ad1da2282e5ab239fad82c/substrate/client/consensus/grandpa/src/justification.rs#L273-L278

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,93 @@ pub fn verify_justification<'a>(
...
+
+    let mut num_ignored_precommits = 0;
+    for precommit in decoded_justification.precommits.iter() {
+        // Walk from the block the pre-commit targets down to the target of the justification.
+        let mut current_hash = *precommit.target_hash;
+        let mut current_number = precommit.target_number;
+        let reaches_target = loop {
+            if current_hash == *decoded_justification.target_hash {
+                break current_number == decoded_justification.target_number;
+            }
```
lexnv, line 539:
```suggestion
                  if current_number != decoded_justification.target_number {
                      return Err(JustificationVerifyError::BadAncestry);
                  }
                  break;
```

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,93 @@ pub fn verify_justification<'a>(
...
+
+            let Some((parent_hash, number)) = votes_ancestries.get(&current_hash) else {
+                break false;
```
lexnv, line 542:
```suggestion
                return Err(JustificationVerifyError::BadAncestry);
```

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,93 @@ pub fn verify_justification<'a>(
...
+            };
+            if *number != current_number {
+                break false;
```
lexnv, line 545:
```suggestion
                return Err(JustificationVerifyError::BadAncestry);
```

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,93 @@ pub fn verify_justification<'a>(
...
+            if let Some(reaches_target) = walked_headers.get(&current_hash) {
+                break *reaches_target;
+            }
+            current_path.push(current_hash);
+
+            current_hash = *parent_hash;
+            current_number = match current_number.checked_sub(1) {
+                Some(n) => n,
+                None => break false,
+            };
```
lexnv, line 557:
```suggestion
            if !walked_headers.insert(current_hash) {
                break;
            }

            current_hash = parent_hash;
            current_number = current_number
                .checked_sub(1)
                .ok_or(JustificationVerifyError::BadAncestry)?;
```

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,93 @@ pub fn verify_justification<'a>(
...
+        };
+
```
lexnv, line 559:
```suggestion
if walked_headers.len() != votes_ancestries.len() {
        return Err(JustificationVerifyError::UnusedAncestryEntry);
    }
```

lexnv (approved):
LGTM! @gab8i Please have another look at https://github.com/paritytech/smoldot/pull/3355#discussion_r3971641107 before merging 🙏

gab8i:
The latest commit should align smoldot to substrate: the pre-commit with the lowest height is the base, every pre-commit must be linked to the base through `votes_ancestries`, the block that the pre-commits finalize (the "ghost") must be the one the justification claims, and every header in `votes_ancestries` must be needed by one of the links.

lib/src/finality/verify.rs
```diff
@@ -489,13 +485,102 @@ pub fn verify_justification<'a>(
...
+        hashbrown::HashMap::<[u8; 32], usize, _>::with_capacity_and_hasher(
+            votes_ancestries.len(),
+            crate::util::SipHasherBuild::new({
+                let mut seed = [0; 16];
+                randomness.fill_bytes(&mut seed);
+                seed
+            }),
+        );
+
+    for precommit in decoded_justification.precommits.iter() {
```
skunert, line 539: Hmm but in this loop again we walk from each precommit to the lowest one without caching the already traversed path.

skunert (approved):

PR #3359, opened 2026-09-07
verify_compact_trie_proof: borrow partial_key instead of allocating per iteration

### LINKED ISSUE #3275: verify_compact_trie_proof: avoid per-iteration partial_key allocation
Follow-up from #3263 ([discussion](https://github.com/paritytech/smoldot/pull/3263#discussion_r3318064338)).

`compact_proof_compute_merkle` in `lib/src/trie/proof_decode.rs` allocates a `Vec<Nibble>` per descend iteration via `decoded.partial_key.clone().collect()`. Avoiding it requires adding a lifetime to `CompactFrame` so `partial_key` can borrow from `node_bytes`.

Tyagiquamar:
## Summary

- Store \CompactFrame::partial_key\ as \DecodedPartialKey<'a>\ borrowed from proof node bytes instead of cloning into a \Vec<Nibble>\ on every descend iteration in \compact_proof_compute_merkle\.
- Teach \compact_encode_node\ to accept any \ExactSizeIterator + Clone\ partial key, matching \	rie_node::encode_to_vec\.

Fixes #3275.

## Root cause

Each descend step allocated \decoded.partial_key.clone().collect()\ into a new \Vec\, even though the partial key bytes already live inside the proof entry being decoded.

## Tests

\\\
cargo +stable-x86_64-pc-windows-gnu test -p smoldot --lib trie::proof_decode
# 27 passed
\\\

### DIFF
--- a/lib/src/trie/proof_decode.rs
+++ b/lib/src/trie/proof_decode.rs
@@ -1859,8 +1859,8 @@ enum CompactStorageValueOwned {
     None,
 }
 
-struct CompactFrame {
-    partial_key: Vec<nibble::Nibble>,
+struct CompactFrame<'a> {
+    partial_key: trie_node::DecodedPartialKey<'a>,
     new_children: [Option<Vec<u8>>; 16],
     new_storage_value: CompactStorageValueOwned,
     descended_into: usize,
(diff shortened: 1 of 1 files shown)


Tyagiquamar:
Closing: author will re-open without co-author trailer.

PR #3361, opened 2026-09-07
fixes: recover WebRTC connections after a peer restart, related libp2p panics 

gab8i:
Fixes surfaced while enabling browser tests for WebRTC.

- libp2p: no panic when a close or a notification crosses a remote reset.
  `close_out_notifications_substream` and `write_notification_unbounded` unwrapped a map entry the remote's reset had already removed, killing the connection task. Both now return early.
- light-base: stop banning peers for justifications we can't verify yet.
  `TooFarAhead` joins `UnknownTargetBlock` and `JustificationEngineMismatch` as non-bannable: the justification is valid, the local node just hasn't caught up.
- light-base: reset multi-stream connections whose substreams never open.
  A restarted WebRTC peer leaves the browser's `RTCPeerConnection` in `connected` with every data channel dead, so the platform never reports a reset. Each pending outbound substream open now has a 20 s deadline, on expiry the connection is reset. Coordinator messages are pulled on every loop iteration so the shutdown actually reaches the coordinator and the peer is re-dialed.
- e2e-tests: stream JS process output live.
 `Command::output()` buffered until exit, so a hung JS test that got dropped lost all of its logs.


e2e-tests/src/lib.rs
```diff
@@ -198,20 +198,38 @@ pub async fn run_js_test(script: &str, env_vars: &[(&str, &str)]) -> Result<(),
...
+    // its output in the log. Inheriting our stdout/stderr instead would hang the CI step: an
+    // orphaned descendant (Chromium, via Playwright) keeps the job's log pipe open forever.
+    cmd.stdout(std::process::Stdio::piped());
+    cmd.stderr(std::process::Stdio::piped());
+
+    async fn forward(stream: impl tokio::io::AsyncRead + Unpin, tag: &str) {
+        use tokio::io::AsyncBufReadExt as _;
+
+        let mut lines = tokio::io::BufReader::new(stream).lines();
+        while let Ok(Some(line)) = lines.next_line().await {
```
lrubasze: nit: `while let Ok(Some(line))` stops draining on the first non-UTF-8 line (`next_line` returns `InvalidData`). If the child keeps writing after that, the pipe fills, the child blocks, and `wait()` never returns, so the test hangs instead of failing.

Unlikely with node and `console.*`, but the fix is the same size:
```rust
  let mut lines = tokio::io::BufReader::new(stream).split(b'\n');
  while let Ok(Some(line)) = lines.next_segment().await {
      eprintln!("[{tag}] {}", String::from_utf8_lossy(&line));
  }
```

Feel free to ignore

light-base/src/network_service/tasks.rs
```diff
@@ -491,6 +522,24 @@ pub(super) async fn webrtc_multi_stream_connection_task<TPlat: PlatformRef>(
...
+                    "substream-open-timeout",
+                    address = address_string,
+                    num_opening = pending_opening_out_substreams.len()
+                );
+                connection_task.reset();
+                // Pending opens will never complete and open substreams are unusable. Dropping
+                // the latter resets them on the platform side.
+                pending_opening_out_substreams.clear();
+                when_substreams_rw_ready.clear();
+            }
```
lrubasze: nit: Could we just unify those 2 arms? I believe calling `clear()` will not harm in case of `ConnectionReset` branch.

```rust
WakeUpReason::ConnectionReset | WakeUpReason::OutSubstreamOpenTimeout => {
    debug_assert!(!connection_task.is_reset_called());
    if matches!(wake_up_reason, WakeUpReason::OutSubstreamOpenTimeout) {
        log!(
            &platform,
            Debug,
            "connections",
            "substream-open-timeout",
            address = address_string,
            num_opening = pending_opening_out_substreams.len()
        );
    } else {
        log!(&platform, Trace, "connections", "reset", address = address_string);
    }
    connection_task.reset();
    // Pending opens will never complete and open substreams are unusable. Dropping
    // the latter resets them on the platform side.
    pending_opening_out_substreams.clear();
    when_substreams_rw_ready.clear();
}
```

lrubasze (approved):
Nice work here!

lib/src/libp2p/connection/established/multi_stream.rs
```diff
@@ -801,6 +801,9 @@ where
     /// determined by calling [`MultiStream::notification_substream_queued_bytes`]) is below a
     /// certain threshold. If above, the notification should be silently discarded.
     ///
+    /// If the substream has already been reset by the remote, the notification is silently
+    /// discarded.
+    ///
     /// # Panic
     ///
     /// Panics if the [`SubstreamId`] doesn't correspond to a notifications substream, or if the
```
skunert: nit: pls update the panic comment here.

lib/src/libp2p/connection/established/multi_stream.rs
```diff
@@ -859,6 +866,8 @@ where
     /// This can be done even when in the negotiation phase, in other words before the remote has
     /// accepted/refused the substream.
     ///
+    /// If the substream has already been reset by the remote, this method has no effect.
+    ///
     /// # Panic
     ///
     /// Panics if the [`SubstreamId`] doesn't correspond to a notifications substream, or if the
```
skunert: Same here, panic comment still described old panic conditions

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1802,6 +1787,32 @@ fn warp_sync_can_proceed(
     }
 }
 
+/// Returns `true` if a justification verification error means the sender misbehaved.
+///
+/// The following errors are benign and must not lead to a ban:
+///
+/// - `JustificationEngineMismatch`: the chain uses a finality engine smoldot doesn't recognize.
```
skunert: I think these descriptions should be on the enum themselves and not here.

skunert (approved):
looks good, only nits

lexnv (approved):
LGTM!

Issue #3288, opened 2026-06-19
Expose GRANDPA finality justifications over the JSON-RPC interface


bkontur:
## Summary

smoldot verifies GRANDPA justifications internally but exposes none of them over JSON-RPC. We'd like JSON-RPC access to the raw SCALE-encoded GRANDPA justification for finalized blocks — both a **live feed** and an **on-demand fetch** for historical (mandatory) blocks.

The use case is a trustless **bridge relayer running on a light client** (no RPC infrastructure): it submits a finalized header + its GRANDPA justification to a target-chain pallet that re-verifies it on-chain. Everything else already works against smoldot (`state_getReadProof`, `state_call`, `chain_getBlock`, `author_submitExtrinsic`, finalized-head subscriptions) — the justification is the only missing piece. Full context: paritytech/parity-bridges-common#3270.

## Current state (`light-base` v1.2.0; re-verified 2026-07-02 against main `386a1f9a` — unchanged)

No JSON-RPC path to a justification exists:
- `grandpa_subscribeJustifications` — **not implemented**.
- `grandpa_roundState` — TODO stub (`lib/src/json_rpc/methods.rs:426`), answered with `-32000 "Not implemented in smoldot yet"`.
- `chain_getBlock` — **always returns `justifications: None`** (`light-base/src/json_rpc_service/background.rs:5824-5827`, "There's no way to verify the correctness of the justifications, consequently we always return an empty list").

The data, however, is already there and fetchable:
- Decoded & verified in `lib/src/finality/verify.rs` (`verify_justification`); the verified bytes are then dropped when the sync service emits its `Notification::Finalized` (`light-base/src/sync_service/substrate_compat.rs`).
- Carried by two protocols smoldot already speaks — the **GRANDPA warp-sync proof** (each fragment = header + justification of an authority-set change, `lib/src/network/codec/grandpa_warp_sync.rs`) and **block requests with justifications** (`lib/src/network/codec/block_request.rs`).
- **light-base already issues justification-bearing block requests itself**: the sync service sets `justifications` in `BlocksRequestFields` when the state machine asks for it (`substrate_compat.rs`), and the public `sync_service::block_query()` / `block_query_unknown_number()` API accepts a `justifications` flag — every JSON-RPC caller merely passes `false` today. Warp sync is also re-armable after startup (`ModeState::AwaitingWarp`), not startup-only.
[...]

Issue #3270, opened 2026-05-26
Support child tries in `chainHead_v1_storage`


skunert:
Currently a non-null value in child_trie is not supported: https://github.com/paritytech/smoldot/blob/bea0e887912d0d91b6144f087d888c0b739c55ff/light-base/src/json_rpc_service/background.rs#L2270-L2278

At the substrate light-client protocol, child reads [should be supported](https://github.com/paritytech/polkadot-sdk/blob/ef3f7f4935c5e6e39c357fe10530069c2ce41ff2/substrate/client/network/light/src/light_client_requests/handler.rs#L159) already.

We should bring support for child_trie queries too. First step is a proof of concept to see if there was good reason (bug, insufficient API, etc.) for it to be skipped initially or just not yet done.

skunert:
Solved by #3278

PR #3363, opened 2026-09-07
move chain_getBlock justifications out of the block object to fix relayer decode issue

rosarp:
This is prerequisite for paritytech/parity-bridges-common#3270 (light-client bridge relayers),
and for paritytech/smoldot#3288 (exposing GRANDPA justifications over JSON-RPC).

## Why

The relayer's job is to submit a finalized source-chain header **plus its SCALE-encoded GRANDPA justification** to a bridge pallet on the target chain, which re-verifies GRANDPA on-chain.
When the Polkadot ↔ Kusama `substrate-relay` is ran against an embedded smoldot light client, it is found that the smoldot response does not decode.

`chain_getBlock` is one of the two paths a justification would have to travel over (the other being a `grandpa_subscribeJustifications` feed), and its response shape is wrong in a way that makes it un-decodable for the client the relayer uses.

### What is required in smoldot

`chain_getBlock`'s response type is `sp_runtime::generic::SignedBlock`:

```rust
#[serde(deny_unknown_fields)]
pub struct SignedBlock<Block> {
    pub block: Block,                          // itself deny_unknown_fields
    pub justifications: Option<Justifications>,
}
```

justifications is a sibling of block, not one of its members, and both SignedBlock and the inner Block carry deny_unknown_fields. smoldot serialized it nested inside block, which fails to decode twice over: the inner block rejects the
unknown justifications key, and the outer struct never sees the field it expects.

Because `deny_unknown_fields` rejects the key, not a value, this broke even when there was nothing to report — and smoldot always reported null here, so the field was never populated and the misplacement went unnoticed. Any Substrate-based JSON-RPC client (subxt, substrate-relay, anything decoding into sp_runtime types) got a deserialization error rather than an empty justification list.

So before a justification can be delivered over chain_getBlock at all, the envelope has to be the shape Substrate clients expect.

How this PR solves it:

- serde::Serialize for methods::Block now emits justifications as a sibling of   block, matching SignedBlock. block contains only header and extrinsics.
- The Block struct itself, and every other API, are unchanged — this is purely the   wire encoding, so there is no impact on callers inside smoldot.
- A comment at the serializer records why the placement matters, so it doesn't get   "tidied" back into the block object later.

The justification encoding is unchanged and already matched `sp_runtime::Justifications` — a list of (`ConsensusEngineId, Vec<u8>`) pairs, each serialized as a two-element array of byte arrays.

Result

`chain_getBlock` responses now decode into `sp_runtime::generic::SignedBlock`, this lets the light-client relayer of #3270 read a block , and what makes the justifications field a usable delivery path once #3288 starts populating it.

### DIFF
--- a/lib/src/json_rpc/methods.rs
+++ b/lib/src/json_rpc/methods.rs
@@ -1403,11 +1403,7 @@ impl serde::Serialize for Block {
     where
         S: serde::Serializer,
     {
-        // The shape below must match `sp_runtime::generic::SignedBlock`, whose `justifications`
-        // field is a *sibling* of `block` and not one of its members. Both `SignedBlock` and the
-        // inner `Block` are annotated with `deny_unknown_fields`, so putting `justifications` in
-        // the wrong object makes the response impossible to decode for Substrate-based clients,
-        // even when there is no justification to report.
+        // The shape below must match `sp_runtime::generic::SignedBlock`.
         #[derive(serde::Serialize)]
         struct SerdeBlock<'a> {
             block: SerdeBlockInner<'a>,
--- a/lib/src/json_rpc/methods.rs
+++ b/lib/src/json_rpc/methods.rs
@@ -1403,28 +1403,33 @@ impl serde::Serialize for Block {
     where
         S: serde::Serializer,
     {
+        // The shape below must match `sp_runtime::generic::SignedBlock`, whose `justifications`
+        // field is a *sibling* of `block` and not one of its members. Both `SignedBlock` and the
+        // inner `Block` are annotated with `deny_unknown_fields`, so putting `justifications` in
+        // the wrong object makes the response impossible to decode for Substrate-based clients,
+        // even when there is no justification to report.
         #[derive(serde::Serialize)]
         struct SerdeBlock<'a> {
             block: SerdeBlockInner<'a>,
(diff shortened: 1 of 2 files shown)


lib/src/json_rpc/methods.rs
```diff
@@ -1403,28 +1403,33 @@ impl serde::Serialize for Block {
     where
         S: serde::Serializer,
     {
+        // The shape below must match `sp_runtime::generic::SignedBlock`, whose `justifications`
+        // field is a *sibling* of `block` and not one of its members. Both `SignedBlock` and the
+        // inner `Block` are annotated with `deny_unknown_fields`, so putting `justifications` in
+        // the wrong object makes the response impossible to decode for Substrate-based clients,
+        // even when there is no justification to report.
```
bkchr:
```suggestion
        // The shape below must match `sp_runtime::generic::SignedBlock`.
```

bkchr (approved):

skunert (approved):

lrubasze (approved):
@rosarp please push with all commits signed - otherwise it cannot be merged.

bkontur (approved):

rosarp:
> @rosarp please push with all commits signed - otherwise it cannot be merged.

Thanks for letting me know @lrubasze

PR #3362, opened 2026-09-07
tools: add bootnode-inspector

lrubasze:
Standalone script in `tools/bootnode-inspector`, depending only on the latest `smoldot` npm package.

- Checks every bootnode of a chain spec on its own: a fresh light client per address, with that address as the only bootnode (possible to use bootnode provided by hand)
- Reports whether the libp2p handshake completed and how long until `chainHead_v1_follow` reported a finalized block.
- A plain DNS and TCP probe runs alongside, so failures are labelled `dns: ENOTFOUND`, `tcp: ECONNREFUSED`, `tcp: connect timeout`, `peer id mismatch`, or `handshake not completed`.

Script arguments:
- `--handshake-only` stops at the handshake,
- `--discover <s>` lists the peers found through the bootnode and marks the ones smoldot connected to,
- `--bootnode <addr>` checks addresses given by hand,
- `--json` - output in JSON format
- `--verbose` for output.

Additionally:
- updated Paseo relay and parachain specs.

PR #3364, opened 2026-09-08
wasm-node: report TCP socket errors as connection resets on NodeJS

lrubasze:
On NodeJS, TCP connection failures and remote closes were never reported to smoldot; a refused port or unresolvable name was only noticed by the 4 second handshake timeout, with no reason logged. WebSocket connections were not affected.

Cause:
- The TCP `close` handler returned early on `socket.destroyed`, but NodeJS marks a socket destroyed before emitting `close`, so it returned on every close.
- The `error` handler was empty, so the error text was lost.

Fix:
- Track `reset()` with an explicit flag instead of `socket.destroyed`.
- Record the `error` message and report it from `close` via `onConnectionReset`.
- Same behaviour as the WebSocket and WebRTC transports; a flag instead of detaching listeners, since a NodeJS socket with no `error` listener throws.

Effect:
- `connections reset` now carries e.g. `connect ECONNREFUSED 127.0.0.1:30333` or `getaddrinfo ENOTFOUND host` within milliseconds.
- The peer is banned and retried right away instead of holding a slot for 4 seconds.

Test:
- `test/tcp-reset-reason.mjs`: refused local port and unresolvable hostname, reset expected in under 3 seconds. Fails on 3.4.1, passes here. Full wasm-node JS suite passes.

Found while building the bootnode inspector (#3362).


wasm-node/javascript/src/no-auto-bytecode-nodejs.ts
```diff
@@ -159,35 +159,42 @@ function connect(config: ConnectionConfig): Connection {
...
+
         socket.setNoDelay();
 
         socket.on('connect', () => {
-            if (socket.destroyed) return;
+            if (resetCalled) return;
             config.onWritableBytes(socket.writableHighWaterMark);
         });
+        socket.on('error', (error) => {
+            lastError = error.message;
```
skunert: The docs say that this can emit and AggregateError if all attempts fail: https://nodejs.org/api/net.html#socketconnectoptions-connectlistener

So should we respect the extra fields there?
```
socket.on("error", (error: NodeJS.ErrnoException & { errors?: Error[] }) => {
  lastError =
    error.errors?.map(cause => cause.message).filter(Boolean).join("; ")
    || error.message
    || error.code
    || "Error";
});
```

skunert (approved):

lexnv (approved):
LGTM as a nonJs expert 🚀

wasm-node/javascript/src/no-auto-bytecode-nodejs.ts
lrubasze: Good catch! Confirmed on Node 20: `localhost` on a closed port gives an `AggregateError` with an empty `message`, so the reason was empty. Same on the WebSocket path via `ws`.

Applied as a shared `errorReason()` helper for both handlers, plus tests dialing `/dns/localhost/tcp/<closed port>` with and without `/ws`.

PR #3366, opened 2026-09-08
light-base: only forward reported pruned blocks to runtime service subscribers

### LINKED ISSUE #3367: light-base: panic in Subscription::unpin_block - runtime service forwards unannounced pruned blocks
```
panicked at light-base/src/runtime_service.rs:493:34
called `Result::unwrap()` on an `Err` value: ()
```

Still seen on smoldot 3.4.1 (Sentry DOTLI-BM, 5 users in 2 weeks) after #3287.

`Notification::Finalized { pruned_blocks }` lists every block pruned from the runtime service tree, including blocks pruned before they were announced with `Notification::Block`. Subscribers unpin what they are told about. The legacy JSON-RPC subscription (created at startup) thus unpins a block it never held and gets `Err(())`, which `Subscription::unpin_block` unwraps.

Unannounced pruned blocks arise when a block is inserted right before a Finalized notification (sync-service notifications are polled before tree advances, and `AsyncTree::try_advance_output` finalizes before emitting pending blocks), or when the block's runtime download is still running when it gets pruned.

The same hashes also leak into `chainHead_v1_follow` `prunedBlockHashes`.

Attribution to DOTLI-BM is by elimination: Sentry events carry no task name, `Err(())` has a single producer in the runtime service, and of the five `Subscription::unpin_block` callers this is the only one fed a list that can contain never-pinned hashes.

Related: #3286 fixed the paraheads caller of the same function.

lrubasze:
Fixes #3367
Fixes the remaining `runtime_service.rs:493` unpin panic (Sentry DOTLI-BM), still seen on 3.4.1 after #3287. 
Also fixes a pinned-blocks budget leak found in review.

### Problem

The runtime service announces blocks to subscribers with `Notification::Block` and pins them. Later it sends `Notification::Finalized { pruned_blocks }` and subscribers unpin those.

`pruned_blocks` included blocks that were pruned before they were ever announced. The legacy JSON-RPC subscription then unpinned a block it never held, got `Err(())`, and panicked. The same hashes also leaked into `chainHead_v1_follow` `prunedBlockHashes`.

Separately, on every finalization subscribers were charged for finalized and pruned blocks they had already unpinned. The slot was never refunded, so early unpins eventually stopped the subscription.

### Fix

- `async_tree`: `pruned_blocks` is now `Vec<PrunedBlock>` with `async_op_user_data` (operation finished) and `reported` (block was emitted). Finished does not imply announced, so both are needed.
- `runtime_service`: forward only pruned blocks with `reported == true`.
- `runtime_service`: charge the pinned-blocks budget only for blocks the subscriber still has pinned.

### Tests

- Existing async_tree test also asserts `reported == false` for the finished-but-unannounced block.
- New test covers announced, finished-but-unannounced, and in-progress pruned blocks.
- New e2e test `chainhead_v1_unpin_early`: unpins every block on `newBlock`, expects 40 `finalized` events without `stop`.


light-base/src/runtime_service.rs
```diff
@@ -1066,17 +1069,26 @@ async fn run_background<TPlat: PlatformRef>(
...
-                    pruned_blocks: pruned_blocks.iter().map(|(_, b, _)| b.hash).collect(),
+                    pruned_blocks: reported_pruned_blocks.clone(),
                 };
 
                 let mut to_remove = Vec::new();
                 for (subscription_id, (sender, finalized_pinned_remaining)) in
                     all_blocks_subscriptions.iter_mut()
                 {
-                    let count_limit = pruned_blocks.len() + 1;
+                    let count_limit = reported_pruned_blocks.len() + 1;
```
lexnv: Could we determine the limit by the number of pins that are still present for the client?

We now charge each subscriber a flat count_limit, but `block_ignores_limit` is flipped only on pins that still exist. Clients might unpin the block while still not yet finalized to clean up resources. But when that happens, we dont give back the slot. 

Basically each unpin block that is unfinalized is leaking a slot. If we do that 32 times, we'd terminate the subscription with a `Stop` event

lexnv (approved):
LGTM! Nice one! 

https://github.com/paritytech/smoldot/pull/3366#discussion_r3971763355 could also be a followup but lets create an issue for it if we deffer so we don;t lose track 🙏

skunert (approved):

light-base/src/runtime_service.rs
lrubasze: Great catch, thanks @lexnv!

Fixing it here. Subscribers are now charged only for the finalized and pruned blocks they still have pinned. A block unpinned early costs nothing.
Also added an e2e test `chainhead_v1_unpin_early` that unpins every block right after `newBlock` and expects 40 finalized events without a stop.

PR #3368, opened 2026-09-09
npm smoldot v3.5.0

### LINKED ISSUE #3349: statement-store: statement_* answers diverge from a full node's

lrubasze:
This is to:
- publish `smoldot-v3.5.0` npm.
- publish crates `smoldot-light v2.0.2`, `smoldot v3.0.0`

### Changed

- `statement_submit` validates statements like a full node (expiry, size, proof) and answers `invalid` with `alreadyExpired`, `encodingTooLarge` or `noProof`. Undecodable payloads and statements no peer accepted are now JSON-RPC error `7001`. ([#3344](https://github.com/paritytech/smoldot/pull/3344); fixes [#3349](https://github.com/paritytech/smoldot/issues/3349))

### Fixed

- `chain_getBlock` returns `justifications` as a sibling of `block`, matching `SignedBlock`, so Substrate clients can decode the response. ([#3363](https://github.com/paritytech/smoldot/pull/3363); related to [#3288](https://github.com/paritytech/smoldot/issues/3288))
- Recover WebRTC connections after a peer restart: substream opens time out after 20 s and the connection is reset and re-dialed. ([#3361](https://github.com/paritytech/smoldot/pull/3361))
- Fix a connection task panic when a substream close or notification crosses a remote reset. ([#3361](https://github.com/paritytech/smoldot/pull/3361))
- No longer ban peers for justifications that target a block too far ahead. ([#3361](https://github.com/paritytech/smoldot/pull/3361))
- Bump `arkworks` to 0.6; a malformed elliptic curve host function input no longer aborts the process. ([#3356](https://github.com/paritytech/smoldot/pull/3356))

### DIFF
--- a/full-node/Cargo.toml
+++ b/full-node/Cargo.toml
@@ -37,6 +37,6 @@ serde_json = { version = "1.0.104", default-features = false, features = ["std"]
 siphasher = { version = "1.0.1", default-features = false }
 soketto = { version = "0.8.0", features = ["deflate"] }
 smol = "2.0.0"
-smoldot = { version = "2.0.0", path = "../lib", default-features = false, features = ["database-sqlite", "std", "wasmtime"] }
+smoldot = { version = "3.0.0", path = "../lib", default-features = false, features = ["database-sqlite", "std", "wasmtime"] }
 terminal_size = "0.4.0"
 zeroize = { version = "1.7.0", default-features = false, features = ["alloc"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2675,7 +2675,7 @@ dependencies = [
 
 [[package]]
 name = "smoldot"
-version = "2.2.0"
+version = "3.0.0"
 dependencies = [
  "ark-bls12-381",
  "ark-ec",
@@ -2773,7 +2773,7 @@ dependencies = [
 
 [[package]]
 name = "smoldot-light"
-version = "2.0.1"
+version = "2.0.2"
 dependencies = [
  "async-channel",
  "async-lock",
@@ -2808,7 +2808,7 @@ dependencies = [
 
 [[package]]
 name = "smoldot-light-wasm"
-version = "3.4.1"
+version = "3.5.0"
 dependencies = [
  "async-lock",
  "async-task",
--- a/benchmarks/Cargo.lock
+++ b/benchmarks/Cargo.lock
@@ -7498,7 +7498,7 @@ dependencies = [
 
 [[package]]
 name = "smoldot"
-version = "2.2.0"
+version = "3.0.0"
 dependencies = [
  "ark-bls12-381 0.6.0",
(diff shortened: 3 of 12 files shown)


bkchr (approved):

full-node/Cargo.toml
```diff
@@ -37,6 +37,6 @@ serde_json = { version = "1.0.104", default-features = false, features = ["std"]
 siphasher = { version = "1.0.1", default-features = false }
 soketto = { version = "0.8.0", features = ["deflate"] }
 smol = "2.0.0"
-smoldot = { version = "2.0.0", path = "../lib", default-features = false, features = ["database-sqlite", "std", "wasmtime"] }
```
skunert, line 40: Why the major bump? Because of the RPC change?

lrubasze: yes, this part to be precise:
https://github.com/paritytech/smoldot/pull/3344/changes#diff-9b9afc030c0604c41af2f45f88c8d2aeead76ff0b0bb7fbb17226ca4d6292a59R1127-R1130

- `StatementSubmitResult` is reachable as `smoldot::json_rpc::methods::StatementSubmitResult`. Both `json_rpc` and methods are `pub mod`, so it is part of the crate's public API.
- The `InternalError` variant and the `InternalError` enum were removed. Any downstream match on this enum, or any code naming `InternalError`, stops compiling.
- `Invalid` changed from a struct variant with a reason field to a tuple variant. Code writing `Invalid { reason }` also stops compiling.

skunert (approved):

PR #3369, opened 2026-09-09
Expose GRANDPA finality justifications over JSON-RPC

### LINKED ISSUE #3288: Expose GRANDPA finality justifications over the JSON-RPC interface

rosarp:
Closes paritytech/smoldot#3288.
Unblocks paritytech/parity-bridges-common#3270 (light-client bridge relayers).

## Why

A bridge relayer submits a finalized source-chain header **plus its SCALE-encoded GRANDPA justification** to a bridge pallet on the target chain, which re-verifies GRANDPA on-chain.

smoldot verifies these proofs internally and then discards them, and exposes no JSON-RPC method that returns one: `grandpa_subscribeJustifications` was not implemented and `chain_getBlock` always returned `justifications: null`. A relayer running against a light client therefore has nothing to submit, so the light-client transport cannot be used.

## How

- Adds `grandpa_subscribeJustifications` / `grandpa_unsubscribeJustifications`, a live feed of every finality proof the light client verifies.
- `chain_getBlock` now returns the block's justification instead of always `null`: either the proof verified locally while finalizing that block, or the one a peer serves alongside the block, verified locally before it is returned.
- In steady state finality arrives as a gossiped GRANDPA *commit* rather than a justification, so verified commits are re-encoded into equivalent justifications.
- Proofs for authority-set-change blocks, including those from warp sync, are retained longer than ordinary ones, since a relayer must submit them in order and cannot skip them.

Both methods match the Substrate legacy API, so a relayer needs no smoldot-specific code path.

### DIFF
--- a/lib/src/finality/encode.rs
+++ b/lib/src/finality/encode.rs
@@ -1,0 +1,514 @@
+// Smoldot
+// Copyright (C) 2019-2022  Parity Technologies (UK) Ltd.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Encoding of Grandpa finality proofs.
+//!
+//! Commits and justifications carry the same information - a round number, a target block and
+//! signed pre-commits - and differ only in their encoding and in how they are transmitted:
+//! commits are gossiped, justifications are stored by full nodes and served over block requests
+//! and warp sync. Consumers that verify Grandpa on their own expect the justification encoding,
+//! which this module produces from a gossiped commit.
+
+use crate::{finality::decode, header, util};
+
+use alloc::{collections::BTreeSet, vec::Vec};
+
+/// Turns a SCALE-encoded Grandpa commit into a SCALE-encoded Grandpa justification targeting the
+/// same block.
+///
+/// The commit **must** have been verified beforehand (see
+/// [`crate::finality::verify::verify_commit`]), as this function performs no verification
+/// of its own beyond decoding.
+///
+/// # Vote ancestries
+///
+/// A commit assumes that its receiver already knows the blocks that the pre-commits target, while
+/// a justification must be verifiable in isolation and therefore also carries the headers linking
+/// each pre-commit target back to the finalized block: the `votes_ancestries` field.
+/// `scale_encoded_header_by_hash` provides those headers, returning `None` for a block the caller
+/// doesn't know about.
+///
+/// Pre-commits that can't be linked back are **left out**: a verifier gives no voting weight to a
+/// pre-commit whose ancestry it can't check, so dropping one doesn't change whether the
+/// justification is accepted, and keeps it smaller. Pre-commits targeting the finalized block
+/// itself, which is the vast majority of them, need no ancestry at all.
+pub fn grandpa_commit_to_justification(
+    scale_encoded_commit: &[u8],
+    block_number_bytes: usize,
+    mut scale_encoded_header_by_hash: impl FnMut(&[u8; 32]) -> Option<Vec<u8>>,
+) -> Result<Vec<u8>, CommitToJustificationError> {
+    let commit = decode::decode_grandpa_commit(scale_encoded_commit, block_number_bytes)
+        .map_err(|_| CommitToJustificationError::CommitDecode)?;
+
+    // The commit stores the signatures and public keys of the pre-commits separately from the
+    // pre-commits themselves, while the justification stores them inline. The two lists
+    // consequently must have the same length.
+    if commit.precommits.len() != commit.auth_data.len() {
+        return Err(CommitToJustificationError::PrecommitsAuthDataMismatch);
+    }
+
+    // SCALE-encoded headers of the `votes_ancestries` field, and the hashes of the blocks that
+    // they correspond to. A block is added to this list at most once, and only if it is on the
+    // path between a pre-commit target and the target of the commit. Verifiers reject
+    // justifications that carry a header that no pre-commit needs.
+    let mut votes_ancestries = Vec::new();
+    let mut votes_ancestries_hashes = BTreeSet::new();
+
+    // SCALE-encoded pre-commits that are kept, and how many of them there are. Built separately
+    // from `out`, as the number of pre-commits isn't known until the ancestry of all of them has
+    // been resolved.
+    let mut precommits =
+        Vec::with_capacity(commit.precommits.len() * (32 + block_number_bytes + 64 + 32));
+    let mut num_precommits = 0;
+
+    for (precommit, (signature, authority_public_key)) in
+        commit.precommits.iter().zip(commit.auth_data.iter())
+    {
+        match ancestry_of(
+            precommit.target_hash,
+            precommit.target_number,
+            commit.target_hash,
+            commit.target_number,
+            block_number_bytes,
+            &votes_ancestries_hashes,
+            &mut scale_encoded_header_by_hash,
+        ) {
+            Some(headers) => {
+                for (hash, scale_encoded_header) in headers {
+                    votes_ancestries_hashes.insert(hash);
+                    votes_ancestries.push(scale_encoded_header);
+                }
+            }
+            // The pre-commit targets a block that isn't known to be a descendant of the block
+            // that the commit finalizes. See the documentation of this function.
+            None => continue,
+        }
+
+        precommits.extend_from_slice(precommit.target_hash);
+        push_block_number(&mut precommits, precommit.target_number, block_number_bytes)?;
+        precommits.extend_from_slice(&signature[..]);
+        precommits.extend_from_slice(&authority_public_key[..]);
+        num_precommits += 1;
+    }
+
+    let mut out = Vec::with_capacity(
+        8 + 32
+            + block_number_bytes
+            + 5
+            + precommits.len()
+            + 5
+            + votes_ancestries.iter().map(|h| h.len()).sum::<usize>(),
+    );
+
+    // `round: u64`
+    out.extend_from_slice(&commit.round_number.to_le_bytes());
+    // `commit.target_hash: [u8; 32]`
+    out.extend_from_slice(commit.target_hash);
+    // `commit.target_number`, encoded on `block_number_bytes` bytes, little endian.
+    push_block_number(&mut out, commit.target_number, block_number_bytes)?;
+
+    // `commit.precommits: Vec<SignedPrecommit>`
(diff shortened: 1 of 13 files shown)


light-base/src/sync_service/substrate_compat.rs
```diff
@@ -596,6 +714,43 @@ pub(super) async fn start_substrate_compatible_chain<TPlat: PlatformRef>(
...
+                                    &target.header,
+                                    task.block_number_bytes,
+                                ),
+                            );
+                        }
+
+                        // Tracked so that a justification fetched from the network later can be
+                        // verified against the set that signed it. A set change can be anywhere
+                        // in the batch, not only at the proof's target.
+                        let enacted_at = finalized_blocks_newest_to_oldest
```
lrubasze, line 742: I believe `enacted_at` is not needed, we could always use `finalized_number` (as in line 1998, when `enacted_at` is `None`). A proof can't finalize past a pending set-change block, so when the set id changes, the batch target is that block and `finalized_number` at line 1998 is already the boundary.

light-base/src/sync_service.rs
```diff
@@ -274,6 +275,161 @@ impl<TPlat: PlatformRef> SyncService<TPlat> {
...
+
+        let proof = FinalityProof {
+            consensus_engine_id: *b"FRNK",
+            scale_encoded_justification,
+            target_hash: *block_hash,
+            target_number,
+        };
+
+        self.to_background
+            .send(ToBackground::RetainFinalityProof {
```
lrubasze, line 408: This stores every fetched justification in `mandatory_finality_proofs`, whether or not the block changes the authority set. According to its doc comment that cache should only hold set-change ones, so this needs a filter, eg.`header_changes_authority_set` like `on_verified_finality_proof` does.

light-base/src/sync_service.rs
```diff
@@ -274,6 +275,161 @@ impl<TPlat: PlatformRef> SyncService<TPlat> {
...
+    /// happened to be commit targets. A consumer that must prove a *specific* block final to a
+    /// third party has to get that proof elsewhere; full nodes keep the justifications of
+    /// authority-set-change blocks permanently and serve them alongside the block itself, at no
+    /// extra round trip.
+    ///
+    /// `block` is unverified data from a peer, so this checks that the header hashes to
+    /// `block_hash`, that the justification targets that same block, and that it is signed by the
+    /// authority set responsible for finalizing it. Returns `None` when that set is no longer
+    /// remembered: being unable to check a proof is treated as not having one.
+    pub async fn verified_finality_proof_from_block(
```
lrubasze, line 344: nice-to-have: 
Some tests would be useful here. Cases:
- invalid returning `None`: 
  - justification for a different block, 
  - wrong set id, 
  - header hash mismatch
- valid one that is returned and then found by `finality_proof`.

light-base/src/sync_service/substrate_compat.rs
```diff
@@ -1706,6 +1938,99 @@ enum RequestOutcome {
...
+        }
+    }
+
+    /// Records the composition of the current Grandpa authority set, closing the previous entry
+    /// if the set has changed.
+    ///
+    /// `enacted_at` is the height of the highest just-finalized block that enacts a change of
+    /// authorities, if any. That block is finalized by the outgoing set, so it is recorded as
+    /// that set's last one.
+    fn refresh_grandpa_authority_sets<TRq, TSrc, TBl>(
```
lrubasze, line 1972: A test for the set change would help: after a change at height N, `grandpa_authority_set_for(N)` gives the old set and `grandpa_authority_set_for(N + 1)` the new one. Taking the chain information instead of `&AllSync` would allow building the input by hand, without a signed set change.

lib/src/finality/encode.rs
```diff
@@ -0,0 +1,514 @@
...
+    let bytes = value.to_le_bytes();
+
+    // Any byte that we are about to truncate away must be zero, otherwise the encoding would
+    // silently alter the block number.
+    if bytes
+        .iter()
+        .skip(core::cmp::min(8, block_number_bytes))
+        .any(|b| *b != 0)
+    {
+        return Err(CommitToJustificationError::BlockNumberTooLarge);
```
lrubasze, line 204: Unreachable: the target numbers were decoded from exactly `block_number_bytes` bytes, so they always fit. Suggest removing the variant and the check.

lib/src/finality/encode.rs
```diff
@@ -0,0 +1,514 @@
...
+        assert_eq!(&justification[8..40], &[0xaa; 32]);
+        assert_eq!(&justification[40..48], &1234u64.to_le_bytes());
+        // Compact-encoded length of the list of pre-commits.
+        assert_eq!(justification[48], 3 << 2);
+        // Compact-encoded length of the (empty) list of vote ancestries.
+        assert_eq!(*justification.last().unwrap(), 0);
+    }
+
+    #[test]
+    fn commit_to_justification_block_number_too_large() {
```
lrubasze, line 500: This tests the decoder rejecting a 4-byte layout read as 2-byte fields, not a too-large number. Suggest renaming.

lrubasze (commented):
Nice work done here 👍  Here are my remarks

light-base/src/json_rpc_service/background.rs
```diff
@@ -1541,6 +1571,61 @@ pub(super) async fn run<TPlat: PlatformRef>(
...
+                            )
+                            .await;
+
+                        me.grandpa_justifications_subscriptions
+                            .insert(subscription_id);
+
+                        // The sync service builds proofs whether or not anybody is listening,
+                        // so this only decides when forwarding starts. Proofs from before the
+                        // first subscription stay reachable through `chain_getBlock`.
+                        if !me.grandpa_justifications_stream_active {
```
skunert: Once the last justification stream unsubscribes, we should let this background task future finish.

skunert:
I took an extended look into this and thought about the implications and alternatives. There are some problems I see with the current approach:

- Every gossiped commit is re-encoded into a justification and cached, even with no consumer and before the proof is known to be valid. 
- We cache the justifications in smoldot, even though this new RPC is probably niche. Also it consumes extra memory which I would not like to have for everyone.
- New unspecced RPC method for this use-case

---
### Alternative design

From a quick look at this relayer, I think we could get away with a smaller change:

- We own the relayer and its native,  so we embed a native smoldot and could expose some simple rust-API instead of introducing a new json-rpc
- This relayer reads a starting point from the runtime of the target chain and uses this as a starting point. Everything we need are some follow up justifictions of the mandatory blocks, so the ones with authority changes
- This requirement is very similar to the justification fetching that warp sync does
- Concrete proposal is to offer a small way in the rust API of smoldot to send a warp proof request with a beginning that is what the target chain knows. The peers in the network will automatically deliver us justifications for the next mandatory blocks. We deliver the raw bytes via this API, the relayer already has some verification code AFAIK.
- With this sketched approach we don't take browser users hostage and don't require an in-node cache or extra business logic.

@rosarp @bkontur Does this work for you?

PR #3373, opened 2026-09-17
executor: fix duplicate and missing TrieNodeRemoveEvent in trie_root_calculator

lrubasze:
`trie_root_calculator` reports to the caller which trie nodes were inserted, updated or removed. Two defects in the removal reports are fixed. The root hash was always correct.

### Fixes
- Duplicate remove: after a node is replaced by its single remaining child, the child's subtree is walked again and its removals were reported a second time. Now suppressed during the re-walk.
- Missing remove: when the diff empties a trie whose root was a branch node without a value (for example a child trie deletion), the destroyed root was never reported. Now it is.

The issue was causing random CI failures of the `trie_root_calculator::tests::fuzzing` test:
- https://github.com/paritytech/smoldot/actions/runs/34229271036/job/102071029890
- https://github.com/paritytech/smoldot/actions/runs/34450305104/job/102784286696

### Tests
- modified fuzz test to print the seed on failure for easier reproduction in the future
- added regression tests for both defects
- added exhaustive enumeration test with every possible set of changes to detect other potential issues. Instead of random cases it tries every possible small trie (alphabet of seven short keys) with every possible combination of operation, about 560k cases.


skunert (approved):

lib/src/executor/trie_root_calculator.rs
```diff
@@ -445,6 +463,8 @@ impl ClosestDescendantMerkleValue {
         // bug somewhere in the API user's code.
         debug_assert!(merkle_value.len() == 32 || trie::trie_node::decode(merkle_value).is_ok());
 
+        self.inner.maybe_finish_rewalk();
```
bkchr: Why are you sprinkling this line of code "everywhere"? Shouldn't it be enough to put this where we push onto the stack?

lrubasze: Good point. Moved the check to the top of `Inner::next()`, so it is done once.

Background: the push of a node's result to its parent is itself "sprinkled" in four places, and all four call `next()` right after. The alternative is a small helper that does the push plus the check, called from those four sites. Happy to switch to that if you prefer.

lib/src/executor/trie_root_calculator.rs
```diff
@@ -656,7 +702,13 @@ struct InProgressNode {
 
 impl Inner {
     /// Analyzes the content of the [`Inner`] and progresses the algorithm.
-    fn next(self: Box<Self>) -> InProgress {
+    fn next(mut self: Box<Self>) -> InProgress {
+        // `next` is called right after a node's result is pushed to its parent's children. If
+        // the stack is back at the depth where the re-walk started, the re-walk is over.
+        if self.rewalk_start_stack_len == Some(self.stack.len()) {
+            self.rewalk_start_stack_len = None;
+        }
```
bkchr: I think adding some `push` helper sounds more reasonable to me and less likely to result in  an issue later on.

lib/src/executor/trie_root_calculator.rs
```diff
@@ -335,10 +338,24 @@ impl StorageValue {
                 })
             }
 
+            (None, None, 0, None) if !calculated_elem.children_partial_key_changed => {
```
bkchr, line 339: I find this check a little bit brittle. (`!calculated_elem.children_partial_key_changed`). But yeah, maybe fine for now.

bkchr (approved):

lib/src/executor/trie_root_calculator.rs
```diff
@@ -589,6 +616,14 @@ impl TrieNodeRemoveEvent {
                 }
             }
             TrieNodeRemoveEventTy::ReplacedWithSingleChild => {
+                // The subtree of the single child is walked again below. Removals in that
+                // subtree were already reported during the first walk, so reporting is turned
+                // off until the re-walk is over. An outer re-walk already in progress covers
+                // this one, so it is left untouched.
```
bkchr: One last  question, where is this rewalk triggered? I would expect that the `stack` is cleared before doing this?

lib/src/executor/trie_root_calculator.rs
lrubasze: True. 
In fact it was like this even before this PR. `children_partial_key_changed` has 2 usages:
- decide if child's Merkle value can be reused or the child has to be walked,
- decide if removed node existed in the base trie before current diff was applied (see [line 300](https://github.com/paritytech/smoldot/pull/3373/changes/BASE..87680d925a74fe38ce08e56ac04169b8f575edf4#diff-6f8df9742fb9c113cb5784838578574694aa1f99adc38668a19d40c451f3dfefR300))

I am applying second usage here.

Maybe we could just rename it to eg. `existed_in_base_trie`. Then the check in `next()` then would read as "reuse the child's Merkle value only if the parent existed".
If this fine, then I could do this here or in a follow-up PR.

lib/src/executor/trie_root_calculator.rs
lrubasze: The rewalk is triggered whenever a node is replaced by its single remaining child, and only then. It starts here: the `ClosestDescendant` returned at the end of this arm asks for the node at that child's key, and answering it pushes the child onto the stack and continues the normal walk. The child's partial key changed, so its Merkle value must be recomputed.

The stack is kept because it is what carries hashes upwards. When the rewalk starts, the removed node has already been popped, so the top of the stack is its parent, waiting for a hash in that child slot. The re-walk computes the child's new hash and puts it there. After that the top of the stack is the parent again, and the calculation continues upwards as if nothing special had happened. `rewalk_start_stack_len` remembers the parent's depth so we know when that point is reached.

Updated the comments for more clarity.

lib/src/executor/trie_root_calculator.rs
lrubasze: Added

lib/src/executor/trie_root_calculator.rs
lrubasze: Added push helper

